PR #76, opened 2015-04-01
add travis file for building on travis-ci.org

gijzelaerr:
This adds a travis file which is used by travis-ci.org to build CASACORE on various platforms and compilers and then runs the test suite. This is done for every branch and every pull request (PR). This is a *very* useful feature to make sure the code base works on various compilers and platforms, and also makes sure a PR doesn't break anything (assuming there is good test coverage).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 doc/html/
+build/
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,32 @@
+language: cpp
+
+compiler:
+  - clang
+  - gcc
+
+before_install:
+  - sudo apt-get update -q
+
+install:
+  - sudo apt-get install -qy cmake flex bison libblas-dev liblapack-dev
+      libcfitsio3-dev wcslib-dev libfftw3-dev gfortran libncurses5-dev
+      libreadline6-dev libhdf5-serial-dev python-all-dev
+
+
+before_script:
+  - mkdir build
+  - cd build
+  - wget ftp://ftp.astron.nl/outgoing/Measures/WSRT_Measures.ztar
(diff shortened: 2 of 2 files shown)


mmarquar:
Great -  I had a look at this earlier. Have you got any experience setting up OSX builds?
Also, we should add building the optional libs  such as casacore_python. Could you update the changeset to do this please.

gijzelaerr:
Yes, I have experience testing for both Linux and OS X. For now it is a bit clumbsy, you need to have a fork of the master branch containing a modified travis.yml file. But multi OS support is on its way, it will just take some time: 
http://docs.travis-ci.com/user/multi-os/

I propose we now just test on Linux (at the moment only Ubuntu 12.04) is supported and we test for both gcc and clang.

Please let me know which configure flags we should use (like the casacore_python flag).

Also, I see now that a naive build fails:

https://travis-ci.org/gijzelaerr/casacore-1/jobs/56689668#L374

any idea what could be trigger this behaviour?

mmarquar:
It looks like the checkout needs to be as "casacore" for it to work.

gijzelaerr:
ah then it should just work

mmarquar:
You have full admin rights. Could you set up the travis integration please.

mmarquar:
Some tests will also fail without the "measures data". We need to download these as part of the set up.

gijzelaerr:
I do have full admin rights, but somehow the casacore organisation doesn't show up on the travis-ci website? can you try to enable it? I don't get it.

mmarquar:
Ok, I think it's done.

gijzelaerr:
yeah i just discovered what was wrong, had to authorise travis manually to these repo's.

PR #84, opened 2015-04-04
Try running test sequentially to address memory errors

mmarquar:


mmarquar:
Ger,
Could you have a look into this please.

PR #101, opened 2015-04-30
vectorize Conversion::boolToBit

juliantaylor:
Use SSE2 _mm_movemask_epi8 (extract most significant bits of each byte)
to implement the bool to bit conversion. This is about 5-10 faster than
the scalar loop.

### DIFF
--- a/casa/OS/Conversion.cc
+++ b/casa/OS/Conversion.cc
@@ -31,6 +31,9 @@
 #include <casacore/casa/aips.h>
 #include <casacore/casa/OS/Conversion.h>
 #include <casacore/casa/iostream.h>
+#ifdef __SSE2__
+#include <emmintrin.h>
+#endif
 
 
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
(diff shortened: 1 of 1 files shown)


gijzelaerr:
Hi Julian,

Thanks for your contribution!

To make sure the code works as intended, can you maybe add a test that compares the results with and without optimization?

juliantaylor:
you mean a test that explicitly checks the non multiple of 16 data sizes? some of that is already covered by the existing tests but it could be extended.
Otherwise it that would require having two separate functions exposed to the testsuite one of which would be always dead because nobody should care about 32 bit machines anymore.

gijzelaerr:
I don't fully understand what the code does, but I thought it would just be a good way do validate the logic by having one test with a call with __SSE2__ defines and one with __SSE2__ undefined. But I trust your judgement if you say that is pointless. 

I think we cannot safely assume 64 bit only, probably 32 is still used here and there. Although speed is less of an issue there, the code should still have the same results.

juliantaylor:
that would require compiling the file twice with different options, possible but tricky to get compiler generic.
the existing testsuite `tConversion.cc` checks all possiblities for a 256 element array and a couple possibilities for a 100 element array which covers the unchanged non sse code. I could add some more tests explicitly for the non sse code, e.g. < 16 elements and a few more possibilities for non 16 multiples.

gijzelaerr:
if it is not too much work feel free to add more tests. 

I assumed undeffing and redding `__SSE2__` would work, but i never tried something like this myself. i think you are right when I think about it more now, have been programming too much Python lately ;)

juliantaylor:
the `__SSE2__` macro is defined by the compiler, by default on amd64 and otherwise when adding e.g. -msse2 with gcc. While you could probably undefine it or to be safer add an own macro to disable the code, you'd then still have to compile two testsuite visible functions.
As the operation here is relatively simple, easily testable (as there are only 256 possible values a byte) can have and the scalar code is still there to handle the non-16 multiples I think it is the better option to extend the testsuite instead of adding code bloat in the library itself.
I'll add some more tests later today.

gervandiepen:
Julian,
I've checked that tConversion works well with an #undef of __SSE2__. I've added a test to make sure it also works fine if not aligned on 16 bytes or so.
Just for curiosity. Is there a particular reason why you use &data[i] instead of data+i?
Ger

juliantaylor:
I prefer using indexing instead of direct pointer arithmetic, to me it makes it more clear you are dealing with an array of data. But its pure personal style, as far as code generation is concerned it makes no difference. The compiler will create the same machine code regardless of how you express it in code.

gervandiepen:
The improved tConversion.cc will be committed later.

PR #102, opened 2015-05-04
put a larger limit on the sort parallelization and limit worker group to 2

juliantaylor:
Even with the previously applied limit there one can still see
measurable performance degradations due to the poor parallization of the
quicksort.
Try to reduce them some more by increasing the threshold and reducing
the worker group to the 2 threads the code is currently using.
This should remove the overhead due to the barrier.

### DIFF
--- a/casa/Utilities/GenSort.tcc
+++ b/casa/Utilities/GenSort.tcc
@@ -77,11 +77,14 @@ void GenSort<T>::quickSortAsc (T* data, Int nr, Bool multiThread)
     swap (*sf, data[nr-1]);
     i = sf-data;
     if (multiThread) {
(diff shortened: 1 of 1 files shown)


juliantaylor:
@gervandiepen how attached are you to the parallelization the sorting here?
I think it would be best to remove it completely. The current implementation has really poor scalability and probably just causes more trouble than its worth as it creeps into all code due to its header nature.

tammojan:
@juliantaylor Ger is not available at the moment, is this issue urgent or could it wait two weeks?

gervandiepen:
Julian,
It's fine to remove the quicksort parallelisation. I fear it was an accidental remnant of a test I once tried.
Ger

PR #107, opened 2015-05-21
Remove sofa from casacore, fixes #105

### LINKED ISSUE #105: Update SOFA?
Under Measures, the fortran code of SOFA (standards of fundamental astronomy) is cloned. It seems that this code stems from 2008, while at the SOFA website http://www.iausofa.org/ , several new releases are available, possibly containing bugfixes etc. Should these be incorporated into casacore?

tammojan:


gijzelaerr:
we can also package up SOFA, making weekly packages just like casacore-data (measures data)

tammojan:
> we can also package up SOFA, making weekly packages just like casacore-data (measures data)

A weekly release is definitely not necessary, SOFA has a decent release system (though not a very modern packaging system). It could be packaged whenever SOFA has a release.

However, since installing sofa from source it pretty straightforward, see the jenkins script, I don't think we need to package it.

gijzelaerr:
many packages are quite straightforward to install from source, that doesn't mean you want to do it all the time :)

is there only a build time or also a runtime dependency on SOFA?

tammojan:
> is there only a build time or also a runtime dependency on SOFA?

There is only a test time dependency on SOFA (and when running the test, it's probably run time).

gijzelaerr:
ah yes, i mean test time

tammojan:
Removing SOFA seems to be good for debian packaging as well, since the SOFA license may not be compatible enough with debian's license:
https://lists.debian.org/debian-astro/2014/11/msg00008.html

olebole:
As you might have seen from the [message](https://lists.debian.org/debian-astro/2014/11/msg00008.html), the inclusion of SOFA was the reason why the package was rejected by the Debian ftp-masters.

In Debian, SOFA is already [packaged](https://packages.debian.org/source/unstable/iausofa-c), but in non-free. However, there is a free replacement available: [**erfa**](https://github.com/liberfa/erfa). Erfa is based on SOFA (and synchronized regularly) and differs only in the license, and in the function prefixes (`era` instead of `iau`). Erfa is also [packaged](https://packages.debian.org/source/testing/erfa) for Debian and Ubuntu.
So, iof you feel it is really worth to keep these checks, it may be worth to consider switching to erfa.

tammojan:
Let's see if the pull request is accepted. If it is, we don't need to switch to erfa. 

Tammo Jan

PR #110, opened 2015-05-26
remove warnings with cmake 3+

gijzelaerr:
When using cmake 3.x you get warnings, see below. Since this is science we want to be compatible with every obsolete, outdated, old fashion, antiquated and engrained cmake version imaginable, so I think best solution is to keep the 'old behavior'.

first warning:
```
CMake Warning (dev) at measures/fortran/sofa/CMakeLists.txt:5 (project):
  Policy CMP0048 is not set: project() command manages VERSION variables.
  Run "cmake --help-policy CMP0048" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  The following variable(s) would be set to empty:

    PROJECT_VERSION
    PROJECT_VERSION_MAJOR
    PROJECT_VERSION_MINOR
    PROJECT_VERSION_PATCH
```
second warning:
```
Warning:Policy CMP0054 is not set: Only interpret if() arguments as variables or keywords
when unquoted.  Run "cmake --help-policy CMP0054" for policy details.  Use the
cmake_policy command to set the policy and suppress this warning.
Quoted variables like "_cmpvar" will no longer be dereferenced when the policy is se
 to NEW.  Since the policy is not set the OLD behavior will be used.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,11 @@ project(casacore)
 
 cmake_minimum_required (VERSION 2.6.0)
 
+# fixes warnings on cmake 3.x+ For now we want the old behavior to be backwards compatible
+if (POLICY CMP0048)
+    cmake_policy (SET CMP0048 OLD)
+endif()
+
 set(PROJECT_VERSION_MAJOR 2)
 set(PROJECT_VERSION_MINOR 0)
 set(PROJECT_VERSION_PATCH 1)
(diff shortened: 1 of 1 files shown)


tammojan:
We could also just remove the quotes around `_cmpvar` .

gijzelaerr:
is that backwards compatible? if so then yes that seems to be a better solution.

tammojan:
I think that should be backward compatible: variables without quotes in if's are all around CMakeLists.txt.

gijzelaerr:
so why are the quotes there in the first place

tammojan:
> so why are the quotes there in the first place

Perhaps for when _cmpvar has a space. But that probably doesn't happen. So just remove the quotes.

gijzelaerr:
the FindcasaCore.cmake file contains similar logic, but with an unquoted _cmpvar so this is most likely all correct.

tammojan:
Merging this one, the quotes around the cmpvar are just wrong. And the policy shouldn't change anything since it's set to old.

The checks failed because travis had a wrong wcslib version, this was fixed with f47eca5

PR #114, opened 2015-06-05
Cube Tile Reuse

JimNrao:
Previously whenver the TSMCube cube object needed to evict a tile from
cache and add a new one it would write the target tile, if needed, and
then deallocate it; then a new tile was allocated and filled.  This
mod caches a tile whenver it is deallocated and allows the allocation
logic to use that cached tile rather than allocated a brand new one.
This approach reduces the amount of allocation which can become
significant when tiles are large (and they are likely to get larger in
HPC applications).

Ran the various tests named tTiled* using valgrind and fixed the one
potential error that was introduced.  Valgrind reports the writing of
unitialized storage but these errors predate this mod.

### DIFF
--- a/tables/DataMan/TSMCube.cc
+++ b/tables/DataMan/TSMCube.cc
@@ -23,7 +23,7 @@
 //#                        520 Edgemont Road
 //#                        Charlottesville, VA 22903-2475 USA
 //#
-//# $Id$
+//# $Id: TSMCube.cc 21521 2014-12-10 08:06:42Z gervandiepen $
 
 
 //# Includes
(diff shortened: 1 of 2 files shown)


gervandiepen:
A nice optimization. One little comment though.
It is not needed to keep cachedTileLength_p because TSMCube has a fixed localTileLength_p. Besides, as it is now it contains a leak because cachedTile_p is overwritten in readTile if the lengths mismatch.

JimNrao:
I'll make those changes.  I did run it on valgrind, but since that case 
can't happen, it never showed up as a leak.

I was doing some profiling and noticed I was getting a lot of soft page 
faults.  Then I used valgrind's DHAT feature to look for memory 
allocation hot-spots.  I don't know how the exact impact on performance 
but it looks like a cheap and safe optimization.

On 6/16/2015 5:31 AM, Ger van Diepen wrote:
>
> A nice optimization. One little comment though.
> It is not needed to keep cachedTileLength_p because TSMCube has a 
> fixed localTileLength_p. Besides, as it is now it contains a leak 
> because cachedTile_p is overwritten in readTile if the lengths mismatch.
>
> —
> <https://github.com/casacore/casacore/pull/114#issuecomment-112391563>.Web 
> Bug from 
> https://github.com/notifications/beacon/AGPGtDicgyPdHCjt96NK__D7XoeKyGPtks5oUAB-gaJpZM4E54yn.gif
>


-- 
Jim Jacobs

PR #115, opened 2015-06-06
Add valgrind to travis build matrix

tammojan:


tammojan:
Perhaps better to check the failing valgrind tests before merging this.

.travis.yml
```diff
@@ -4,6 +4,10 @@ compiler:
   - clang
   - gcc
 
+env:
+  - BUILD_TYPE=Release
+  - BUILD_TYPE=Debug
```
gijzelaerr: You can set `CASACORE_CHECK=yes` here also.

So this would look something like this:
```
 - BUILD_TYPE=Release CASACORE_CHECK=no
 - BUILD_TYPE=Debug CASACORE_CHECK=yes
```
which looks much better than the `if` statement below

tammojan:
@gijzelaerr, I tried to reproduce the valgrind errors on a Ubuntu 12.04 VM (fresh install, apt-get update, apt-get upgrade). No luck, valgrind runs mostly fine on my VM. Do you know how to produce an environment exactly like a travis machine?

gijzelaerr:
not sure. You can always contact travis support describing the problem,
they have been amazingly fast and helpful until now.

2015-06-08 12:40 GMT+02:00 Tammo Jan Dijkema <notifications@github.com>:

> @gijzelaerr <https://github.com/gijzelaerr>, I tried to reproduce the
> valgrind errors on a Ubuntu 12.04 VM (fresh install, apt-get update,
> apt-get upgrade). No luck, valgrind runs mostly fine on my VM. Do you know
> how to produce an environment exactly like a travis machine?
>
> —
> <https://github.com/casacore/casacore/pull/115#issuecomment-109946143>.
>



-- 
Gijs Molenaar
http://pythonic.nl

.travis.yml
gervandiepen: I believe Gijs has tried this before, but Travis timed out because it takes too long.

gijzelaerr: indeed, that was in this branch

https://github.com/casacore/casacore/tree/valgrind

may still be interesting to contact travis support.

tammojan:
I contacted Travis support, and indeed they responded promptly. I forgot to install valgrind, which is why it didn't run before. Indeed, now it takes too long. Saving this PR for when we fix #142.

juliantaylor:
an option could be to use the -fsanitize compiler options newer gcc and clang provide. They offer a only subset of valgrinds memory features but are also significantly faster. But it might still be too slow to really run on every PR.

tammojan:
I did a successful valgrind run on a VM, see the log for how long the various tests take
https://gist.github.com/tammojan/eaf0b6e9145e53663fd1

@juliantaylor I'm guessing `-fsanitize` will just get the compile time down, not the test time. Is that correct?

juliantaylor:
no it will likely increase compile time a bit and test time too as it is essentially valgrind compiled into the binary.
I can do some local tests to see how much slower it gets. You enable it by adding -fsanitize=address to the CXXFLAGS and LDFLAGS

juliantaylor:
oh its much better, regular tests take 2:46 on my machine, with address sanitation and gcc 4.8 4:47 and with valgrind 46:57
so its possibly fast enough for travis, but unfortunately with the sanitation also 7 tests fail due to out of bound access. I'll open a separate issue for that after checking if it also happens with gcc 5.1

Issue #129, opened 2015-06-23
Install the build-tools


gervandiepen:
I got a request from the Apertif group to install the valgrind suppression file in $PREFIX/share. I think it also makes sense to install tools like casacore_floatcheck as they can be useful.
A few more valgrind suppressions have to be added to catch strlen optimizations and leaks resulting from getpwnam when using a newer libc.

PR #130, opened 2015-06-23
Fixed issue #129 Install suppression file and added a few suppressions

gervandiepen:


build-tools/CMakeLists.txt
```diff
@@ -1,5 +1,15 @@
...
+install (PROGRAMS
   casacore_assay
   casacore_floatcheck
+  casacore_memcheck
 DESTINATION bin
 )
+
+install (FILES
+  casacore_memcheck.supp
+DESTINATION share
```
juliantaylor: shouldn't this be share/casacore?

gervandiepen: Julian,
Thanks. You're right. I'll change it right away.
Ger

PR #132, opened 2015-06-24
Allocator features and making element-initialization controllable.

ithubg:
Hi Ger,
Please review the modification.
I tried to keep compatibilities of the interface but please check.

Kohji

### DIFF
--- a/casa/Containers/test/tBlock.cc
+++ b/casa/Containers/test/tBlock.cc
@@ -39,7 +39,7 @@
 #include <casacore/casa/Exceptions/Error.h>
 #include <casacore/casa/iostream.h>
 #include <vector>
-
+#include <stdint.h>
 #include <casacore/casa/namespace.h>
 
 namespace {
--- a/casa/Containers/test/tBlock.cc
+++ b/casa/Containers/test/tBlock.cc
@@ -42,6 +42,36 @@
 
 #include <casacore/casa/namespace.h>
 
+namespace {
+
+struct LifecycleChecker {
+  LifecycleChecker() {
+    ++ctor_count;
+  }
+  LifecycleChecker(LifecycleChecker const &) {
+    ++ctor_count;
+  }
+  ~LifecycleChecker() {
+    ++dtor_count;
+  }
+  LifecycleChecker & operator =(LifecycleChecker const&) {
+    ++assign_count;
+    return *this;
+  }
+  static void clear() {
+    assign_count = ctor_count = dtor_count = 0;
+  }
+  static size_t assign_count;
+  static size_t ctor_count;
+  static size_t dtor_count;
+};
+
+size_t LifecycleChecker::assign_count = 0;
+size_t LifecycleChecker::ctor_count = 0;
+size_t LifecycleChecker::dtor_count = 0;
+
+}
+
 void doit()
 {
     // We do the tests in another function to make it easier to check for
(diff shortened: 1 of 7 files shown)


tammojan:
@ithubg Thanks for your contribution. Something goes wrong in the linking phase of compiling your code, could you fix this? See the log at https://travis-ci.org/casacore/casacore/builds/68108702

ithubg:
Thank you for letting me know the problem, Tammo.
I overlooked the problem because the references to the missing static member were optimized out since I specified as CMAKE_BUILD_TYPE=RelWithDebInfo when I built casacore and ran tests.
My subsequent commit to array_align_init branch will solve the problem.

tammojan:
Thanks for fixing  it. Now some tests fail, could you look into that? (See the travis details):
```
The following tests FAILED:
	359 - tLatticeStatsDataProvider (Failed)
	416 - tMSIter (Failed)
Errors while running CTest
```

ithubg:
Yes. I fixed the cause of " 416 - tMSIter (Failed)".
However, I couldn't reproduce "359 - tLatticeStatsDataProvider (Failed)" on my Linux environment and it passed on OS X environment of Travis CI.
Do you have any information about tLatticeStatsDataProvider failure? For example, the reason of the exception.

tammojan:
I haven't reproduced anything myself, I was just quoting the travis CI. After that passes, @gervandiepen can perhaps have a look at what you've done (magic to me at this point). 

You can check the travis reports on the github page (but I think you've found them by now).

ithubg:
Thanks, Tammo.
Since the tLatticeStatsDataProvider failure seems to depend on test environment not on compiler/Linux/OSX, I suspect that the size of the array below is too large to run the test on Travis CI.
When I shrank the size as below, the test case passed.
// tLatticeStatistics.cc
    // Array<Float> array1(IPosition(3,1024, 1024, 1024));
    Array<Float> array1(IPosition(3,512, 512, 512));

gervandiepen:
tLatticeStatsDataProvider is written by Dave Mehringer. I don't know that program.

I do not like that by default e.g. Array<Float> will always be initialized. The gains you can achieve by not initializing Array<Complex> are undone by this, unless we go very carefully through all code using Array<T> where T is a basic type. That is a lot of code, not only in casacore, but also in CASA, LOFAR, ASKAP, etc. IMHO this impact is too much.
I'd very much prefer a default initialization policy that leaves the current behaviour unchanged, thus 3 instead of 2 policies (DEFAULT, INIT, NO_INIT) might be required.

This week is an all-week SKA SDP-meeting, so I have very little time. Next week I'll be on vacation, so that is bad for handling this issue. How badly do you need it?

ithubg:
I see. I'll add the hybrid policy soon as a behavior when the init-policy parameter is omitted.

I'd like to give casacode developers(including NAOJ single-dish developers)  enough time to add NO_INIT parameter as possible and to implement features using aligned memory within CASA 4.5 development cycle.
So the sooner, the better.
It would be nice if my pull request with the hybrid policy would be okay for you and be merged to master.

casa/Containers/test/tBlock.cc
```diff
@@ -42,6 +42,36 @@
 
 #include <casacore/casa/namespace.h>
```
juliantaylor: this file needs a `stdint.h` (or cstdin for c++11) include for `intptr_t´ otherwise it fails on some platforms (e.g. ubuntu 14.04)

juliantaylor:
would it make sense to set the default alignment to 32 bytes or maybe even 64 bytes? a few bytes wasted memory per allocation should not be a too big concern for `Block`

casa/Containers/test/tBlock.cc
ithubg: Thanks, Julian!
I fixed it.

ithubg:
Yes, of course. I changed the default value to 32 bytes and it can be configured from cmake command line as below.
 -DCASA_DEFAULT_ALIGNMENT:STRING=64

Thanks again.

gervandiepen:
Kohji,
I've merged the request. I had no time to look at it in detail, but the builds succeeds and I'm sure you've done a good job.
Ger

ithubg:
Thank you for your support despite how busy you are!
I'll  move to casa::Array.

juliantaylor:
this PR forgot to install Allocator.h, fixed in linked PR

PR #135, opened 2015-06-24
Remove use of deprecated numpy API

juliantaylor:
Direct access to members of numpy structures is deprecated.
Remove the structure access with the associated access functions which
exist since at least numpy 1.1.


gervandiepen:
Julian,
Great, Thanks for the changes.
Ger

PR #136, opened 2015-06-24
Introduce pread/pwrite calls to IO

juliantaylor:
pread/pwrite combines lseek + read/write into one system call but does
not change the offset of the file descriptor.
This is useful when multiple threads work on the same file descriptor.
Also most of casa's IO is actually of this type so this could be used to
cut the number of system calls per IO request in half. But as it does
not change the file descriptor actually using it needs some
investigation if code relies on the position of the descriptor.

For now only use it for the MultiFile IO as that is a new method that is
not directly used anywhere else except casacore and Lockfile which is also very self contained.

the calls need POSIX 2001, do we need some cmake check for this?


gervandiepen:
Thanks for this change. It looks very valuable.
The standard is so old, that I cannot imagine systems not supporting it.
What do you think?
Yet, I'll ask the people using Grid systems, because they tend to be very
conservative.

tammojan:
> Yet, I'll ask the people using Grid systems, because they tend to be very conservative.

@gervandiepen , what is te status of this? @ygrange , are you a people using Grid systems? Can we assume POSIX 2001?

ygrange:
Well, "using the grid system" is a bit much (I can use it, have used it but not extremely seriously). Not from the top of my head. I will also ask some of the people too and come back to you.

ygrange:
SARA grid support has informed me of the following:
The standard that applies to grid locations is called the UMD (voor unified middleware distribution) standard. That standard requires the OS to be Scientific Linux 5 or 6, Centos 6 or Debian 6 and some middleware. 

One of their admins has tested compliance using the following test: http://sourceforge.net/projects/posixtest/ and reports that in essence their system is not POSIX 2001 compliant according to this test.  I am however not sure if this test is really sensible (if we define a piece of code that tests the functionality you want to implement, I could run it on a bunch of Grid nodes in Amsterdam and Groningen and just see how it returns).

According to Groningen their OS reports POSIX 2003 (which seems a non-existant standard).

tammojan:
@juliantaylor Could you add the following lines somewhere in the CMakeLists.txt in the root directory?
```
# Since 2015, we need pwrite and pwrite (POSIX 2001)
include(CheckFunctionExists)
check_function_exists(pread HAVE_PREAD)
check_function_exists(pwrite HAVE_PWRITE)
if (NOT (${HAVE_PREAD} AND ${HAVE_PWRITE}))
        message(FATAL_ERROR "Casacore requires pread and pwrite functionality" )
endif (NOT (${HAVE_PREAD} AND ${HAVE_PWRITE}))
```

juliantaylor:
added the checks

PR #137, opened 2015-06-25
Made change in tArray that corresponds to the method name change.

JimNrao:


JimNrao:
New pull request to fix #113 which had needed a method name change to correspond to the main logic modified in that pull request.

PR #139, opened 2015-06-26
Adding CMAKE_INSTALL_NAME_DIR in CMakeLists.txt

vsuorant:
I added CMAKE_INSTALL_NAME_DIR in CMakeLists.txt as this is needed for OS X builds of CASA.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,6 +65,7 @@ if (ENABLE_SHARED)
     option (BUILD_SHARED_LIBS "" YES)
     if (ENABLE_RPATH)
         # Set RPATH to use for installed targets; append linker search path
+        set(CMAKE_INSTALL_NAME_DIR "${CMAKE_INSTALL_PREFIX}/lib" )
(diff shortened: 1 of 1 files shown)


tammojan:
@vsuorant : the two test commits also landed on this pull request, you could get rid of those by using `git rebase -i HEAD-3` and using the squash option (or removing the test commits which in this case is the same),, see for example http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

Other than this, I see no problems with this change. The travis job failed, but only because of a timeout. In fact, three of the four travis jobs finished successfully before the timeout.

Since this is a build system change, should @mmarquar merge it?

vsuorant:
Thank you for pointing out the error in the build request and for the instructions. This should be fixed now. Let me know if there are further problems with this.

tammojan:
I'll merge this change because it a) only affects OSX b) follows standard CMake syntax and c) passes all tests. Hope @mmarquar doesn't mind.

Issue #140, opened 2015-06-29
MSSummary now needs at least 1 row in STATE table


gervandiepen:
The LOFAR program msoverview uses the casacore class MSSummary which in its turn uses MSMetaData. The program used to run fine if the STATE subtable is empty, but now it aborts with the following error:

orru@locus070:/data/scratch/lofarsys/Observation343426$ msoverview in=concat.ms
Error: Exception: MS only has 0 rows in its STATE table, but references STATE_ID 0 in its main table..
... thrown by casa::CountedPtr<casa::Vector<int> > casa::MSMetaData::_getStateIDs() const at File: /opt/cep/casacore/builds/casacore-2.0.1/ms/MSOper/MSMetaData.cc, line: 556

Although I agree the MeasurementSet is not fully conforming th MSv2 standard, it always used to run fine. I would like to change MSMetaData such that an empty STATE table is acceptable, because it is impossible to change existing LOFAR MSs. If Dave agrees, I can make the change myself.

tammojan:
This issue was resolved in #141

PR #141, opened 2015-06-29
Allow an empty STATE subtable in MSMetaData (for LOFAR MSs) (fix issue #140)

gervandiepen:
Allow an empty STATE subtable in MSMetaData (for LOFAR MSs)


tammojan:
@gervandiepen Please use a descriptive commit message, as the tracker is not strictly part of the repository. 

The test fails because you based the change on a non-compiling master. Probably `git fetch`, `git rebase origin/master` will fix this (but also I am still learning).

PR #143, opened 2015-06-30
Docs: Add 'build-essential' and 'g++' to README package list.

timstaley:
These seem to be necessary when building against Ubuntu 14.04 under vagrant,
and aren't installed as part of the default box image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 
 # Casacore
 
-A Suite of c++ libraries for radio astronomy data processing.
+A suite of c++ libraries for radio astronomy data processing.
 
 
 # Installation
@@ -22,6 +22,7 @@ To compile casacore you need to meet the following requirements:
 
 * cmake
 * gfortran
+* g++
 * flex 
 * bison
 * blas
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,14 +38,15 @@ To compile casacore you need to meet the following requirements:
```
juliantaylor: this typo should also be fixed as your at it, should probably be `ncurses`

README.md
```diff
@@ -37,14 +38,15 @@ To compile casacore you need to meet the following requirements:
...
+$ sudo apt-get install build-essential cmake gfortran g++ libncurses5-dev \
+    libreadline-dev flex bison libblas-dev liblapacke-dev libcfitsio3-dev \
+    wcslib-dev
 ```
 
 and the optional libraries:
 ```
 $ sudo apt-get install libhdf5-serial-dev libfftw3-dev python-numpy \
-    libboost-python-dev
+    libboost-python-dev libopenmpi-dev
```
juliantaylor: hm, what in casacore needs mpi? didn't find anything with grep

timstaley: I'm pretty sure I needed that to build with the `-DUSE_OPENMP=ON` option, but I could be mistaken.

juliantaylor: openmp is not the same as openmpi, (shared memory vs message passing)
openmp needs libgomp but on debian based systems that is already a dependency of gcc

timstaley: OK ta, I'll recheck it

gijzelaerr:
fix, rebase and merged https://github.com/casacore/casacore/commit/6aa2d1187f287a4f03b7e71c36ca3ed08f41c1bc

Issue #148, opened 2015-07-02
tUDFMSCal needs the LD_LIBRARY_PATH to be set


gervandiepen:
tUDFMSCal failed to run after the recent merges. It loads derivedmsca.so dynamically, but it appeared an incorrect version was loaded because the LD_LIBRARY_PATH was no set correctly.
It needs a .run file to set the paths correctly.

PR #149, opened 2015-07-02
Issue #148: set LD_LIBRARY_PATH for tUDFMSCal

gervandiepen:
Added tUDFMSCal.run to set LD_LIBRARY_PATH correctly


juliantaylor:
if you put the issue number into the commit message instead of the subject github will link them together properly.
Also then the associated issues should get closed automatically when you merge the PR.

PR #152, opened 2015-07-02
Install Allocator.h needed by public Block.h

juliantaylor:
The file was introduced in gh-132 but it was forgotten to install it.


ithubg:
Oops. I's my fault. Thank you very much!

Issue #56, opened 2015-03-31
Non-LGPL compliant licenses in source files


GoogleCodeExporter:
```
The files install/printer/range.c and lwf_c.c have licenses incompatible to the 
LGPL-2 (do-not-sell clause). I contacted their author and got the permission to 
re-release these files under LGPL and to change the copyright information 
accordingly. The files in the casacore tree should be changed to reflect this 
situation. A patch is attached.

Also, the file images/Images/ImageReorderer.h is put under GPL (and not LGPL) 
which infects the whole casacore package (which therefore would need to be 
licensed as GPL as well). This file should be relicensed as well.

Best regards

Ole Streicher <debian@liska.ath.cx>
```

Original issue reported on code.google.com by `ole.stre...@gmail.com` on 18 Sep 2012 at 8:09

Attachments:
* [lgpl_change_range_c.patch](https://storage.googleapis.com/google-code-attachments/casacore/issue-56/comment-0/lgpl_change_range_c.patch)

gijzelaerr:
So we can close this issue by removing the install directory from the repository right?

gervandiepen:
Right. Malte said he would remove all old stuff, but obviously he is too
busy with ASKAP.
Gijs, do you remove install?
I'll remove the other old stuff at a later stage.

gijzelaerr:
well i'm quite busy also at the moment so please don't wait for me for removing a folder the repo.

gervandiepen:
Allright. I'm also busy (next week the SKA SDP meeting), but I'll might
have time tomorrow.

gervandiepen:
I'm removing all of install and all other obsolete files.
I plan to move install/COPYING to casacore, but I'm not sure if it has the right text for all of casacore. Maybe Ole can tell.

tammojan:
Hi @gervandiepen , if you mention Ole, it's better to use an @-mention like @olebole so he'll be notified.

tammojan:
SOFA used to have a different license, but since it is now removed from casacore, that is not a problem anymore. Perhaps there are other imported libraries? What about the license of Mirlib?

PR #153, opened 2015-07-02
Issue #56 Remove all obsolete files and directories

gervandiepen:
Removed all obsolete files and directories


juliantaylor:
this probably shouldn't remove install/COPYING
isntead casa/test/COPYING could be removed as that its location does not imply its valid for the full sourcve

gervandiepen:
You might be right. I'll check.

gervandiepen:
Maybe better to move COPYING to the casacore directory. What do you think?

juliantaylor:
thats probably better, the top level directory is usually the place you find the license

tammojan:
Rebased this on top of the current master, so checks will run with proper travis settings.

tammojan:
@juliantaylor Could you merge this?

PR #138, opened 2015-06-25
CAS-7319, added several new methods to MSMetaData

dmehring:
added various features to MSMetaData

	modified:   ms/MSOper/MSKeys.cc
	modified:   ms/MSOper/MSKeys.h
	modified:   ms/MSOper/MSMetaData.cc
	modified:   ms/MSOper/MSMetaData.h
	modified:   ms/MSOper/MSSummary.cc
	modified:   ms/MSOper/test/tMSMetaData.cc

PR #158, opened 2015-07-07
fix infinity loop and missing check in MSMetaData

juliantaylor:
on behalf of @dmehring, fixes issues introduced in gh-138


juliantaylor:
some unittest for the issue would be good but as it currently blocks casa from working I'll merge after the tests ran.

dmehring:
Unittests would require significant sized data sets to be committed to the repository.

dmehring:
Our (CASA) listobs unit tests tripped up on this issue. They now run correctly.

PR #159, opened 2015-07-07
Compatibility with CFITSIO < 3.181

tammojan:
Added some CMake magic for getting the cfitsio version (which is stored as a float, argh). CFITSIO_VERSION_MINOR contains the decimal part up to three digits without leading zeros.

Because the preprocessor can only compare integers, parsing the version string to major and minor versions is necessary. This is done in FindCFitsIO.cmake .


juliantaylor:
tested on a redhat5 with 3.030 and works, thanks

JimNrao:
Does this mean that troublesome code in blockio.cc can be ifdefed?

tammojan:
> Does this mean that troublesome code in blockio.cc can be ifdefed?

Yes, it already is ifdefed.

JimNrao:
Outstanding.  Thanks.

On 7/7/2015 3:24 PM, Tammo Jan Dijkema wrote:
>
>     Does this mean that troublesome code in blockio.cc can be ifdefed?
>
> Yes, it already is ifdefed.
>
> —
> <https://github.com/casacore/casacore/pull/159#issuecomment-119346908>.Web 
> Bug from 
> https://github.com/notifications/beacon/AGPGtH8LdplzIKBH9TULFCCs52e_mfpWks5obDspgaJpZM4FT47V.gif
>


-- 
Jim Jacobs

PR #165, opened 2015-07-13
Missing CASA to CasaCore Merges (Retry)

JimNrao:
Several files modified late in the transition period failed to make it into
CasaCore as the code became unified.  This commit includes the missing files
and some test data associated with the related unit-tests.

A previous PR for this purpose was killed because it became muddied some how.


tammojan:
Still some files missing (I'm quoting travis)
```
CMake Error: File /home/travis/build/casacore/casacore/images/Images/test/decon_test.im/table.lock does not exist.
```

Does it compile at your end? What perhaps `git status` could show the missing files.

JimNrao:
Added the lock files which cmake apparently wants.  Seems sort of odd to me, but ...

PR #173, opened 2015-07-21
Prepare for 2.0.3 bugfix release

tammojan:
Let's skip a 2.0.2 release because somehow the patch version was already at 3. The fix for msoverview is quite important for LOFAR, that's why I'm making a release now. 

In addition to the standard release stuff, I added some defines: `CASACORE_MAJOR_VERSION`, `CASACORE_MINOR_VERSION` and `CASACORE_PATCH_VERSION`. This should make it easier for external software to do an ifdef based on the casacore version.


tammojan:
Trying again, I got the order of the commits broken.

Issue #169, opened 2015-07-16
MSAntennaParse cannot handle an empty selection


gervandiepen:
MSAntennaIndex throws an exception if no matches are found for an antenna name pattern, even if used with negation. It makes it impossible to have standard selection parameters in a pipeline where you do not know if such antennae are part of the observation or not.
Another problem is that UDFMSCal does not set a pointer to an error handler in case MSAntennaParse needs to report an error.

gervandiepen:
Also fixes segfault in UPDATE command

PR #174, opened 2015-07-21
Issue #169: Make empty MS selections possible in UDFMSCal

gervandiepen:
Issue #169: Make empty MS selections possible in UDFMSCal


gervandiepen:
Also fixes segfault in UPDATE command

PR #175, opened 2015-07-21
commit for CAS-4852, so that a single field in a multifield ms can be…

dmehring:
… written to an msfits file


tammojan:
@dmehring I edited the commit message to be a bit shorter (should be about 75 characters), will merge the fixedup version in #176.

PR #177, opened 2015-07-22
Array align init2

ithubg:
This is the enhancements for casa::Array to handle aligned memory and to skip unnecessary initialization.
It would  be nice if you could merge this pull request as soon as possible.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,7 +70,6 @@ option (USE_STACKTRACE "Show stacktrace in case of exception" NO)
 option (CASA_BUILD "Building the in the CASA (http://casa.nrao.edu) environment" NO)
 
 set(CASA_DEFAULT_ALIGNMENT "32" CACHE STRING "Default alignment of casa::AlignedAllocator")
-add_definitions("-DCASA_DEFAULT_ALIGNMENT=${CASA_DEFAULT_ALIGNMENT}")
 
 # Test if shared libraries have to be built.
 if (ENABLE_SHARED)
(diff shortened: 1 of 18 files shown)


ithubg:
Could someone restart the Travis CI builds above for OS X?
They failed before starting casacore build due to timeout of 3rd party library installation.

tammojan:
I'm not sure a restart will do the trick, the master is failing for the same reason. I'll try to restart however.

ithubg:
Thanks so much, Tammo.

tammojan:
Travis is having some issues with the OSX builds. The linux builds seem to be fine.

@gervandiepen : could you review this change?

ithubg:
Hi Ger,
Because this pull request need to be merged to master before CASA4.5 feature freeze, it would be nice if  you could review and merge it by July 29th.

ithubg:
OS X builds failed with the following message this time.
The job exceeded the maxmimum time limit for jobs, and has been terminated.

tammojan:
Travis is having issues with OSX now, so I'm afraid we cannot test on osx. Have you tested this PR yourself on OSX?

ithubg:
I'm afraid I have not tested on OS X.

gervandiepen:
Why has NAOJ been added to the copyright section? It has never been done for ASTRON, ATNF, ESO, etc.

ithubg:
Because there is a possibility to reuse some part of the code in other NAOJ products, we want to preserve the copyright for the code.
Of course, we agree that the code is used and distributed under the license of casacore.

gervandiepen:
I've done the merge.
Array is getting quite complex. Maybe we need a (short) note in casacore-notes describing the Array class and its allocator policies.
Maybe the changes have an effect on compilation time which has increased. For C++11 we could preinstantiate Array and Vector for the commmonly used types (e.g. the column types in tables). In this way they do not need to be instantiated in each compilation unit. Could you try this out?

ithubg:
Thanks Ger.

The description for developers are available below.

https://safe.nrao.edu/wiki/bin/view/Software/CASA/TipsToOptimizeCppCodes#CASA_Specific_Tips

Yes for the preinstantiation.
Are Bool, Int, Float, Double, Complex, DComplex, String, MEpoch, MPosition and MDirection enough for "extern template"?

PR #179, opened 2015-07-22
require c++11 compatible compiler for CASA_BUILD

juliantaylor:


juliantaylor:
casa now requires c++11 to compile, so casacore also needs to be compiled with this standard in that mode. This makes using a incompatible compiler a configure failure for CASA_BUILD only.

tammojan:
I hope this does not mean that Casa will add c++11-only features to casacore? Policy up to now is to have some alternative for older compilers.

juliantaylor:
I am not aware of plans to require c++11 in core in the near future. Though this might screw with plans of reusing the same casacore for multiple applictions which @gijzelaerr mentioned in another issue.
Transition to c++11 is all or nothing as mixing 98 and 11 symbols in the same application can cause trouble due to binary incompatibility.

tammojan:
@juliantaylor Agreed. This will mean that if you want to compile CASA from source, you'll have to enable c++11-support for all other applications using the same casacore.

juliantaylor:
please don't merge yet to give some CASA devs time to comment (pinged via email)

PR #181, opened 2015-07-22
add better c++11 macro

juliantaylor:
AIPS_HAVE_CXX11 denotes c++11 support or newer. This is usually better
because c++ standards are usually souce level backward compatible so a
minimum version check is more future proof.

### DIFF
--- a/casa/aipsenv.h
+++ b/casa/aipsenv.h
@@ -32,8 +32,8 @@
 
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
-// Set if compiler supports C++11.
-#if __cplusplus == 201103L
+// Set if compiler supports C++11 or newer
+#if __cplusplus >= 201103L
 #define AIPS_CXX11
 #endif
 


juliantaylor:
similar to the recent gcc macro change, making stuff require an explicit version when not actually required just leads to stuff failing in future.

tammojan:
Now the define `AIPS_CXX11` is used nowhere anymore. I agree that the name `AIPS_HAVE_CXX11` is more descriptive, so perhaps just remove the define `AIPS_CXX11`?

juliantaylor:
that would be better, but technically not backward compatible. But the macro is relatively new so its not too likely it would cause major headaches.
Less disruptive but also not fully backward compatible would be to change the meaning of `AIPS_CXX11`. It is very unlikely someone has found a source incompatibility in c++14 and is using that macro to guard against it ;)

juliantaylor:
updated to just change the meaning of AIPS_CXX11

PR #187, opened 2015-07-23
CAS-5524: throw exception when importing uvfits files with stokes I,Q,U, or V

dmehring:


tammojan:
I see that you use an external FITS file (`$CASAPATH/data/regression/unittest/uvfits/1331+305_I.UVFITS`) in the test. I'm not too enthousiastic about dependency on casa files. Could we include it in the casacore repository so that the test can be run everywhere?

dmehring:
If you'd like. In the past, Ger preferred keeping test fixtures, especially large ones, out of casacore. This one is about 3 MB in size.

dmehring:
We can also remove this test and I can implement it in casa only, whichever you prefer. I do note that this is not the only case where we have tests like this. These tests still "run" in casacore; they just exit immediately without failing if the necessary test fixtures are no present. So there really is no dependency in that sense.

dmehring:
@tammojan This implementation is time critical to our project, so please let me know which option you would prefer as I outlined yesterday:

1. remove the test
2. commit the uvfits file to casacore
3. do nothing since this test does not constitute a hard dependency on casa and since there are already other tests in casacore which use this same exact pattern.

If 3. please complete my pull request.
Thanks

tammojan:
@dmehring No idea who should decide on this, but let's decide on 3 for time's sake. We can move to 1 or 2 after an elaborate discussion with drink's.

dmehring:
@tammojan I'm all for 3. Thank you.

PR #188, opened 2015-07-23
CAS-7584 allow caller to specify seeds so the same random numbers are…

dmehring:
…n't generated each time


tammojan:
@dmehring Could you make your commit messages a bit shorter next time? The standard is about 70 characters. This is useful in overviews like https://github.com/casacore/casacore/commits/master .

Otherwise, seems like a good change, I'll merge it because LatticeMath currently has no module owner.

PR #195, opened 2015-08-07
fix cmake output message on missing c++11 support

juliantaylor:
non CASA_BUILD setup still printed c++11 support enabled when the
compiler check failed.


juliantaylor:
It might make sense to make that case fatal too. I don't think its a good to silently ignore what the user requested.

tammojan:
Can I suggest #196 instead?

PR #196, opened 2015-08-07
Fix CMake output when checking C++11 support

tammojan:
Alternative to #195 (achieves the same thing). I don't really like that CMake changes an option that is explicitly set.


juliantaylor:
this is a bit nicer, so ok

Issue #194, opened 2015-08-07
MSAntenna destructor throws exception


juliantaylor:
throwing exceptions in destructors is problematic as it will lead to termination when called during stack unwinding in c+98 and always terminate in c++11 as destructors are implicitly marked with `noexcept`

this can be seen been running the tMeasurementSet unit test using -DCXX11.

tammojan:
If I run the test, even when using -DCXX11, the test passes fine, and building it is no problem either. Am I missing something?

juliantaylor:
do you have a c++11 compatible compiler? the cmake output should state that in the `C++11 support` line

tammojan:
Yes, `-- C++11 support ......... = True`

juliantaylor:
which compiler is it? it fails for me with gcc 4.8 and 4.9

tammojan:
It's gcc 4.6.3.

juliantaylor:
that one probably did not implement the full standard yet, it actually predates it by a year, weird it actually has the c++11 flag and not only the c++0x one.

juliantaylor:
`g++-4.6 -std=c++11 unrecognized commandline options -std=c++11`
do you have some kind of patched version? casacore should not accept that as a compatible compiler

tammojan:
It's not patched. Indeed, the check fails: 
```
-- Performing Test HAS_CXX11
-- Performing Test HAS_CXX11 - Failed
```
I set `-DCXX11=True` on the cmake command line. Line 147 of CMakeLists should probably make it False again, but I think the wrong syntax is used. Shouldn't `option(CXX11 "" No)` be replaced by `set(CXX NO)`?

gervandiepen:
I'll put in the change made by Harm Munk some time ago.

Issue #134, opened 2015-06-24
Remove RO classes in MeasurementSets


gervandiepen:
Some time ago all Tables RO classes like ROTableColumn have been integrated in TableColumn and so. Typedefs are created for backward compatibility.
The same changes could be made for the MeasurementSet column classes. It will simplify the classes and reduce the amount of code.

gervandiepen:
Also investigate if a base class MSTableBase could be created to factor out common code in the various MSxxx classes.

PR #197, opened 2015-08-10
Exceptions thrown in destructors of MS classes

gervandiepen:
Fixes issue #194

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,31 @@
+language: cpp
+
 os:
   - linux
   - osx
 
-language: cpp
-
 compiler:
   - clang
   - gcc
+
+env:
+  - CXX11=True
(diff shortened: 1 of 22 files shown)


tammojan:
Probably you don't want to change the namespace back to casa.. And why comment the exceptions instead of deleting them?

gervandiepen:
Good point. I'll change the namespace to casacore.
It is outcommented because Harm did so.

HarmMunk:
Delete them. I just left them in so that you could see the change without going through the (limited) hassle of a compare.

tammojan:
It's not really pretty to see the amount of code duplication here. But perhaps that can be fixed some other time. See issue #134. Otherwise the code seems fine, so merging.

PR #210, opened 2015-08-14
make initialization behavior of array::resize without ArrayInitPolicy parameter inconsistent.

ithubg:
To improve the performance of resize without ArrayInitPolicy parameter, the behavior of the resize becomes inconsistent.


ithubg:
Could you update the API document below?
http://casacore.github.io/casacore/

The API is not up-to-date and doesn't include new functions and parameters for Block/Array/Allocator.
Thanks in advance.

PR #212, opened 2015-08-14
Modify MSIter to allow using memory resident sort tables

JimNrao:
Added a feature to modify MSIter so that the user has the option of specifying
that MSIter use a memory-resident table for the SORT_TABLE that it creates.
In concurrent operation a disk-resident SORT_TABLE allows two readonly
processes to interfere via the sort table.  When the new, optional ctor
parameter storeSort is set to false MSIter will use a memory resident
sort table that will vanish with the end of the creating process.  The
default value of the new parameter is true which means that the operation
of the ctor will be the same for existing code.


tammojan:
The diff of this change is strange, it looks as if all lines were changed. Was an encoding changed or something?

JimNrao:
I saw that too but didn't investigate; it was so global I just assumed 
it was a formatting issue (maybe running everything through something 
like astyle would eliminate those).

----

I got the original file from another CASA member.  Looks like the lines 
delimiters are \r\n rather than \n.  I think that only Windows uses that 
so I'm puzzled about how they got in there since I don't think anyone 
here is using Windows although maybe when he e-mailed the files 
something helpfully converted them to the "MS standard".

On 08/14/2015 02:57 PM, Tammo Jan Dijkema wrote:
>
> The diff of this change is strange, it looks as if all lines were 
> changed. Was an encoding changed or something?
>
> —
> <https://github.com/casacore/casacore/pull/212#issuecomment-131237317>.Web 
> Bug from 
> https://github.com/notifications/beacon/AGPGtBKcuLevuhs63lXpxgyDvXd68NPxks5onk2ogaJpZM4Fr8V_.gif
>

-- 
Jim Jacobs

phone: 835-7235
office: 301

PR #218, opened 2015-08-19
Slight interface changes to allow specifying all obsids and arrayids

dmehring:


JimNrao:
Did these changes affect (or should affect) the related unit tests?

dmehring:
No. The unit tests aren't affected. Just changes of uInts to Ints.

PR #221, opened 2015-08-25
CAS-7770, added getRefDir() for ephemeris sources

dmehring:
added getReferenceDirection() which returns reference direction for both non-ephemeris and ephemeris (if optional epoch is provided).

### DIFF
--- a/ms/MSOper/MSMetaData.cc
+++ b/ms/MSOper/MSMetaData.cc
@@ -1003,6 +1003,20 @@ MDirection MSMetaData::phaseDirFromFieldIDAndTime(const uInt fieldID,  const MEp
 	return msfc.phaseDirMeas(fieldID, inSeconds);
 } 
 
+MDirection MSMetaData::getReferenceDirection(
+    const uInt fieldID,  const MEpoch& ep
+) const {
+    _hasFieldID(fieldID);
+    ROMSFieldColumns msfc(_ms->field());
+    if(! msfc.needInterTime(fieldID)) {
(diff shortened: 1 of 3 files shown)


dpetry:
Hi, in principle the test for the intertime (i.e. for the time dependence) is not needed
as it happens also  inside referenceDirMeas. If you supply a time to a time-independent
field direction, it is ignored.
But then maybe what you coded is slightly faster because you save the construction
of the epoch object.

In the unit test, one should maybe add one case where the ep argument is explicitly provided.

I would merge this but I can't because there is some message that your branch has conflicts.
Will need to investigate ...

dpetry:
Dave, could you update your branch and the pull request?

dmehring:
Dirk,
Is the problem that I cannot submit multiple pull requests for the same file simultaneously? ie can I only submit the second request after the first request has been merged?

dpetry:
That could be. The other request has been merged. May you could create this pull request again?

dmehring:
OK, I'll see what I can do. Given the frequency with which I need to modify certain casacore files, if I have to wait to submit more changes until a merge has been completed each time I change something, then that will be a bottleneck, so I will note that down to pass along to management.

dpetry:
I think this is a special case where you were submitting a change to the _same_ file in two different requests at nearly the same time. Maybe you could just combine the requests the next time.
But I agree, it is a bit awkward.
In our old system, it would be like waiting for Jenkins to complete before making the next commit.

PR #222, opened 2015-08-25
CAS-7837 account for case of no intents, so segfault doesn't occur

dmehring:
In the case of the MS having no intents, a segfault would occur in _getFieldsAndIntentsMaps(). This submissions fixes that by returning immediately if that condition is met.


dpetry:
I'll merge this now but please still fix the indentation at some point.

Issue #217, opened 2015-08-18
imager.clean crashes with "pure virtual method called"


david-terrett:
When run on a heavily loaded multi-core system, imager.clean() crashes with the message "pure virtual method called". A stack dump shows that the origin of the fault is Table::reliquishAutoLocks() and it appears that this method is calling the hasLock() method of a PlainTable that has been, or is in the process of being deleted, in another thread. The likely result is a call to a virtual function in the base class (see http://www.artima.com/cppsource/pure_virtual.html).

The basic problem is that although the methods that access the table cache use a mutex to protect the table while it is being accessed, they return a raw pointer to a PlainTableand that table may be deleted by another thread while the PlainTable is being accessed through the pointer. (there is another race condition, but not one we have seen in practice, in that the code fetches the number of entries in the table cache and then iterates over the table using an index. If an entry in the table is deleted by another thread, it could attempt to fetch an entry beyond the end of the table and get a null pointer returned).

The problem can be fixed by moving the code of Table::relinquishAutoLocks into TableCache and making the calls to the Table methods while holding the TableCache mutex which blocks the deletion of the tables in other threads. This solution has been implemented and tested and does fix the problem.

I can supply a patch file (I don't seem to be able to attach it to the issue)

gervandiepen:
Please send me your patch, so I can take a look at it and generate a pull
request.
Ger

david-terrett:
Ger,

Here are the modified a files and a patch file.

Just so you know, we are seeing another failure related to the same area of code. imager.regrid() throws a TableError because a table cannot be deleted because it is in use by another process. I suspect it is another issue with finding a table in the cache that is in the process of being deleted but I haven’t engineered and tested a fix yet although I think I can see how to do it.

David

---------------------------------------------------------------
David Terrett                          david.terrett@stfc.ac.uk
                        RAL Space
                        STFC Rutherford Appleton Laboratory
                        Harwell Oxford
                        Didcot, Oxfordshire
Tel: +44-1235-446478    OX11 0QX
Fax: +44-1235-446362    United Kingdom
---------------------------------------------------------------
   The contents of this email are sent in confidence for the
   use of the intended recipients only.  If you are not one of
   the intended recipients do not take action on it or show it
   to anyone else, but return this email to the sender and
   delete your copy of it.
---------------------------------------------------------------

gervandiepen:
Thanks for the patch.

The "in use" error message is another issue.
[...]

PR #223, opened 2015-08-26
Make TableCache fully thread-safe

gervandiepen:
Issue #217 Make TableCache fully thread-safe

### DIFF
--- a/tables/Tables/PlainTable.cc
+++ b/tables/Tables/PlainTable.cc
@@ -42,7 +42,7 @@
 #include <casacore/casa/OS/HostInfo.h>
 #include <casacore/casa/OS/File.h>
 #include <casacore/casa/System/AipsrcValue.h>
-
+#include <time.h>    //# for nanosleep
 
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
@@ -288,6 +288,9 @@ void PlainTable::closeObject()
         //# File locking support in Lustre (maybe other file systems too)
         //# seems to be asynchronous to some degree, so try a few times.
         int nTrys = 5;
+        timespec timet;
+        timet.tv_sec = 1;
+        timet.tv_nsec = 0;
         while (isMultiUsed(False)) {
             if (nTrys == 0) {
                 unmarkForDelete (False, "");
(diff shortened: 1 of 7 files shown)


tables/Tables/PlainTable.cc
```diff
@@ -285,12 +285,19 @@ void PlainTable::closeObject()
...
+        //# seems to be asynchronous to some degree, so try a few times.
+        int nTrys = 5;
+        while (isMultiUsed(False)) {
+            if (nTrys == 0) {
+                unmarkForDelete (False, "");
+                throw (TableError ("Table " + name_p + " cannot be deleted;"
+                                   " the table or a subtable is still used"
+                                   " in another process"));
+            }
+            sleep(1);
```
juliantaylor: sleep has poor interaction with signals, please use nanosleep

gervandiepen: Thanks. Good point.
I got the fix from somebody else.

tables/Tables/TableCache.cc
```diff
@@ -47,19 +57,6 @@ PlainTable* TableCache::operator() (const String& tableName) const
...
-    ScopedMutexLock sc(itsMutex);
-    return (PlainTable*) (tableMap_p.getVal (index));
-}
-
-uInt TableCache::ntable() const
-{
-    ScopedMutexLock sc(itsMutex);
-    return tableMap_p.ndefined();
-}
-
```
juliantaylor: removal of these two functions is not backward compatible and broke CASA compilation, I'll check if it is easy to fix within casa.

juliantaylor: was easy, casa build should be fixed.

gervandiepen: Great, thanks.

tables/Tables/TableCache.cc
```diff
@@ -95,5 +92,113 @@ void TableCache::rename (const String& newName, const String& oldName)
...
+}
+
+Vector<String> TableCache::getTableNames() const
+{
+    ScopedMutexLock sc(itsMutex);
+    uInt ntab = tableMap_p.ndefined();
+    Vector<String> names;
+    for (uInt i=0; i<ntab; i++) {
+	PlainTable& table = *static_cast<PlainTable*>(tableMap_p.getVal(i));
+        names[i] = table.tableName();
```
juliantaylor: does this work for casa::Vector? shouldn'T this be push_back?

PR #228, opened 2015-09-04
CAS-7637 add getRestFrequencies() and getTransitions() to MSMetaData

dmehring:
added SourceKey struct to MSKeys

added getRestFrequencies() and getTransitions() to MSMetaData along with private method for retrieving and storing data from SOURCE table.

Replaced CountedPtr with SHARED_PTR in MSMetaData, since that's what we are moving toward.

Replaced a couple of instances of "brute force" code with uses to Measures columns.


dmehring:
@gervandiepen @dpetry would one of you please merge pull request #228 I submitted last week? Thanks.

PR #233, opened 2015-09-14
Remove extraneous call to sqrt()

mpokorny:
In converting StatsData to Record, application of sqrt() to stddev field
value is removed.

### DIFF
--- a/scimath/Mathematics/StatisticsTypes.tcc
+++ b/scimath/Mathematics/StatisticsTypes.tcc
@@ -99,8 +99,7 @@ Record toRecord(const StatsData<AccumType>& stats) {
 		StatisticsData::toString(StatisticsData::RMS), stats.rms
 	);
 	r.define(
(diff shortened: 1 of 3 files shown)


juliantaylor:
@dmehring seems to be one of your changes, can you check the impact of this.
this should probably also have a unittest

dmehring:
@juliantaylor looks reasonable to me, although I agree it should have a unit test as verification, so hopefully @mpokorny will write one.

juliantaylor:
thanks I have merged the commit as 44119fb26346f927dacbf5 (squashed)

github does not send out notifications when you add commits to a branch, if you want a rereview after changes please make a comment in the issue

mpokorny:
@juliantaylor thanks. I had some doubts about a notification being sent out after my additional commits, but decided to wait a while to see what happened before adding more comments.

PR #236, opened 2015-09-16
Addition of SPW error handler and tMSSelection test program

sanbw:


casacore
```diff
@@ -1 +0,0 @@
-.
```
juliantaylor: this removal breaks the build, do you know how to get rid of it?

ms/CMakeLists.txt
```diff
@@ -260,6 +261,7 @@ MSSel/MSSelectableTable.h
 MSSel/MSSelection.h
 MSSel/MSSelectionError.h
 MSSel/MSSelectionErrorHandler.h
+MSSel/MSSSpwErrorHanlder.h
```
juliantaylor: I guess this should be Handler not Hanlder?

ms/MSSel/MSSSpwErrorHandler.h
```diff
@@ -0,0 +1,73 @@
...
+#define MS_MSSPWERRORHANDLER_H
+
+//# Includes
+#include <casacore/casa/aips.h>
+#include <casacore/casa/Logging/LogIO.h>
+#include <casacore/casa/Exceptions/Error.h>
+#include <casacore/ms/MSSel/MSSelectionErrorHandler.h>
+#include <casacore/ms/MSSel/MSSelectionErrorHandler.h>
+#include <vector>
+using namespace std;
```
juliantaylor: the vector and LogIO includes look unnecessary
`using namespace` in a header is a no-go

ms/MSSel/MSSelection.cc
```diff
@@ -366,14 +372,34 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
...
+    	// mssErrHandler_p=MSStateParse::thisMSSErrorHandler=NULL;
+    	// mssErrHandler_p=MSSpwParse::thisMSSpwErrorHandler=NULL;
       }
-    mssErrHandler_p=MSAntennaParse::thisMSAErrorHandler=NULL;
-    mssErrHandler_p=MSStateParse::thisMSSErrorHandler=NULL;
+    MSStateParse::cleanupErrorHandler();
+    MSSpwParse::cleanupErrorHandler();
+    // mssErrHandler_p=MSAntennaParse::thisMSAErrorHandler=NULL;
+    // mssErrHandler_p=MSStateParse::thisMSSErrorHandler=NULL;
+    // mssErrHandler_p=MSSpwParse::thisMSSpwErrorHandler=NULL;
```
juliantaylor: the comments here do not look like they need to be kept

ms/CMakeLists.txt
sanbw: Indeed.  Thanks for finding it (though I don't understand how I was able to build without fixing this typo).  I just pushed the fix.

juliantaylor: make install should produce an error, normal make or make test should work
maybe there also was a wrongly named file in present in your working copy

sanbw:
After fixing the casacore symlink, the only build that fails is clang, CXX11=True on OSX.  I have no idea why.

tammojan:
It shows a segfault in the clang linker ( https://travis-ci.org/casacore/casacore/jobs/80691244#L1949 ). I'll restart the build.

tammojan:
The build succeeded now, the previous failure was probably due to a travis hiccup.

Could you fix the suggestions by @juliantaylor ? Particularly the `using namespace std;` in MSSSpwErrorHandler.h shouldn't be there..

sanbw:
Removed the namespace instruction fromMSSSpwErrorHandler.h.  It need not be there but was also not bothering thing in the software.  For future information for me -- where would it become a problem?

gervandiepen:
Header files (and.tcc files) are intended to be included in other source files, possibly by other packages. You should never use "using namespace" in header (and .tcc) files because it adds all names to the global namespace effectively undoing the namespace idea.
In .cc files you can use it, because they are not included.

gervandiepen:
I assume this change also requires UDFMSCal.cc (in derivedmscal) to be changed. In that file the MSSel error handlers are set; most likely this new SPW error handler needs to be set as well.
UDFMSCal.cc ignores errors reported to the error handler.
E.g., it ignores MSAntennaGram errors like an unknown antenna in something like mscal.baseline("!DE006") because it is not a real error. A selection like mscal.ant1name()!='DE006' isn't an error either if no DE006 is found (and the same if ! is not used).
I assume that the errors reported to SPWErrorHandler fall in the same category; they are not real errors but only unknown identifiers.

juliantaylor:
can the test data be shrunk or be made optional? 8mb for only one test is a bit large

for merging I'd also recommend squashing, the intermediate commits which don't build do not seem worth keeping.

sanbw:
Shrinking the test data is on my to-do list. So I will get to it, but not immediately. I thought 8MB is small, but apparently not.

Squashing intermediates commits is fine.

juliantaylor:
the issue with git is that latter shrinking the file just makes the repository even larger as the old file is never removed and is still part of each checkout, this is different to svn where the history data stays on the remote server

tammojan:
> the issue with git is that latter shrinking the file just makes the repository even larger as the old file is never removed and is still part of each checkout

I think there is some garbage collection of 'unreachable checkouts' after a month or so. Squashing the commit with the large file would make the file unreachable. So it would be nice if the file could be made smaller before merging this (squashed) pull request.

juliantaylor:
@sanbw this branch seems to break the CASA `test_flagdata[test_unknown_intent]` and `test_flagdata[test_CASA_4_0_bug_fix]` test, can you confirm this on your end?

sanbw:
I will try, but it needs more time (which I have less and less of).  I will
let you know when I can find out.

Meanwhile, you could check with Sandra (who needs this change and will know
about flag_data).

Given all the comments for this check-in, I can think for the following two
options going forward:

1. I disable the tMSSelection test program and not check-in test MS till I
figure out how to reduce it's size even further (it is about 8MB right now).

2. We just wait for this pull request till I figure out how to reduce the
size of the MS.

I prefer (1).  And open for alternatives.

Sanjay


On Thu, Sep 17, 2015 at 6:31 AM, Julian Taylor <notifications@github.com>
wrote:

> @sanbw <https://github.com/sanbw> this branch seems to break the CASA
> test_flagdata[test_unknown_intent] and
> test_flagdata[test_CASA_4_0_bug_fix] test, can you confirm this on your
> end?
>
> —
> <https://github.com/casacore/casacore/pull/236#issuecomment-141065374>.[image:
> Web Bug from
> https://github.com/notifications/beacon/AM3Q7VkgFAnbT9Ge2Kxtr0m3x-FWGH5Oks5oyqoHgaJpZM4F-kwZ.gif]
>



-- 
संजय

sanbw:
Ger,

The SPW parser was inconsistent in how it was handling a list of indices
and list of names.  If an elements in a list (or indices or names) does not
find a match, the default error handler will now collect those messages and
post a warning at the end of the parsing cycle indicating the elements that
did not find a match.   If this is not what UDFMSCal needs, then yes -- you
can install the error handler of choice there as well.

AFASICT, this change did pass all of CASACore tests.

As for what is "real error", IMO it is a matter of convention here.  My
model for designing the error handling system for the MSSel module was that
identifiers are "variables" in an expression.  And variables not found in
the symbol table (which are the MS sub-tables here) is an error.  When
using this with my astronomer hat, I find the error message for unmatched
identifiers very very useful  (the alternative being that the application
layers keep quiet about it and later possibly produce garbage when it is
harder to trace back and find the source of the problem...when using the
software for scientific work, mind is also not in the programmer mode).

Over the year, I have found that different applications need different
behaviour of error handling (e.g. pipeline processing vs. interactive
processing).  Hence the solution of pluggable error handlers.

Regards,
sanjay

On Thu, Sep 17, 2015 at 12:34 AM, Ger van Diepen <notifications@github.com>
wrote:

> I assume this change also requires UDFMSCal.cc (in derivedmscal) to be
> changed. In that file the MSSel error handlers are set; most likely this
> new SPW error handler needs to be set as well.
> UDFMSCal.cc ignores errors reported to the error handler.
> E.g., it ignores MSAntennaGram errors like an unknown antenna in something
> like mscal.baseline("!DE006") because it is not a real error. A selection
> like mscal.ant1name()!='DE006' isn't an error either if no DE006 is found
> (and the same if ! is not used).
> I assume that the errors reported to SPWErrorHandler fall in the same
> category; they are not real errors but only unknown identifiers.
>
> —
> <https://github.com/casacore/casacore/pull/236#issuecomment-140981127>.[image:
> Web Bug from
> https://github.com/notifications/beacon/AM3Q7akHpkHXXs-sKdge8ZVNEeYG-smSks5oylZfgaJpZM4F-kwZ.gif]
>



-- 
संजय

sanbw:
Is 600K small enough for the size of the test data set?

tammojan:
Yes, 600k is small enough.

juliantaylor:
how much is it uncompressed? unless working copy size is a concern leaving it uncompressed and let git handle it might be better

sanbw:
12M          mssel_test_small.ms
600K         mssel_test_small.ms.tgz

Uncompressed size is a bit large.  I can do a bit more (and hence take a bit more time) to find out if the MS itself can be written in a more compressed manner.  But can't predict how much more compression can be achieved without gzip compression.

The run-time for the test (including untar'ing) is of the order of 2 sec.  In the interest of time, unless run time for tests is a problem,  I would like to have my mods pull'ed in (so that my fork does not diverge too much from casacore:master).  

I did not quite understand your suggestion earlier, but if there is a way to delete the MS, we can replace it with a smaller uncompressed one.  If it is clear to you how to achieve that, I will rely on that.

gervandiepen:
Printing an SPW warning is fine. 
I would commit the tar-gzipped MS; it is much smaller and it is also easier to handle in cmake/ctest. I assume the tests take little time.
Has the CASA test_flagdata failure been solved? If so, we can do the merge.

tammojan:
Before merging the commits probably need to be cleaned up (rebased). If necessary I could do that.

juliantaylor:
not yet, though that may need resolving in casa (thus this is not backward compatible)

the branch still contains the 8mb file which is too large
also whoever merges this must be a squash merge unless it is cleaned up before.
@sanbw what did you do create such a mess on the branch? possibly some documentation needs tuning. Maybe check with a local git expert on how to clean it up (I'd recommend a local `merge --squash` and the cherry pick that onto a new branch and force push it to the PR)

sanbw:
There is indeed a non-backward compatible change in these mods.  The change is also a fix for a bug which then becomes backwards incompatible.  The flag_data failure has been fixed, but another fix (and the only other fix) in casa code requires the new casacore mods.  So I think that should be done by the casa team once they get the new casacore.

@juliantaylor: :-)  My apologies for the "mess" on the branch (I suppose you are talking of my fork -- hope I have not disrupted others).  The excuse is that I thought I have learned enough of Git/GitHub to do this work, but apparently not!  I will consult @JimNrao and see if we can cleanup the mess (though frankly I do not really know what needs fixing.  I do have many small commits on the branch -- that's my style of development.  Is that the problem?  Too many commits?).

sanbw:
@juliantaylor: Can you tell me specifically what you want cleaned-up on the branch?  I can then work with @JimNrao.

juliantaylor:
many commits are good, but they should be atomic, i.e. every one should ideally compile and pass the tests. This greatly improves regression hunting via bisection.
Also you have a bunch of commits that are seemingly unrelated to your changes, e.g. a merge commit from 2008 and stuff that is already on master. I assume you have at some point merged the origin repository back into your fork. Instead of merging into your fork it is usually better to rebase your fork onto the origin. Otherwise you introduce a bunch of criss-cross merges that don't really provide any value to the history.

what you should still do is replace the 8mb test file with the smaller one. Then I guess @tammojan or me an do a squash merge to dump all your changes into one atomic commit.
If you want to do it yourself, usually you would use rebase, but that seems to cause conflicts with your branch state, so this is how I'd do it:

add the new testfile, commit as usual
```
git fetch git://github.com/casacore/casacore master
git checkout FETCH_HEAD
git merge --squash master
git commit
rev=$(git rev-parse HEAD)
git checkout master
git reset --hard $rev
git push -f git://github.com/sanbw/casacore master:master
```

in future it is also better to not work on the master branch but on a dedicated branch containing only what you want to have merged. Otherwise the master branch of your fork gets messy and needs frequent rebasing.

sanbw:
Thansk @juliantaylor.  

1. I will replace the 8mb test file (I know how to do that! :-)).
2. I will appreciate if one you could do whatever it is you can do now to cleanup the mess.
3. And I will work with @JimNrao here to understand some of the commands you have suggested and see what I can do locally here.  

Do you need to wait for (3) to do (2)?  If not,  after I replace the 8m test data file, you can go ahead.

juliantaylor:
adding the smaller file as a new commit is enough for merging this PR, assuming @gervandiepen and @tammojan are ok on merging it.

sanbw:
Done.  Did all that @juliantaylor requested.  Ready for the merge.

The squashing bit is good and useful (I will try to understand all the commands @juliantaylor sent for this!).

juliantaylor:
hm something went wrong, you probably merged in the origin master branch instead of your own and this also seems to confuse the github gui because its showing diffs that are already in master

sanbw:
Hum.  So what next?  Jim is not in office today -- so can't get him to look.

juliantaylor:
I think this branch is not salvage everything is a merge from something :)
When you aren't experienced with git yet I recommend to abort when git asks you for a merge commit message, it is usually not what you want. Merging is in most workflows only required when going back to the origin repository (casacore/casacore in this case) or really complex feature branches. Any smaller addition branch should by itself not contain merges.

I think I fixed your branch up but it needed some manual patch picking, please do these steps and check everything is still there:

```
git fetch https://github.com/juliantaylor/casacore.git fixed-sanbw
git checkout -f FETCH_HEAD -b fixed-branch
#push the fixed branch to this PR, assuming ssh authentification, this is probably your origin remote
git push -f git@github.com:sanbw/casacore.git fixed-branch:master
```

sanbw:
fixed-branch did not have many of the mods that are in my master branch.

sbhatnag@tara>git branch
* fixed-branch
  master
sbhatnag@tara>git diff fixed-branch..master | grep "diff --git"
diff --git a/.gitignore b/.gitignore
diff --git a/ms/MSSel/MSSelection.cc b/ms/MSSel/MSSelection.cc
diff --git a/ms/MSSel/MSSelection.h b/ms/MSSel/MSSelection.h
diff --git a/ms/MSSel/MSSelectionErrorHandler.cc b/ms/MSSel/MSSelectionErrorHandler.cc
diff --git a/ms/MSSel/MSSelectionErrorHandler.h b/ms/MSSel/MSSelectionErrorHandler.h
diff --git a/ms/MSSel/MSStateParse.cc b/ms/MSSel/MSStateParse.cc
diff --git a/ms/MSSel/MSStateParse.h b/ms/MSSel/MSStateParse.h
diff --git a/ms/MSSel/test/tMSSelection.cc b/ms/MSSel/test/tMSSelection.cc
diff --git a/ms/MSSel/test/tMSSelection.out b/ms/MSSel/test/tMSSelection.out

I copied these from my master branch, examined the some files manually to make sure they have the new code and rebuild from fixed-branch.  All tests passed.  So committed and pushed fixed-branch via

git push -f git@github.com:sanbw/casacore.git fixed-branch:master

Hope this fixes it.

ms/MSSel/test/tMSSelection.run
```diff
@@ -0,0 +1,82 @@
...
+# This script executes the program tMSSelection to test the
+# MSSelection module.
+#
+# $Id$
+#-----------------------------------------------------------------------------
+
+  #CASADEMO=`echo $CASAPATH | awk '{printf("%s/casacore.git/ms/MSSel/test/",$1)}'`+'/'
+  MS=$CASADEMO"mssel_test_small.ms"
+
+  BINPATH='/home/tara2/sanjay/casa_trunk_pristine/trunk/casacore.git/build/ms/MSSel/test/'
```
juliantaylor: this should be removed

ms/MSSel/test/tMSSelection.run
```diff
@@ -0,0 +1,82 @@
...
+
+  if [ ${#CASAPATH} = 0 ]
```
juliantaylor: this is bash specific, set interpreter to bash or use more portable -z

ms/MSSel/test/tMSSelection.run
```diff
@@ -0,0 +1,82 @@
...
+     exit 3
+  fi
+
+  function runCmd(){
+      #echo "Test:$1 "$2; 
+      eval "$2"
+      }
+
+  function getTestMS() {
+      tar zxf $testsrcdir/$MS.tgz
```
juliantaylor: this does not appear to work for out of source builds

ms/MSSel/test/tMSSelection.run
```diff
@@ -0,0 +1,82 @@
...
+
+  BINPATH='/home/tara2/sanjay/casa_trunk_pristine/trunk/casacore.git/build/ms/MSSel/test/'
+
+  if [ ${#CASAPATH} = 0 ]
+  then
+     echo "UNTESTED: tMSSelection.run (CASAPATH not defined)"
+     exit 3
+  fi
+
+  function runCmd(){
```
juliantaylor: function is also bash specific and should be removed

juliantaylor:
does the test even work? from what I can tell its simply skipped in travis

ms/MSSel/test/tMSSelection.run
juliantaylor: actually this should be fine, it were the other issues that threw me off

sanbw: OK. Can you list the issues that need attention?  The CASA specific
CASAPATH check should of course be removed.

juliantaylor: I already did, read the inline comments

ms/MSSel/test/tMSSelection.run
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  function cleanup() {
+      \rm -rf $MS
+  }
+
+  # Get the test MS (un-tar it from $testsrcdir).
+  getTestMS;
+
+  runCmd 1 "$casa_checktool tMSSelection ms=$MS baseline='1,2,3'";
```
juliantaylor: should probably be `./tMSSelection`

ms/MSSel/test/tMSSelection.run
sanbw: There are multiple comments you made, some withdrawing the previous ones!  Anyway...I ,make the fixes.

ms/MSSel/test/tMSSelection.run
sanbw: @juliantaylor: While I am fixing tMSSelection, is there anything else that needs changes before this request can be closed?  Please let me know soonish.  I would like to focus on this and close this ASAP.

juliantaylor:
the test now works for me, thanks
merged as 7e787ead7346fa26fa9886faa9d683aa9fe11b92

PR #237, opened 2015-09-17
Avoid heap allocations for small Unit strings

juliantaylor:
Small allocations are very expensive with glibc, if they are performed
often it can be very worthwhile to replace them with stack based
buffers.
Unit::check is a function which can be called very often by unoptimized
code. To reduce the impact of this avoid the heap allocation for small
strings.


gervandiepen:
Nice optimization. Close

Issue #238, opened 2015-09-19
annoyance with BUILD_PYTHON flag


twillis449:
At the moment the top-level CMakeLists.txt file has BUILD_PYTHON set to NO as the default. This causes  the libcasa_python shared library not to be build. But then when I go to build and install the python-casacore package almost the firrst thing that's looked for and not found is the libcasa_python library! I suggest that either you set BUILD_PYTHON to YES or put a warning in the top level README.md file that this flag should be set to YES if the user intends to build python-casacore.

PR #239, opened 2015-09-19
Build python bindings by default (not for CASA build)

tammojan:
This was triggered by #238. Building the python libraries does not have a disadvantage except for compilation time. Of course it can be manually disabled, and is disabled by default if CASA_BUILD=True.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,8 +57,9 @@ enable_language (Fortran)
 
 set (CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 
+# By default build the Python bindings
+option (BUILD_PYTHON "Build the python bindings" YES)
 # By default build shared libraries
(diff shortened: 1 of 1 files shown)


juliantaylor:
it should be fine to also have it default with CASA_BUILD too, asap needs it which is bundled in our tarballs.

tammojan:
Changed, the PR so that python is always built by default.

JasonRuonanWang:
Another possible disadvantage would be that it requires boost. So on a system where you can't run apt-get or yum as root, and there is no available modules loadable, it is really a lot of extra time and effort...  I know you can choose to disable it anyway.

tammojan:
> Another possible disadvantage would be that it requires boost.

Agreed. Perhaps this can be fixed with some documentation, either in the README on this site, or in the CMake output (like 'Boost not found, try setting -DBUILD_PYTHON=False') ?

PR #242, opened 2015-09-21
CAS-5996, plug memory leak, close open table upon exception

dmehring:
Upon certain failure modes of opening an image, a memory leak would occur and a table would be left open upon an exception being thrown. This modification fixes that.


dmehring:
@gervandiepen @juliantaylor @dpetry Would one of you please merge this change? Thanks!

PR #246, opened 2015-09-24
Added methods to ImageProxy to return objects

dmehring:
Added imageInfoObject() and type() methods to return native casacore objects/types rather than Records/Strings.

### DIFF
--- a/images/Images/ImageProxy.cc
+++ b/images/Images/ImageProxy.cc
@@ -917,6 +917,11 @@ namespace casacore { //# name space casa begins
     return coord;
   }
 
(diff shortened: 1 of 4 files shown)


dmehring:
@gervandiepen ImageProxy mods, as we discussed.

gervandiepen:
Nice.
I only wonder why you return ImageInfo by value and not by const ref. By value is more expensive.

You could add a similar function for the CoordinateSystem.

dmehring:
True, const ref would be better.

dmehring:
@gervandiepen Modified so imageInfoObject() returns const ref, added coordSysObject() which also returns const ref.

PR #248, opened 2015-09-25
CAS-7913 fix segfault on OSX

dmehring:
set::rend() seems to behave differently on linux and OSX. On OSX it was causing a segfault in tMSMetaData. Changed to more portable solution.


juliantaylor:
`--times.end()` does also not work?

this needs a comment in the code on why its done like that

dmehring:
@juliantaylor Actually times.rbegin() is the thing to use. So, I've updated to use that and added a comment in the code as to the special significance of begin() and rbegin() in a std::set

juliantaylor:
ah now it makes more sense, so it was always technically broken, it just happened to work with libstdc++

Issue #243, opened 2015-09-22
Python exception converter for std::exception breaks other library


bmerry:
python/Converters/PycExcp.cc registers an exception-converter for std::exception. This will override exception converters for any Boost.Python-using package imported before casacore. In the case of [spead2](https://github.com/ska-sa/spead2), this causes it to malfunction.

It's also not necessary since Boost.Python will translate any uncaught std::exception to RuntimeError [anyway](https://github.com/boostorg/python/blob/develop/src/errors.cpp#L50)).

gervandiepen:
Thanks for the info. I'll change it. Does spead2 uses its own exception
converter(s)?
How about other converters (e.g. a Python string to a std::string? I guess
they do not interfere with spead2 converters apart from getting a Boost
warning about a doubly registered converter.

bmerry:
spead2 has its own exception translators, but only for classes within the spead2 namespace. It relies on Boost.Python's builtin str<->std::string converters but doesn't define any converters for standard-library C++ classes.

gervandiepen:
Also extended test program with exceptions

PR #250, opened 2015-09-29
Do not convert a normal exception

gervandiepen:
Issue #243
Added exceptions to tests


juliantaylor:
the three commits could be squashed

juliantaylor:
it would also be better when you stopped using the main repo for your branches or at least delete the old stuff, the main repository is full up with lots of junk branches.

gervandiepen:
I thought the revert would completely undo the commit, but I found out too
late that it results in a separate commit.

juliantaylor:
its never to late with git, just do git rebase -i HEAD^^^ and squash it

PR #254, opened 2015-09-30
CAS-5346 also return stats from getHistograms()

dmehring:
Requirement to return stats for each histogram, so overloaded getHistograms() with a version that also returns the stats array. Much of the changes are formatting only changes (eg indent 4 spaces not 3, and elimnate a lot of unnecessary blank lines). Also added tests for new method version.


dmehring:
@gervandiepen Would you please review this pull request; I submitted it several days ago. Thanks.

PR #258, opened 2015-10-26
various coordinate performance improvements

juliantaylor:
coordinate transformations in casa are spending most of their time allocating memory instead of actually computing. This set of patches improves this a situation a bit via a few small changes, see the commits for details.

The largest change is using a small Vector cache to avoid creating new Vectors for each MVPosition and Euler temporary object that are created en mass in the conversion code.
This cache could in theory be moved lower into the stack, e.g. in the allocator itself and the Array object would need to use placement new's to create its Block object, though this is quite tricky to get right so I opted for the simpler change of just caching the objects that are problematic.
A disadvantage is that the small cache is never free'd so it would show up in valgrind, but that can be handled with a suppression file.

### DIFF
--- a/casa/Quanta/Euler.cc
+++ b/casa/Quanta/Euler.cc
@@ -38,14 +38,44 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
 // Euler class
 
+// simplistic Vector(3) cache to reduce allocation overhead for temporaries
+#if defined(AIPS_CXX11) && !defined(__APPLE__)
+thread_local size_t Euler::available = 0;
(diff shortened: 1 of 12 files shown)


dpetry:
This looks great, Julian. Thanks.
The unit tests are quite detailed. So I think there is little risk in these changes.
But we should keep an eye on test failures in the CASA regression tests in the next few days.

juliantaylor:
I run the casa regression tests on casacore PRs already, though I didn't find the time to integrate the result reporting with github yet.
This branch tested fine (but there is an unrelated failure in test_regrid since earlier today)

PR #259, opened 2015-11-02
Optimization of MSMetaData::getPointingDirection (CAS-7643).

naojKanaS:
Two changes are made to optimize MSMetaData::getPointingDirection (CAS-7643).
1. use ScalarColumn::get to access a cell instead of generating a vector of column values in all rows by ScalarColumn::getColumn. The latter causes severe performance degrade.
2. omitt redundant operation for rows of auto-correlation.
See https://bugs.nrao.edu/browse/CAS-7643 for more details.

### DIFF
--- a/ms/MSOper/MSMetaData.cc
+++ b/ms/MSOper/MSMetaData.cc
@@ -3878,17 +3878,23 @@ std::pair<MDirection, MDirection> MSMetaData::getPointingDirection(
 		row >= this->nRows(),
 		"Row number exceeds number of rows in the MS"
 	);
-	SHARED_PTR<Vector<Int> > ant1, ant2;
-	_getAntennas(ant1, ant2);
-	antenna1 = (*ant1)[row];
-	antenna2 = (*ant2)[row];
-	time = (*_getTimes())[row];
+	// KS note: 2015/10/30
+	// getColum method (also used in _getAntennas and _getTimes) is
+	// too expesive for the current purpose of getting single cell.
+	String ant1ColName = MeasurementSet::columnName(MSMainEnums::ANTENNA1);
+	String ant2ColName = MeasurementSet::columnName(MSMainEnums::ANTENNA2);
+	antenna1 = ROScalarColumn<Int>(*_ms, ant1ColName).get(row);
+	antenna2 = ROScalarColumn<Int>(*_ms, ant2ColName).get(row);
+	bool autocorr = (antenna1==antenna2);
+	String timeColName = MeasurementSet::columnName(MSMainEnums::TIME);
+	time = ScalarColumn<Double>(*_ms, timeColName).get(row);
 	ROMSPointingColumns pCols(_ms->pointing());
 	Int pidx1, pidx2;
 	pidx1 = pCols.pointingIndex(antenna1, time, initialguess);
-	pidx2 = pCols.pointingIndex(antenna2, time, initialguess);
+	if (autocorr) pidx2 = pidx1;
+	else pidx2 = pCols.pointingIndex(antenna2, time, initialguess);
 	String intervalColName = MeasurementSet::columnName(MSMainEnums::INTERVAL);
-	Double interval = ScalarColumn<Double>(*_ms, intervalColName).getColumn()[row];
+	Double interval = ScalarColumn<Double>(*_ms, intervalColName).get(row);
 	MDirection dir1, dir2;
 	if (!interpolate || interval >= pCols.interval()(pidx1)) {
 		dir1 = pCols.directionMeas(pidx1);
@@ -3896,7 +3902,10 @@ std::pair<MDirection, MDirection> MSMetaData::getPointingDirection(
 	else {
 		dir1 = _getInterpolatedDirection(pCols, pidx1, time);
 	}
-	if (!interpolate || interval >= pCols.interval()(pidx2)) {
+	if (autocorr) {
+	  dir2 = dir1;
+	}
+	else if (!interpolate || interval >= pCols.interval()(pidx2)) {
 		dir2 = pCols.directionMeas(pidx2);
 	}
 	else {


ms/MSOper/MSMetaData.cc
```diff
@@ -3878,17 +3878,23 @@ std::pair<MDirection, MDirection> MSMetaData::getPointingDirection(
 		row >= this->nRows(),
 		"Row number exceeds number of rows in the MS"
 	);
-	SHARED_PTR<Vector<Int> > ant1, ant2;
-	_getAntennas(ant1, ant2);
-	antenna1 = (*ant1)[row];
-	antenna2 = (*ant2)[row];
-	time = (*_getTimes())[row];
+	// KS note: 2015/10/30
```
juliantaylor: this comment is not necessary as the code it refers to is gone

ms/MSOper/MSMetaData.cc
```diff
@@ -3878,17 +3878,23 @@ std::pair<MDirection, MDirection> MSMetaData::getPointingDirection(
...
+	antenna1 = ROScalarColumn<Int>(*_ms, ant1ColName).get(row);
+	antenna2 = ROScalarColumn<Int>(*_ms, ant2ColName).get(row);
+	bool autocorr = (antenna1==antenna2);
+	String timeColName = MeasurementSet::columnName(MSMainEnums::TIME);
+	time = ScalarColumn<Double>(*_ms, timeColName).get(row);
 	ROMSPointingColumns pCols(_ms->pointing());
 	Int pidx1, pidx2;
 	pidx1 = pCols.pointingIndex(antenna1, time, initialguess);
-	pidx2 = pCols.pointingIndex(antenna2, time, initialguess);
+	if (autocorr) pidx2 = pidx1;
```
juliantaylor: one line if is inconsistent with the rest of the file which uses
    if {
      //
    }

ms/MSOper/MSMetaData.cc
```diff
@@ -3896,7 +3902,10 @@ std::pair<MDirection, MDirection> MSMetaData::getPointingDirection(
 	else {
 		dir1 = _getInterpolatedDirection(pCols, pidx1, time);
 	}
-	if (!interpolate || interval >= pCols.interval()(pidx2)) {
+	if (autocorr) {
+	  dir2 = dir1;
```
juliantaylor: wrong indentation compared to the rest of the file

juliantaylor:
looks very good to me besides some very minor style nitpicks

ms/MSOper/MSMetaData.cc
```diff
@@ -3878,17 +3878,23 @@ std::pair<MDirection, MDirection> MSMetaData::getPointingDirection(
...
-	SHARED_PTR<Vector<Int> > ant1, ant2;
-	_getAntennas(ant1, ant2);
-	antenna1 = (*ant1)[row];
-	antenna2 = (*ant2)[row];
-	time = (*_getTimes())[row];
+	// KS note: 2015/10/30
+	// getColum method (also used in _getAntennas and _getTimes) is
+	// too expesive for the current purpose of getting single cell.
+	String ant1ColName = MeasurementSet::columnName(MSMainEnums::ANTENNA1);
+	String ant2ColName = MeasurementSet::columnName(MSMainEnums::ANTENNA2);
```
juliantaylor: probably its not significant but I guess these could be const references instead of copies.

naojKanaS:
Thank you very much Julian for the quick response!
I'll work on your comments when I get back to office on Wed.

juliantaylor:
the issues are only stylistic, if it is important we can also merge now

naojKanaS:
Thanks Julian. If it's possible to merge now, it would be wonderful. It's because the performance gain by the change is huge (25hrs -> less than 2min). And the issue is preventing commissioning of a new obsmode in Cycle 4.
In any case, I'll update the code based on your comment on Wed.

juliantaylor:
I wonder if the non-ascii commit author string will screw things up, svn checkout just displays it as `...`
well only one way to find out :)

naojKanaS:
I don't know. This one was my first attempt to use github and I obviously screwed up. git command automatically  set my name somewhere in my computer. I tried to change it when I add comment for the commit but the change was ignored. I need to figure out how to make it in English.

Anyway, thank you very much for merging the commit. I hope it won't cause any issue....

juliantaylor:
the name and email used for commits can be configured with:

    git config --global user.name "my name"
    git config --global user.email "my@mail"

that puts the configuration into ~/.gitconfig
existing top commit can be changed with:

    git commit --amend --author="myname <my@mail>"

or

    git commit --amend --reset-author

ms/MSOper/MSMetaData.cc
naojKanaS: Julian, I believe what you mean by this is, something like below. Is it correct?
I see there are several other calls to MeasurementSet::columnName  in other functions of MSMetaData class, which do not use const reference.
Do you think I'd better modify those too?

```javascript
-       String ant1ColName = MeasurementSet::columnName(MSMainEnums::ANTENNA1);
-       String ant2ColName = MeasurementSet::columnName(MSMainEnums::ANTENNA2);
+       const String& ant1ColName = MeasurementSet::columnName(MSMainEnums::ANTENNA1);
+       const String& ant2ColName = MeasurementSet::columnName(MSMainEnums::ANTENNA2);
        antenna1 = ROScalarColumn<Int>(*_ms, ant1ColName).get(row);
        antenna2 = ROScalarColumn<Int>(*_ms, ant2ColName).get(row);
```

ms/MSOper/MSMetaData.cc
naojKanaS: done in #262.

ms/MSOper/MSMetaData.cc
naojKanaS: done in #262

ms/MSOper/MSMetaData.cc
naojKanaS: done in #262.

ms/MSOper/MSMetaData.cc
juliantaylor: yes this is what I meant. Indeed these copies appear all over the place in this file (though sometimes they are static), though as long as they don't show up as relevant in a performance profile its probably not necessary to change them all. I'll leave it up to you whether you consider it worthwhile to do.

PR #261, opened 2015-11-03
CAS-7980 control the amount of contiguous memory required

dmehring:
This was trying to load all pixel values into a contigous slice of memory, and failing for large values. So, I've added code to make the required memory chunks no larger than 1024*1024 pixels.


juliantaylor:
I'd prefer a comment where this number comes from, but I'll just assume its well chosen.

Issue #265, opened 2015-11-19
tTableGramGroupAggr test fails using Xcode 7


vsuorant:
tTableGramGroupAggr segfaults when compiling with Xcode 7 and optimization levels above -O0.

This doesn't seem to be an issue with Xcode 6.

vsuorant:
Backtrace

(gdb) bt
#0  0x0000000100306fc9 in std::__1::vector<casa::CountedPtr<casa::TableExprGroupFuncSet>, std::__1::allocator<casa::CountedPtr<casa::TableExprGroupFuncSet> > > casa::TableParseSelect::doGroupByAggrSingleKey<long long>(std::__1::vector<casa::TableExprNodeRep*, std::__1::allocator<casa::TableExprNodeRep*> > const&) () from /Users/vsuorant/tmp/casacore/casacore/build/tables/libcasa_tables.2.dylib
#1  0x00000001002fde34 in casa::TableParseSelect::doGroupByAggr (this=0x1, aggrNodes=...) at /Users/vsuorant/tmp/casacore/casacore/tables/TaQL/TableParse.cc:2422
#2  0x00000001002fd805 in casa::TableParseSelect::doGroupby (this=0x101004a00, showTimings=false, aggrNodes=..., groupAggrUsed=<optimized out>) at /Users/vsuorant/tmp/casacore/casacore/tables/TaQL/TableParse.cc:2287
#3  0x00000001002f6ab1 in casa::TableParseSelect::execute (this=0x1, showTimings=false, setInGiving=<optimized out>, mustSelect=<optimized out>, maxRow=<optimized out>, doTracing=false) at /Users/vsuorant/tmp/casacore/casacore/tables/TaQL/TableParse.cc:3189
#4  0x00000001002d8000 in casa::TaQLNodeHandler::visitSelectNode (this=0x101004a00, node=...) at /Users/vsuorant/tmp/casacore/casacore/tables/TaQL/TaQLNodeHandler.cc:499
#5  0x00000001002d182b in casa::TaQLSelectNodeRep::visit (this=0x101805ca0, visitor=...) at /Users/vsuorant/tmp/casacore/casacore/tables/TaQL/TaQLNodeDer.cc:965
#6  0x00000001002d4b27 in visit (this=<optimized out>, visitor=...) at /Users/vsuorant/tmp/casacore/casacore/casacore/tables/TaQL/TaQLNode.h:132
[...]

PR #266, opened 2015-11-19
Fixing issue 265

vsuorant:


juliantaylor:
I always though that was intentional given its the only compiler warning casacore produces. Interesting it actually causes issues somehow.
A better commit message would be good explaining how it is related to the issue (which can be automatically closed via `Closes gh-265`)

vsuorant:
I updated the commit message. I think Ger needs to take a look at this before this gets merged to master. It fixes the issue for me and doesn't seem to cause any harm, but I don't know if this is logically the correct thing to do.

gervandiepen:
I'll look into it. The old code is clearly wrong, thus the change is an improvement, but I have to make sure it works fine for a String or Bool key.

gervandiepen:
It is only used for Int64 and Double, thus the fix is fine.

PR #267, opened 2015-11-23
Changed signature of uvrdvr_c to take a "const char *" 

JimNrao:
Changed signature of uvrdvr_c to take a "const char *" rather than a char *; since the parameter is read-only, that's the appropriate type.


juliantaylor:
thanks

PR #268, opened 2015-11-29
Fix endian detection for ARM

bmerry:
This is a belt-and-braces fix: if `__BYTE_ORDER__` is set (defined by GCC and
Clang), then that is used instead of trying to guess based on a list of known architectures. The architecture list is still there as a fallback, and I've added armel and aarch64el to the list of little-endian architectures.

I've also added aarch64 to the list of 64-bit architectures.

I've tested that this works on an armel system, which was previously unable to
read a measurement set due to the endianness being mis-detected. The aarch64
support is untested.

### DIFF
--- a/casa/aipsenv.h
+++ b/casa/aipsenv.h
@@ -175,7 +175,7 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
 #define AIPS_CRAY_CATAMOUNT
 #endif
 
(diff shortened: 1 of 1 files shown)


tammojan:
Thanks @bmerry! Using the compiler provided byte order is definitely better than relying on known architectures.

The fallback can later be improved by using cmake to detect the endianness, I'll add that in the future (it's a bit less important now that the compiler gets it right).

Issue #249, opened 2015-09-29
Building on ARM architecture


tammojan:
I compiled casacore on a Raspberry Pi, running Raspbian (Debian 7.8). Everything compiles fine, although it does run out of memory sometimes when specifying `make -j4`.

The following tests fail however:
[...]

PR #269, opened 2015-11-29
Add -fsigned-char to compiler arguments

bmerry:
casacore assumes in several places that char is signed, even though C/C++
does not guarantee it. Passing -fsigned-char forces the compiler to treat it as
signed even on architectures (such as ARM) where it is normally unsigned.

This is potentially risky if inlined header code in any included library
assumes that char is unsigned, but this change makes all the unit tests pass
so it is almost certainly better than not using it.

See #249.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,7 +112,7 @@ endif (ENABLE_SHARED)
 # Casacore uses longlong, so no warnings for it.
 # Clang gives warning on bison generated code; disable unneeded-internal-declaration.
 if (NOT CMAKE_CXX_FLAGS)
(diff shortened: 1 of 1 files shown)


tammojan:
I'm not sure about this change. The proper fix would of course be to fix the places where casacore  assumes signed char.

@gervandiepen , could you either merge this or give an estimate on how involved it would be to fix it properly?

(I had to add this flag myself when building on Raspberry Pi, so at the very least it's something that needs to be documented.)

bmerry:
Certainly fixing the code to not assume this would be better. This is a stop-gap fix. Assuming the unit test coverage is good, it seems to be better to have this in than not have it (makes several tests pass, doesn't break any).

gervandiepen:
Thanks Bruce.  think that for now the change is fine, but I agree with Tammo Jan one time it should be fixed properly. I'll add it to my todo list. I've merged the pull request.

PR #271, opened 2015-12-02
allow the share/casacore/data path to be used relative to CASAPATH in…

schiebel:
… addition to the hardcoded CASADATA path


juliantaylor:
looks straightforward enough
searched should probably just be a vector using push_back though that can be done another time

Issue #270, opened 2015-12-01
Add image functionality to get image meta data in a memory buffer


gervandiepen:
To be able to stream images to another process, it is necessary to get all image meta data in a single buffer. It is best to do that as a single string in FITS header format.

PR #272, opened 2015-12-03
make it possible to create a FITS header in memory

gervandiepen:
Solving issue #270


tammojan:
@gervandiepen , can you assign a volunteer to review this?

dpetry:
Thanks. I was already looking at it but wanted to wait until all checks had finished.
And yesterday I was not at work.

images/Images/ImageFITS2Converter.cc
```diff
@@ -804,13 +808,13 @@ Bool ImageFITSConverter::ImageToFITSOut(
 	// and make Stokes the very last if requested
 	//
 	IPosition shape = image.shape();
-	IPosition newShape = shape;
+        fhi.newShape = shape;
```
tammojan: Should have been indented with tabs here, like the rest of the file. Sorry for nitpicking..

images/Images/ImageFITS2Converter.cc
```diff
@@ -842,52 +846,51 @@ Bool ImageFITSConverter::ImageToFITSOut(
...
+				fhi.needNonOptimalCursor=True;
 			}
 		}
 
 	}
 	//
-	Bool applyMask = False;
-	Array<Bool>* pMask = 0;
+	fhi.applyMask = False;
+        fhi.pMask = 0;
```
tammojan: Same here, should have been tabs.

dpetry:
Ger, so you introduced the new struct ImageFITSHeaderInfo in roder to encapsulate all information
pertaining to the header such that this can be kept in memory.
This led to several interface changes.
Have you tried to build CASA with this?
If not, I should probably try that before we merge.

gervandiepen:
The original ImageToFITS function has been split into 2 parts with the
ImageFITSHeaderInfo struct as the link between the two. Thus the original
API has not changed.
But it is certainly a good idea to build CASA before doing the merge.

I'll take a look at the indentation.

juliantaylor:
casa builds and tests fine with this change

dpetry:
You also mean the NRAO CASA including code, asap, and gcwrap, right?

juliantaylor:
yes

dpetry:
Then I think, pending the things Tammo found, this is good to go.

dpetry:
Ger, let me know when you think the cosmetic parts are OK.

gervandiepen:
Re-indented all of ImageFITS2Converter.tcc to get it right.

PR #276, opened 2015-12-11
fix regression from gh-272

juliantaylor:
gh-272 introduced a regression in BITPIX 16 storing as the min and max pixel values where not progated from the fits header function to the fits output function.

also use a CountedPtr for the header structure to avoid leaking the mask from the header structure (like the one in the testsuite)


dpetry:
Julian, you verified that the formerly failing fits-import-export regression passes with your fix, right?

juliantaylor:
yes that test and imagetest work with this change, I'll also see if I can add these tests to the job that checks github pull requests.

Issue #273, opened 2015-12-04
Make measures search path smaller


tammojan:
I think the current list of directories being searched is too large. This could lead to confusion if multiple instances of measures data are installed.

I propose a small refactor, where only to paths are searched:
* A custom path, set in `.casarc`
* A default path, set at compile time (`cmake -DCASADATA=...`)

To acommodate for a directory relative to casaroot, we could accept some syntax within the pathnames, for example `%CASAROOT%` and `%CASAHOME%` in both the custom and default path.

PR #277, opened 2015-12-15
Meake measures data search path smaller

tammojan:
Philosophy: only two paths are searched.
1. A user path, set in .casarc
2. A system path, set at compile time (actually cmake time)

This change removes hardcoded paths like /share/casacore/data. For compatibility, I kept the special paths casahome() and casaroot(), which can be specified at compile time with %CASAHOME% and %CASAROOT% and will be expanded at run time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ build/
 *table.lock
 *CMakeCache.txt
 *CMakeFiles
+.project
+.cproject
+.settings
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -339,19 +339,13 @@ if(USE_THREADS)
 endif(USE_THREADS)
 
 # Check whether DATA_DIR contains data (only if measures is build)
-# Warning: this check is not exhaustive
+if (NOT DATA_DIR)
+   set (DATA_DIR ${CMAKE_INSTALL_PREFIX}/share/data/casacore)
+endif (NOT DATA_DIR)
 if (";${_modules};" MATCHES ";measures;")
-    if (DATA_DIR)
-        get_filename_component(ABS_DATA_DIR "${DATA_DIR}" ABSOLUTE)
-        if (NOT EXISTS "${ABS_DATA_DIR}/ephemerides/"
-            AND NOT EXISTS "${ABS_DATA_DIR}/share/casacore/data/ephemerides")
-            message(FATAL_ERROR "No ephemerides data found in the specified DATA_DIR")
-        endif()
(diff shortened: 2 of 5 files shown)


tammojan:
This would fix #273. To repeat: the issue is that if you would have more than one instance of measures data installed, it would be unclear to users which data instance would be picked. Making the search path much smaller would make it clearer.

tammojan:
This effectively reverts PR #271 by @schiebel . His use case would be covered by specifying `-DDATA_DIR=%CASAROOT%/share/casacore/data/` at compile time.

@schiebel, could you comment on this PR?

@dpetry I think you should review this.

dpetry:
Looking at it now. But we need to wait for Darrell's opinion.

schiebel:
CASA distributions can be relocated at will by users. For this reason, we require dynamic paths which could change from one startup to the next.

tammojan:
> CASA distributions can be relocated at will by users. For this reason, we require dynamic paths which could change from one startup to the next.

Agreed. So that is why I put the option %CASAROOT% in. That would solve this issue, wouldn't it?

schiebel:
CASA is distributed as a tar file containing a binary distribution. The users download the tar file and unpacks it in an arbitrary directory (without compiling anything). At runtime, we discover the directory, set CASAPATH, and then exec the binary which actually runs CASA. So any solution must permit finding the data et al. directories based upon an environment variable.

tammojan:
> So any solution must permit finding the data et al. directories based upon an environment variable.

I claim that this solution does permit that. The packager (you) needs to set `-DDATA_DIR=%CASAROOT%/share/casacore/data` at compile time. The `%CASAROOT%` is replaced by the result of `casaRoot()` at run time, which should find the current path from where casa is run (I didn't change this code).

schiebel:
OK, got it, and I'm satisfied. I took the comment that this is %CASAROOT% is filled from .casarc too literally.

schiebel:
Hi Dirk,

I just created a pull request #278 to fix the CASA builds WRT this change,
could you review it (after tests complete), please.

thanks,
Darrell

vsuorant:
Hello,

This caused trouble with our Jenkins build which runs unit tests before the application is packaged. I fixed this by setting the DATA_DIR to a fixed location, but this raised a question about running these tests a part of our nightly builds (which does the packaging). If I set the DATA_DIR at the compile time, is the %CASAROOT% going to work at runtime? If not, what is the proper way of setting the DATA_DIR?

--Ville

tammojan:
@vsuarant this should be solved by #278, which should be merged. If @dpetry doesn't do that soon, I'll do it :)

PR #280, opened 2015-12-18
fix some python3 issues, enable python3 travis

gijzelaerr:


cmake/FindNUMPY.cmake
```diff
@@ -39,11 +39,7 @@
 #============================================================================
 
 # Finding NumPy involves calling the Python interpreter
-if(NumPy_FIND_REQUIRED)
-    find_package(PythonInterp REQUIRED)
-else()
-    find_package(PythonInterp)
-endif()
+find_package(PythonInterp REQUIRED)
```
tammojan: Why did you change this? If FindNumPy is now called without REQUIRED, it will fail if there is no interpreter, right?

gijzelaerr: I don't see how you can find a Numpy installation without a Python interpreter.

gijzelaerr: maybe you are right there, you don't want to fail when numpy is not a requirement.

PR #281, opened 2015-12-21
Add calls to FITS and miriad registration open functions 

dmehring:
added registration of FITS and MIRIAD open image calls to openImage() so they don't have to be sprinkled at various places in higher level classes.


dmehring:
@gervandiepen Hi Ger, modified ImageUtilities::openImage() as we discussed. Please review. Thanks!

PR #284, opened 2016-01-08
Fix taql if readline is not available

tammojan:
This fixes the following compile error (with CXX11=True, but probably it's a bug everywhere). Probably just a typo.

```
[ 37%] Building CXX object tables/apps/CMakeFiles/taql.dir/taql.cc.o
/root/lofar/casacore/src/tables/apps/taql.cc: In function 'bool readLine(casa::String&, const casa::String&)':
/root/lofar/casacore/src/tables/apps/taql.cc:80:10: error: cannot convert 'std::istream {aka std::basic_istream<char>}' to 'bool' in return
   return cin;
          ^
tables/apps/CMakeFiles/taql.dir/build.make:62: recipe for target 'tables/apps/CMakeFiles/taql.dir/taql.cc.o' failed
make[2]: *** [tables/apps/CMakeFiles/taql.dir/taql.cc.o] Error 1
CMakeFiles/Makefile2:7731: recipe for target 'tables/apps/CMakeFiles/taql.dir/all' failed
make[1]: *** [tables/apps/CMakeFiles/taql.dir/all] Error 2
Makefile:160: recipe for target 'all' failed
make: *** [all] Error 2
```

### DIFF
--- a/tables/apps/taql.cc
+++ b/tables/apps/taql.cc
@@ -77,7 +77,7 @@ bool readLine (String& line, const String& prompt)
 {
   if (!prompt.empty()) cerr << prompt;
   getline (cin, line);
-  return cin;
+  return cin.good();
 }
 #endif
 


gervandiepen:
This fix is not correct.
In the old STL an istream is convertible to void* and gives the status (it was not a typo). The fix should be:
    return cin.good();
which also works fine for C++98.

tammojan:
I changed the fix, it's being processed by travis now (but Travis has readline, so the fix itself is not actually tested).

PR #286, opened 2016-01-09
Added MS feed selection to MSSelection

pford:
Added code to enable MS feed selection in MSSelection including flex scanner and bison parser files.  Feed selection is a simplified form of antenna selection, as feeds are indices only and not names and there are no stations to select.


dpetry:
Hi Pam, the build with your changes has failed. Please have a look.

juliantaylor:
@pford if git asks you to merge you are doing something wrong in 9 out of 10 cases
please try to avoid merges like here in future

PR #289, opened 2016-01-21
Write FEED::RECEPTOR_ANGLE to AN::POLAA/POLAB

dmehring:
Write receptor angles to uvfits file + minor style changes (eg bracket if blocks, un-cuddle else if blcoks) + remove commented code block.

### DIFF
--- a/msfits/MSFits/MSFitsOutput.cc
+++ b/msfits/MSFits/MSFitsOutput.cc
@@ -1734,10 +1734,9 @@ Bool MSFitsOutput::writeAN(FitsOutput *output, const MeasurementSet &ms,
         ROMSFeedColumns feedCols(feedTable);
         ArrayColumn<String> inpoltype(feedCols.polarizationType());
         ScalarColumn<Int> inantid(feedCols.antennaId());
+        ArrayQuantColumn<Double> receptorAngle(feedCols.receptorAngleQuant());
 
-        FITSTableWriter writer(output, desc, strlengths, nant, header, units,
-                False);
-
+        FITSTableWriter writer(output, desc, strlengths, nant, header, units, False);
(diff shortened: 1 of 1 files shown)


msfits/MSFits/MSFitsOutput.cc
```diff
@@ -1756,42 +1755,10 @@ Bool MSFitsOutput::writeAN(FitsOutput *output, const MeasurementSet &ms,
         // Set the ones we're not going to change once
         *orbparm = 0.0;
         *poltya = " ";
-        *polaa = 0.0;
```
juliantaylor: what will it now write when it can't find a polarization type? I assume this should not happen, but as the function just warns and continues would that result in uninitialized data?

dmehring:
Well, it looks to me like it shouldn't happen, ie if it does, the MS is invalid to start with, so GIGO. I can put the 

*polaa = 0.0 

etc initializers back in, but since the MS cell array's length is equal to the number of polarizations, this just seems like extra, do-nothing code that impacts performance for no reason. I could also have an exception thrown if no polarizations are found, which probably makes sense. What would you prefer?

juliantaylor:
hm an issue with an exception is that you may want to write out the invalid ms for debugging and it might break code assuming this passes on invalid ms

I think initializing it is best, it doesn't look like that would be slow and you don't need valgrind to spot uninitialized use as you have the warning printed to the log.

dmehring:
Done.

juliantaylor:
thanks
to make reviews easier please split cosmetic changes (comment removal, braces, ...) from functional changes into separate commits.
this is very easy to do with git, use `git add -p` to selectively add changes to a commit, retroactively that can be done with `git reset HEAD^; git add -p`

PR #291, opened 2016-01-27
CAS-5818 only use AN::DIAMETER values if they are valid

dmehring:
AIPS writes zeros to the AN::DISH_DIAMETER table, so fall back to determining diameters the usual way if the AN table values are obviously invalid.

Also a bit of refactoring, used git add -p to add changes in stages to make them easier to review. The functionality change represents about 10 lines of code.

### DIFF
--- a/msfits/MSFits/MSFitsInput.cc
+++ b/msfits/MSFits/MSFitsInput.cc
@@ -297,147 +297,147 @@ MSFitsInput::MSFitsInput(const String& msFile, const String& fitsFile,
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


dpetry:
Hi Dave,
what kind of tests did you run to confirm that this works?

dmehring:
I ran tests on the data set provided in the JIRA ticket to confirm it works. Assuming the submitter of the ticket does not have objections, I will place the dataset in the data repos and write gcwrap unit tests for it.

Issue #292, opened 2016-01-28
FITSImage fails for 64-bit


gervandiepen:
FITSImage fails to handle 64-bit values properly, because it returns TpDouble as data type, while FITSImage is derived from ImageInterface<Float>.
It can easily be fixed by returning TpFloat, but at some later stage it should be fixed properly by making FITSImage a template class.

PR #293, opened 2016-01-28
Always return TpFloat as data type

gervandiepen:
Issue #292 Always return TpFloat as data type
close

### DIFF
--- a/images/Images/FITSImage.h
+++ b/images/Images/FITSImage.h
@@ -219,8 +219,11 @@ public:
   // Check class invariants.
   virtual Bool ok() const;
 
-  // Return the (internal) data type (TpFloat or TpShort).
-  DataType dataType () const
+  // Return the data type (TpFloat).
+  virtual DataType dataType() const;
+
+  // Return the (internal) data type.
+  DataType internalDataType() const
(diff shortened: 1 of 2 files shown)


images/Images/FITSImage.h
```diff
@@ -219,8 +219,11 @@ class FITSImage: public ImageInterface<Float>
   // Check class invariants.
   virtual Bool ok() const;
 
-  // Return the (internal) data type (TpFloat or TpShort).
-  DataType dataType () const
+  // Return the data type (TpFloat).
+  virtual DataType dataType() const;
+
+  // Return the (internal) data type.
+  DataType internalDataType() const
```
juliantaylor: this sounds like it should be private or it would be good to have some documentation what the difference is to the datatype

juliantaylor:
a slightly more length commit message would be appreciated, also the issue is a bit vague. Its not obvious (to me) what this has to do with 64 bit values.
But the change seems to make sense as the underlying template class is of type float.

I guess you also wanted "closes #292" in the commit message
see here for details autoclose syntax:
https://help.github.com/articles/closing-issues-via-commit-messages/

PR #294, opened 2016-01-29
CAS-3660 bug fix for gaps in AN table, better error handling

dmehring:
Cleaned up logic in the method to write the ANTENNA table including defect fixes and better error handling and logging. Refactored common code into a method of its own. Other refactoring. Since I'm the only one working on these files now, I'm using the industry standard _ prefix for private fields and methods. I will slowly be migrating existing fields to this format as I continue refactoring this code.

### DIFF
--- a/msfits/MSFits/MSFitsInput.cc
+++ b/msfits/MSFits/MSFitsInput.cc
@@ -237,7 +237,7 @@ void MSPrimaryTableHolder::detach() {
 //------------------------------------------------------------
 MSFitsInput::MSFitsInput(const String& msFile, const String& fitsFile,
         const Bool useNewStyle) :
-    infile_p(0), msc_p(0), restfreq_p(0), addSourceTable_p(False), itsLog(LogOrigin(
+    infile_p(0), msc_p(0), _uniqueAnts(), _nAntRow(0), restfreq_p(0), addSourceTable_p(False), itsLog(LogOrigin(
             "MSFitsInput", "MSFitsInput")), newNameStyle(useNewStyle), _msCreated(False) {
     // First, lets verify that fitsfile exists and that it appears to be a
     // FITS file.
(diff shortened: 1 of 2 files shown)


juliantaylor:
_prefix is only a "industry standard" in some languages like python.
In C underscore followed by capital letter are reserved keywords of the language.

for c++ it doesn't matter due to namespaces, but the casa(core?) convention for private fields is a `_p` suffix

juliantaylor:
oh `_p` is just members not functions

dmehring:
So ... you're not going to merge it because of that? Note there are private fields (not created by me) in this same file that do not follow such a pattern.

Can you point me to the document that defines casacore code standards please? Thanks.

juliantaylor:
I am just pointing out that you are introducing inconsistencies, even within the same file
its not much of a document but there is this: https://safe.nrao.edu/wiki/bin/view/Software/CASA/CasaCoreInfo#Coding_Style
which says either `_p` suffix or `its` prefix.
I don't mind either approach, but it would be good to not mix the two in the same subsystem.

Also please split bug fixes from pure style changes in different commits to make reviewing easier.

dmehring:
I'm going to be working on these files a lot, so my goal is that in the end they will be consistent. Apologies if that wasn't clear in my commit comment. I can't make all the style changes in one go in a several thousand line file, and it would be a nightmare to review anyway if I did. 

I did two git add -p to try to separate things to make it easier to review as you had requested previously. Unfortunately, it is simply not effecient to purely seperate style changes from functionality changes; sometimes if I don't make style changes simultaneously with functional changes, style changes simply won't get made.

juliantaylor:
fair enough, the changes do look like the improve the code a lot.
thanks

dmehring:
Thanks!

PR #299, opened 2016-02-05
Rework default DATA_DIR

alxames:
Swap order of `data/` and `casacore/` in the default DATA_DIR, such that all casacore files under `share/` are installed and expected under `share/casacore/`. This brings it in line with where the casacore valgrind suppression file is installed, i.e. also under `share/casacore/`.


tammojan:
This does not affect the CASA build. @gijzelaerr , does this hurt your packages? I think the changes makes sense.

gijzelaerr:
package should not be affected.

PR #301, opened 2016-02-05
CAS-7696: Turn off extended filename support if specified file exists

dmehring:
Before this change, importing uvfits files with certain names would fail because FITS extended filename support was enabled. Now we first check the file name, and if that file exists, this support is disabled to allow import of those files which would otherwise fail if such support was enabled.

### DIFF
--- a/fits/FITS/blockio.cc
+++ b/fits/FITS/blockio.cc
@@ -75,7 +75,7 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
 	// using cfitsio of NASA to open fits file for writting and reading.
 	int l_status, iomode;
 	File myfile(m_filename);
-	if (myfile.exists() && myfile.isReadable()) {
+	if (myfile.isReadable()) {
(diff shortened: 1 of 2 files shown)


fits/FITS/blockio.cc
```diff
@@ -73,7 +74,14 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
 	strcpy(m_filename,f);
 	// using cfitsio of NASA to open fits file for writting and reading.
 	int l_status, iomode;
-	l_status = 0; 
+	File myfile(m_filename);
+	if (myfile.exists() && myfile.isReadable()) {
```
juliantaylor: `isReadable` also checks existance, so the existance check is unnecessary (but also harmless)

dmehring:
removed exists() check

juliantaylor:
thanks

juliantaylor:
for reference, an example filename is: `name+434` which is interpreted by cfitsio as open extension `434` of file `name`

PR #302, opened 2016-02-05
Compressing repeated, unwieldy template definitions to make code more…

dmehring:
… readable. No functionality changes. Just macro substitutions to make code more readable.

### DIFF
--- a/scimath/Mathematics/ClassicalStatistics.h
+++ b/scimath/Mathematics/ClassicalStatistics.h
@@ -119,10 +119,15 @@ public:
 	// at index int(N/2) in the equivalent sorted dataset, where N is the number of points.
 	// For a dataset with an even number of points, the median is the mean of the values at
 	// indices int(N/2)-1 and int(N/2) in the sorted dataset.
+	// <src>nBins</src> is the number of bins, per histogram, to use to bin the data. More
+	// bins decrease the likelihood that multiple passes of the data set will be necessary, but
+	// also increase the amount of memory used. If nBins is set to less than 1,000, it is
+	// automatically increased to 1,000; there should be no reason to ever set nBins to be
+	// this small.
 	virtual AccumType getMedian(
 		CountedPtr<uInt64> knownNpts=NULL, CountedPtr<AccumType> knownMin=NULL,
 		CountedPtr<AccumType> knownMax=NULL, uInt binningThreshholdSizeBytes=4096*4096,
-		Bool persistSortedArray=False
+		Bool persistSortedArray=False, uInt64 nBins=10000
 	);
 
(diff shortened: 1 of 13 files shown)


scimath/Mathematics/ClassicalStatistics.h
```diff
@@ -38,6 +38,9 @@
 #include <vector>
 #include <utility>
 
+// because the template signature has become unwieldy
+#define _TPARMS AccumType, DataIterator, MaskIterator, WeightsIterator
```
juliantaylor: hm underscore followed by uppercase letter is reserved by the c++ standard for any use (17.6.4.3.2)
I a header it should at least be `#undef` at the end or have a `CASA? string in it

dmehring:
OK. My next set of commits will contain functionality changes. Then after that, I'll fix up the macro so you don't have to review both types of changes at once.

juliantaylor:
can c++11 template aliases be used here? (more out of curiosity as casacore is not c++11 only right now)
maybe one could even use them when the whole template class is a macro in c++98 and a template alias in c++11

JimNrao:
Another option that would reduce the template parameter list usage 
considerably would be to be to include the implementation inside the 
class construct:

template <typename A, typename B, typename C, typename D>
class Classical {

     Classical & operator= (const Classical & other) { ...}

};

vs

template <typename A, typename B, typename C, typename D>
ClassicalOperator<A,B,C,D> & ClassicalOperator<A,B,C,D>::operator=(const 
ClassicalOperator<A,B,C,D> & other) {...}


On 02/05/2016 10:44 AM, Julian Taylor wrote:
>
> can c++11 template aliases be used here? (more out of curiosity as 
> casacore is not c++11 only right now)
> maybe one could even use them when the whole template class is a macro 
> in c++98 and a template alias in c++11
>
> —
> <https://github.com/casacore/casacore/pull/302#issuecomment-180463358>.Web 
> Bug from 
> https://github.com/notifications/beacon/AGPGtIAgi_S3K8qBNK9rh8kk3U7nsGUDks5phNbugaJpZM4HUbsH.gif
>

-- 
Jim Jacobs

phone: 835-7235
office: 301

dmehring:
committed backward compatible api changes to allow caller to specify number of bins to use in qunatile computations and modified LatticeStatistics implementation to decrease the probability that multiple passes of large images will be necessary when computing quantile-related quantities. I'll work on cleaning up the aliases now. While I understand Jim's point, combining the implementation with the method definitions is not a good idea in this case IMO because this is not generally how we do things and because I think combining large chunks of code with API documentation would make things more difficult to maintain.

dmehring:
committed alias stuff. No functionality changes in this commit.

juliantaylor:
thanks

PR #303, opened 2016-02-09
Update value of 1 astronomical unit to conform to IAU standards

tammojan:
http://www.iau.org/static/resolutions/IAU2012_English.pdf recommends that 1 astronomical unit be defined as 149597870700 meter. Currently, quanta uses C::c * IAU_tauA, where IAU_tauA = 499.0047837. This is 149597870659.2 m, a relative difference of 10^-10.

### DIFF
--- a/casa/Quanta/UnitMap4.cc
+++ b/casa/Quanta/UnitMap4.cc
@@ -31,11 +31,6 @@
 
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
-//# constants
-
-// IAU definition of light time (s) unit distance to calculate IAU units
-static const Double IAU_tauA=499.0047837;	
-
 // Initialise the maps
 void UnitMap::initUMSI2() {
   // non-metric SI units
@@ -91,7 +86,7 @@ void UnitMap::initUMSI2() {
 						 "jansky")));
   UnitMap::mapSI->insert(map<String, UnitName>::value_type
 			 ("AU",   	UnitName("AU",
-						 UnitVal(C::c * IAU_tauA,"m"),
+						 UnitVal(149597870700,"m"),
 						 "astronomical unit")));
   UnitMap::mapSI->insert(map<String, UnitName>::value_type
 			 ("UA",   	UnitName("UA",


dpetry:
What triggered this change?

tammojan:
> What triggered this change?

I was writing a tutorial for measures and noticed this value does not conform to the ''official'' value. Wim provided good reasons not to merge this, I'll paste them below.

tammojan:
Wim Brouw wrote:

> Life is never that simple. Changing the one, you should also change some time related stff and k, the earlier defining value.
> 
> I read that now:
> The speed of light could then be expressed exactly as c0 = 299792458 m/s, a standard also adopted by the IERS numerical standards.[18] From this definition and the 2009 IAU standard, the time for light to traverse an AU is found to be τA =499.0047838061±0.00000001 s, more than 8 minutes. By simple multiplication then, the best IAU 2009 estimate was A =c0τA = 149597870700±3 m,[19] based on a comparison of JPL and IAA–RAS ephemerides.[20][21][22]"
> 
> 
> But also look at http://syrte.obspm.fr/IAU_resolutions/capitaine_jsr11.pdf,
> 
> where things are explained.
> 
> In addition, the ephemeris used at the moment (JPL DE whatever) have a closed system of constatnts, and if you use them you should use the constants that go with it. The JPL database has those constants, and they can be read.to get e.g. the AU value belonging to the ephemerides.
> 
> I suggest not to change the AU as such; but look at all IAU constants as a whole, and in cases where needed, use the values provided bij the JPL DE used in the calculations.

Wim Brouw wrote:

> The casa set of constants is based on the 1976 iau definitions.
> 
> See: https://en.wikipedia.org/wiki/IAU_(1976)_System_of_Astronomical_Constants
> 
> For later:
> ftp://tai.bipm.org/iers/conv2010/chapter1/tn36_c1.pdf
>

dpetry:
Do we agree then that this merge should not take place and be discarded?

tammojan:
Yes, I'll close it.

PR #306, opened 2016-02-11
CAS-5492: Add overwrite option to writeFITSFile()

dmehring:
Allow users the option of overwriting an existing file


dpetry:
I will merge this now. Could you maybe add a case to the relevant unit test which exercises this?

dmehring:
I've written a unit test and pushed it on the same branch but it does not show up in the diffs. Does the pull request have to be open in order for further changes to be made?

tammojan:
Yes it does. Probably easiest is to make a new pull request (once a PR has been merged, it cannot be reopened).

dmehring:
Open PR 308.

PR #307, opened 2016-02-11
CAS-3883: add log message when filling source table tha it make take …

dmehring:
…some time. Apparently a user had a data set for which the import of the source table took a long time, and he was confused why nothing seemed to be happening.


dmehring:
@juliantaylor  @dpetry Would one of you please review this minor change? Thanks.

PR #309, opened 2016-02-16
CAS-7927: Allow clients to choose between native and conversion layer…

dmehring:
… frames when calling toWorld()

Add a Bool parameter which defaults to True to toWorld(Vector<Double>&, const Vector<Double>&)

If True, conversion layer frame, if present, is used as before. If false, native frame is used for conversion. No effect if coordinate class does not support conversion layer frame (only CoordinateSystem, DirectionCoordinate, and SpectralCoordinate support conversion layer frames at the present time).

Unit tests added for the affected classes.


dpetry:
Thanks. I had been thinking of suggesting something like this for a long time.
Please don't forget when you propagate this to ia.coordmeasures() for CAS-7927 that the
spectral ref frame information in the output record needs to be set properly.

PR #312, opened 2016-02-22
style, mostly indentation, changes

dmehring:


juliantaylor:
less tabs is always an improvement, thanks

Issue #55, opened 2015-03-31
lib path for Debian/Ubuntu MultiArch


GoogleCodeExporter:
```
Debian and Ubuntu are moving to a Multiarchitecture approach:

http://wiki.debian.org/ReleaseGoals/MultiArch

For this, the libraries should be installed in /usr/lib/<arch>, with <arch> 
being something like i386-linux-gnu or x86_64-linux-gnu (keep in mind that 
Debian is also ported to other CPUs and other kernels).

While the current cmake-2.8.4 supports finding libraries in these directories, 
the casacore libraries are not installed there.

The attached patch fixes this: it allows to specify the architecture with

cmake -DLIBRARY_OUTPUT_PATH=lib/$(DEB_HOST_GNU_TYPE)

This is similar to <http://code.google.com/p/casacore/issues/detail?id=54>, so 
maybe the two bugs could be merged.
```

Original issue reported on code.google.com by `ole.stre...@gmail.com` on 14 Sep 2012 at 3:35

Attachments:
- [debian_patches_multiarch.patch](https://storage.googleapis.com/google-code-attachments/casacore/issue-55/comment-0/debian_patches_multiarch.patch)

tammojan:
@olebole, could you assess whether this is still necessary? (Original patch is from 2012.)

olebole:
Sorry, I didn't follow the development since then. @gijzelaerr , could you have a look?

gijzelaerr:
@tammojan I still think this is a proper improvement, since it would make use of more default CMake configuration variables. But probably we should use `LIBRARY_OUTPUT_DIRECTORY` now. Also we need to check how this works with pre 3.0 cmake.

http://www.cmake.org/cmake/help/v3.0/prop_tgt/LIBRARY_OUTPUT_DIRECTORY.html

gijzelaerr:
maybe implemented by https://github.com/casacore/casacore/pull/139

gijzelaerr:
ah never mind, it is set but not used.

Which makes me think, I'm not sure if you are supposed to set this manually this way (as done in the accepted PR)

gijzelaerr:
For Debian it looks like it is possible to set the lib directory:

https://salsa.debian.org/debian-astro-team/casacore/blob/master/debian/rules#L21

closing this issue.

PR #314, opened 2016-02-23
add optional LIBRARY_OUTPUT_PATH option to set lib install path

gijzelaerr:
Should solve https://github.com/casacore/casacore/issues/55 and make Ole happy.

this adds an optional LIBRARY_OUTPUT_PATH option which you can override to set installation location for libraries. Useful for the Debian multiarch implementation.

Should not interfere with current build scripts.


gijzelaerr:
Ok i'm not sure how this starts failing. I only set `LIBRARY_OUTPUT_PATH` and use it. It should have been set to `lib` on all platforms and it looks like it is set like that now also. Weird thing is, for this PR the tConvert test puts the python library in a `lib` folder (normally not), so it can't find find the so and the test fails.

If i run this `mv python/Converters/test/lib/* python/Converters/test` the tConvert test works. Why tmeas, no idea, can't replicate it on my system for now.

PR #317, opened 2016-02-23
make String::before() const, provide OO interface to MSFitsOutput

dmehring:
I provided an OO interface to MSFitsOutput. I moved the implementation of MSFitsOutput::writeFitsFile() to MSFitsOutput::write(), changing variable names to their private counterparts, and changed some style elements. 

For some reason SubString::before() and SubString::_substr() were not const, which required callers to make copies of const Strings. I fixed that.


dmehring:
@juliantaylor @dpetry Would one of you please review this PR? Thanks!

PR #318, opened 2016-02-24
add MSMetaData::getSpwIDs(), continuing refactor of MSFitsOutput

dmehring:
Ongoing refactoring of MSFitsOutput with help from MSMetaData. No MSFitsOutput functionality changes in this batch.

### DIFF
--- a/ms/MSOper/MSMetaData.cc
+++ b/ms/MSOper/MSMetaData.cc
@@ -1953,13 +1953,15 @@ vector<String> MSMetaData::getSpwNames() const {
 }
 
 std::set<uInt> MSMetaData::getSpwIDs() const {
-    Vector<Int> ddIDs = *_getDataDescIDs();
-    vector<uInt> ddIDToSpw = getDataDescIDToSpwMap();
+    const Vector<Int> ddIDs = *_getDataDescIDs();
+    const vector<uInt>& ddIDToSpw = getDataDescIDToSpwMap();
     Vector<Int>::const_iterator iter = ddIDs.begin();
     Vector<Int>::const_iterator end = ddIDs.end();
     std::set<uInt> spws;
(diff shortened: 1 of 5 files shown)


ms/MSOper/MSMetaData.cc
```diff
@@ -1952,6 +1952,18 @@ vector<String> MSMetaData::getSpwNames() const {
 	return out;
 }
 
+std::set<uInt> MSMetaData::getSpwIDs() const {
+    Vector<Int> ddIDs = *_getDataDescIDs();
+    vector<uInt> ddIDToSpw = getDataDescIDToSpwMap();
```
juliantaylor: this is unused -> function returns ddis?

juliantaylor: also looks like it could all be const references instead of copies

dmehring:
Good catch. That was a bug. I now use the appropriate mapping. Note that = should provide a reference to a casa::Vector() (and somewhat confusingly to me, explicitly using an & on that line to denote a reference causes the resulting Vector to be corrupted in one unit test, causing a segfault).

juliantaylor:
ah right I missed that it was a capital V vector which is already a reference.

juliantaylor:
not sure why it crashed as a reference though.

can the test be updated to catch that bug?

dmehring:
I'm compiling with C++11 enabled, so not sure if that makes a difference or not.

Which bug? The reference bug or the bug I fixed? I really can't test the latter without swapping in a new dataset; in the current dataset there is only one polarization so the spwID is equal to the ddID which is why the test passed previously with the buggy version.

juliantaylor:
hm ok, probably not worth adding more data to test such a small function, merging

PR #321, opened 2016-02-26
Style changes, indentation, brackets, loop counter postfix -> prefix

dmehring:
Implementation changes to follow.


dmehring:
@juliantaylor If I make a pull request and then push more changes to the same file on the same branch before the first pull request is merged, is the reviewer able to see both my changes individually, or are they only able to see the union of both changes? I ask because this first change is style only, and the next change involves implementation changes, so it would be easier for the reviewer  to review them individually. If the reviewer is unable to distinguish between the two, I will hold off on my next push until the first has been merged.

juliantaylor:
if you just add commits they can be reviewed separately, its a bit less convenient but not a problem
make the added commits as self contained as reasonable.

dmehring:
OK thanks. So just committing separately is OK? I don't have to push them each time and so can push a set of multiple commits in one go and that will be OK to review?

juliantaylor:
yes you can push a set of commits in one go, but try to be descriptive in the commit messages.

gervandiepen:
I don't agree with the style changes in existing source files.
The code in Casacore had this style for a very long time, and it makes no sense to change it. We agreed that existing style would not be changed.

dmehring:
Hi Ger
Do you object to all changes or just some? 

In the case of the indentation changes, the header file uses spaces for indents but the .cc file uses tabs.  It takes unnecessary time to reconfigure ones editor when switching between the two files. 

Prefix ++ is more efficient than postfix ++ for simple incrementing of counters.

gervandiepen:
I don't object to replacing tabs by spaces and using prefix instead of
postfix (it is indeed better in a loop). I also do that in a regular way.
But you also change the placement of braces, the way function signatures
are written, constructor initializition is done, etc.
We all have our favourite style and our opinion about what is most
readable. It is fine for new code. But IMHO existing code should not be
changed like that.

dmehring:
So if I revert the changes in bracket placement, method signatures, and constructor initialization, then you'd be willing to accept the code? Anything else? 
Ultimately I'm adding code so this class will support creation of an ellipsoid with center outside the lattice, which is a required use case for our users. 
Thanks.

dmehring:
@gervandiepen I've reverted my style changes to which you object. I hope this now meets with your approval. I've implemented support for the case when the ellipsoid center is outside the lattice. I've added tests. I've added a test fixture to which test results can be compared.
Thanks.

gervandiepen:
Looks nice.
Ger

dmehring:
Thanks!

PR #328, opened 2016-03-11
Cas 8405: logging support for Int64 and uInt64 various MSMetaData/MSSummary mods

dmehring:
Added support for logging uInt64 and Int64.
Fixed bug in MSMetaData::_getSubScanProps() which caused incorrect size of data structure to be computed.

_getSubScanProperties() -> getSubScanProps() (private to public) to allow caller to hold the entire property structure so it doesn't have to call for each subscan. => performance improvement

added progress meter when running getSubScanProps()

added MSMetaData::getNRowMap() to allow caller to hold entire data structure rather than having to call another method for each separate subscan => performance improvement

Added method to allow caller to specify that MSMetaData should always cache subscan properties structure since it can be very expensive to recompute if max cache size has already been exceeded.

Added MSMetaData::getSubScanToIntentsMap() to allow caller to retrieve entire data structure rather than having to call another method for each subscan => performance improvement.

Changed several row determination methods from uInt to uInt64 because VLA is not writing MSes with > 100 mega rows, so it is just a matter of time before table interface limitations are going to result in overflows.

Made use of various performance improvements in MSSummary.

### DIFF
--- a/ms/MSOper/MSMetaData.cc
+++ b/ms/MSOper/MSMetaData.cc
@@ -29,6 +29,7 @@
 
 #include <casacore/casa/Arrays/MaskArrMath.h>
 #include <casacore/casa/OS/File.h>
+#include <casacore/casa/System/ProgressMeter.h>
 #include <casacore/measures/Measures/MeasTable.h>
 #include <casacore/measures/TableMeasures/ArrayQuantColumn.h>
 #include <casacore/ms/MSOper/MSKeys.h>
@@ -90,7 +91,7 @@ MSMetaData::MSMetaData(const MeasurementSet *const &ms, const Float maxCacheSize
 	  _taqlTempTable(
 		File(ms->tableName()).exists() ? 0 : 1, ms
 	  ), _flagsColumn(),
-	   _spwInfoStored(False) {}
+	   _spwInfoStored(False), _forceSubScanPropsToCache(False) {}
 
 MSMetaData::~MSMetaData() {}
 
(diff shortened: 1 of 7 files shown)


juliantaylor:
changing to int64 in the public interface breaks the binary compatibility. While I don't believe this is a valuable goal for casacore, it can still be avoided here as just changing this here will not help. All of casacore (and casa) would need a major overheal to support more than 2**32 rows in tables.

ms/MSOper/MSMetaData.cc
```diff
@@ -252,7 +254,34 @@ uInt MSMetaData::nRows(CorrelationType cType) {
...
+    );
+    if (cType == AUTO) {
+        return subScanToNACRowsMap;
+    }
+    else if (cType == CROSS) {
+        return subScanToNXCRowsMap;
+    }
+    SHARED_PTR<map<SubScanKey, uInt64> > mymap(
+        new map<SubScanKey, uInt64>()
+    );
```
juliantaylor: I guess we don't have a `make_shared` in our abstraction? would save a (probably irrelevant) memory allocation.

ms/MSOper/MSMetaData.cc
```diff
@@ -3550,28 +3625,30 @@ std::map<SubScanKey, MSMetaData::SubScanProperties> MSMetaData::_getSubScanPrope
...
-    uInt dSize = sizeof(Double);
-	uInt structSize = 3*dSize + iSize;
-    uInt keySize = 4*iSize;
-    std::map<SubScanKey, SubScanProperties>::const_iterator mIter = mysubscans.begin();
-    std::map<SubScanKey, SubScanProperties>::const_iterator mEnd = mysubscans.end();
-    uInt mapSize = mysubscans.size() * (structSize + keySize);
-    while (mIter != mEnd) {
+    static const uInt iSize = sizeof(Int);
+    static const uInt dSize = sizeof(Double);
+	static const uInt structSize = 3*dSize + iSize;
```
juliantaylor: wrong indentation

ms/MSOper/MSMetaData.cc
```diff
@@ -3550,28 +3625,30 @@ std::map<SubScanKey, MSMetaData::SubScanProperties> MSMetaData::_getSubScanPrope
...
         );
-        mapSize += (dSize + iSize) + mIter->second.timeProps.size();
+        mapSize += (dSize + iSize) + miter->second.timeProps.size();
         mapSize += (iSize + dSize) * miter->second.meanInterval.size();
-        ++mIter;
     }
     if (_cacheUpdated(mapSize)) {
-    	_subScanProperties = mysubscans;
+        _subScanProperties = mysubscans;
 	}
```
juliantaylor: wrong indentation I think

ms/MSOper/MSMetaData.cc
dmehring: Not that I'm aware of.

dmehring:
fixed indentation.

dmehring:
@juliantaylor  I've reverted the uInt64 changes back to uInt and corrected yet another indentation issue. Hopefully this addresses your remaining concerns.

juliantaylor:
thanks

PR #330, opened 2016-03-14
Improve testing of MSSummary

### LINKED ISSUE #329: Improve testing of MSSummary
MSSummary used to fail in previous casacore versions:
- For a short while it used the column keyword TabRefCodes, which does not need to be present.
- TableMeasures could not handle new TabRefTypes; this arose when MFrequency type Undefined was added.

It has been fixed, but it was not properly tested.

gervandiepen:
tMSSummary has been extended for better checking of TableMeasures used in the MS.

Close #329


tammojan:
Travis failed again (at the added test)

tammojan:
Squashed two commits, thanks for adding the test Ger.

PR #334, opened 2016-03-16
improve efficiency of tilestmanager

juliantaylor:
This changeset aims to improve the cpu efficiency of non-strided part of the tiled storage manager.
I noticed that a CASA task doing nothing but IO is significantly bottlenecked by the cpu and not able to fully utilize a mid-range SSD (700mb/s bandwidth).
The main culprit is the TSMCube::accessSection which reads non-stride tiles. In ALMA and EVLA data the tiles are typically (X, Y, Z) with X being only of size 2 or 4 elements, the main copy-loop in this function loops over each dimension of the tiles so it iterates on a tiny dimension. As the buffers read and written are usually contiguous we can often merge the loops into larger copy calls.

There are four commits, the first just reindents some code, the second refactors the copy loop into a function without using local variables. The third replaces a memset loop with a memset call which is a lot faster, especially with old compilers that can't do this replacement for you.
The loop merging is in the 4th commit.

In total this improves the throughput of our supposedly IO-only task on this hardware by about 25%. But the mileage may vary depending on tileshape, shape of the involved buffers and libc version.

### DIFF
--- a/tables/DataMan/TSMCube.cc
+++ b/tables/DataMan/TSMCube.cc
@@ -48,55 +48,30 @@
 
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
-// Find out if local size is a multiple of 4, so we can move as integers.
-#define TSMCube_FindMult \
-  uInt localPixelWords = 0; \
-  if (localPixelSize % sizeof(Int) == 0) { \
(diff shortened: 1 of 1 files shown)


gervandiepen:
Thanks Julian. Looks like a nice improvement.

PR #336, opened 2016-03-18
Align gridding operation in ggridsd2 with the corresponding part of ggridsd

tnakazato:
This change intends to make result of ggridsd2 identical with the one from ggridsd. To do that, gridding  operation in ggridsd2 are aligned with the corresponding part of ggridsd.

### DIFF
--- a/scimath_f/fgridsd.f
+++ b/scimath_f/fgridsd.f
@@ -293,6 +293,8 @@ C
       integer chanmap(nvischan), polmap(nvispol)
       integer dowt
 
+      complex nvalue
+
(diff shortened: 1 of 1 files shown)


tammojan:
Who can review this?

alxames:
Malte or Ger. (?)

git blame suggests the code was originally written by Tim Cornewell, moved and partly rewritten by Malte Marquarding 10 years later, then further patched by Ger van Diepen.

gervandiepen:
Seems like a good change to me.

PR #338, opened 2016-03-18
Add std to isinf and isnan, fixes #337

### LINKED ISSUE #337: Seems missing "std::" for some isinf, isnan function call in some .cc files
I generate the makefile with the command:
mkdir -p build; cd build; cmake -DCXX11=ON -DUSE_FFTW3=ON -DDATA_DIR=/usr/local/share/casacore/data -DUSE_OPENMP=ON -DUSE_HDF5=ON -DBUILD_PYTHON=ON -DUSE_THREADS=ON -DPYTHON2_Boost_LIBRARIES=/usr/lib/libboost_python.so ..

and the run the make command.

Then some error messages appear when compiling BasicMath/Math.cc, and BasicMath/test/tMathNaN.cc, complaining that it cannot find the declaration of isnan, isinf functions.

Then I modified above two files and add a std:: before the isinf, (and some others std:: in other places, see the attached contents), then all the errors disappear.

My system is archlinux with the latest update. My gcc version:
gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-unknown-linux-gnu/5.3.0/lto-wrapper
Target: x86_64-unknown-linux-gnu
Configured with: /build/gcc/src/gcc-5-20160209/configure --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=https://bugs.archlinux.org/ --enable-languages=c,c++,ada,fortran,go,lto,objc,obj-c++ --enable-shared --enable-threads=posix --enable-libmpx --with-system-zlib --with-isl --enable-__cxa_atexit --disable-libunwind-exceptions --enable-clocale=gnu --disable-libstdcxx-pch --disable-libssp --enable-gnu-unique-object --enable-linker-build-id --enable-lto --enable-plugin --enable-install-libiberty --with-linker-hash-style=gnu --enable-gnu-indirect-function --disable-multilib --disable-werror --enable-checking=release
Thread model: posix
gcc version 5.3.0 (GCC) 

uname -a
Linux 4.4.5-1-ARCH #1 SMP PREEMPT Thu Mar 10 07:38:19 CET 2016 x86_64 GNU/Linux

The result of git diff is shown as follows, which is also attached as the file diff.txt

`
diff --git a/casa/BasicMath/Math.cc b/casa/BasicMath/Math.cc
index d1bd984..00659e3 100644
--- a/casa/BasicMath/Math.cc
+++ b/casa/BasicMath/Math.cc
@@ -178,7 +178,7 @@ Bool isInf(Float val) {
   // infinite. I can only have access to Solaris, Linux and SGI machines to
   // determine this.
 #if defined(AIPS_LINUX)
-  return (isinf(Double(val)));
+  return (std::isinf(Double(val)));
 #elif defined(AIPS_DARWIN)
   return (std::isinf(Double(val)));
 #elif defined(AIPS_SOLARIS) || defined(AIPS_IRIX)
@@ -212,7 +212,7 @@ Bool isInf(Double val) {
   // infinite. I can only have access to Solaris, Linux and SGI machines to
   // determine this.
 #if defined(AIPS_LINUX)
-  return (isinf(Double(val)));
+  return (std::isinf(Double(val)));
 #elif defined(AIPS_DARWIN)
   return (std::isinf(Double(val)));
 #elif defined(AIPS_SOLARIS) || defined(AIPS_IRIX)
diff --git a/casa/BasicMath/test/tMathNaN.cc b/casa/BasicMath/test/tMathNaN.cc
index f1a73bc..0c330d1 100644
--- a/casa/BasicMath/test/tMathNaN.cc
+++ b/casa/BasicMath/test/tMathNaN.cc
@@ -43,7 +43,7 @@
 			    ((*(Int *)(x) & 0x007fffff) != 0x00000000))
 
 inline Bool isNaN_isnan(Float val) {
-  return (isnan(Double(val)));
+  return (std::isnan(Double(val)));
 }
 
 inline Bool isNaN_isnanf(const Float& val) {

`

[diff.txt](https://github.com/casacore/casacore/files/179024/diff.txt)

tammojan:
As per #337. Not sure this is portable, there may be problems with old intel compilers ( https://software.intel.com/en-us/forums/intel-c-compiler/topic/297833 )

### DIFF
--- a/casa/BasicMath/Math.cc
+++ b/casa/BasicMath/Math.cc
@@ -178,7 +178,7 @@ Bool isInf(Float val) {
   // infinite. I can only have access to Solaris, Linux and SGI machines to
   // determine this.
 #if defined(AIPS_LINUX)
-  return (isinf(Double(val)));
+  return (std::isinf(Double(val)));
 #elif defined(AIPS_DARWIN)
   return (std::isinf(Double(val)));
 #elif defined(AIPS_SOLARIS) || defined(AIPS_IRIX)
@@ -212,7 +212,7 @@ Bool isInf(Double val) {
   // infinite. I can only have access to Solaris, Linux and SGI machines to
   // determine this.
 #if defined(AIPS_LINUX)
-  return (isinf(Double(val)));
+  return (std::isinf(Double(val)));
 #elif defined(AIPS_DARWIN)
   return (std::isinf(Double(val)));
 #elif defined(AIPS_SOLARIS) || defined(AIPS_IRIX)
--- a/casa/BasicMath/test/tMathNaN.cc
+++ b/casa/BasicMath/test/tMathNaN.cc
@@ -43,7 +43,7 @@
 			    ((*(Int *)(x) & 0x007fffff) != 0x00000000))
 
 inline Bool isNaN_isnan(Float val) {
-  return (isnan(Double(val)));
+  return (std::isnan(Double(val)));
 }
 
 inline Bool isNaN_isnanf(const Float& val) {


tammojan:
Please wait with merging until after the 2.1.0 release.

gervandiepen:
If you are not sure about old compilers, it might be better to use
   using namespace std;
instead of std::

juliantaylor:
I do not believe it is worthwhile to worry too much about hypothetical less used compilers failing prematurely, we revisit it should an issue turn up. this is needed now for a common compiler, so merging.

PR #339, opened 2016-03-18
bug fix, exposure times were incorrectly being used as intervals

dmehring:


juliantaylor:
urg that should have been spotted by review, sloppy me ...

dmehring:
Thanks. Given that I coded the bug, I cannot pass judgement on any one else for not spotting it :) .

juliantaylor:
@tammojan just made a release for other projects a few hours ago.
but this doesn't look like its necessarily a big problem as it deals with a new feature for casa specific data right?

dmehring:
I suspect it does not pose a major issue for other groups, but not knowing the extent of which they use this and related classes nor the details of their data sets (in ours exposure times and intervals are often the same or very close to being equal, so for us this isn't a big problem), I cannot say for certain to the degree others might be affected.

PR #350, opened 2016-03-23
Cas 8447, two commits

dmehring:
The first commit on this branch represents the functionality change. This allows one to subclass MSMetaData and override the base class implementation for computing the scan and subscan data structures. eg, subclasses might implement multithreading to compute these structures.

The second commit on this branch changes tabs to spaces in these files, there are no functionality changes in this commit.


juliantaylor:
this should not have been merged, the implications of this are still under discussion, sorry I did not note this here

dpetry:
That is strange. The pull request came from Dave who probably tested this with CASA.
Is this why the trunk doesn't build now?

juliantaylor:
no, the issue is more we haven't decided yet if this is the best approach to allow parallelization of the msmetadata.
this approach does require a not insignificant amount of code in casa, maybe it could also be done within the original msmetadata in casacore.

PR #362, opened 2016-03-29
Added stand-alone program to convert text file to a table

### LINKED ISSUE #345: Create a program to convert text file to table
A stand-alone program to convert a text file to a table is useful.
Also fixed ReadAsciiTable to correctly handle a cr-lf end-of-line sequence as used in DOS-files.

gervandiepen:
It is handy to have a program converting a text file to a table. It should support cr-lf as end-of-line
Close #345


gervandiepen:
tablefromascii also supports array columns (see comments in ReadAsciiTable.h).

PR #373, opened 2016-03-30
Change specification of modules to build

### LINKED ISSUE #365: Improve definition of modules to build
Lattices should be built if images are built.
It should be possible to build ms/derivedmscal without fits.

gervandiepen:
It was not possible to build module derivedmscal without fits; this has been improved.
Close #365


tammojan:
I think that this change makes it impossible to build images without ms. Not sure that's a problem though.

gervandiepen:
You're right. I'll make a better one.

PR #376, opened 2016-03-30
set defaults for -DCASA_BUILD=1 (i.e. building within CASA context)

schiebel:
defaults (protected by 'CASA_BUILD=1') for developer builds in the context of a wider CASA build


juliantaylor:
there is no need to make a new pull request, just commit --amend and then push -f <fork>

juliantaylor:
seems to work, thanks

PR #380, opened 2016-03-31
Array functors derive from a base class

### LINKED ISSUE #371: Let Array Functors derive from a base class to reduce code bloat
The partialMath functions make use of Functors to apply the required operation to an array section. Currently, it is fully templated, so the functor is a template parameter resulting in quite some code bloat.
It would be better to let the Functors derive from a base class, so only a single template instantiation per data type. The functor's apply function will be virtual, but that overhead is hardly a problem because the functor operates on an array, not on a single value.

gervandiepen:
Array functors derive from a base class to reduce code bloat in arrayPartMath.
Furthermore, a few functors have been added.
Close #371

### DIFF
--- a/casa/Arrays/ArrayLogical.h
+++ b/casa/Arrays/ArrayLogical.h
@@ -32,6 +32,7 @@
 #include <casacore/casa/aips.h>
 #include <casacore/casa/Arrays/Array.h>
 #include <casacore/casa/Arrays/LogiArray.h>
+#include <casacore/casa/Arrays/ArrayMathBase.h>
 
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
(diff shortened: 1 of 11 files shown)


casa/Arrays/ArrayLogical.h
```diff
@@ -390,13 +391,18 @@ template<class T> Bool anyOR (const T &val, const Array<T> &array);
...
 inline Bool anyTrue (const Array<Bool>& array)
   { return anyEQ (array, True); }
 
-// 
+// The same functions as above, but for selected axes.
+Array<Bool> partialAllTrue (const Array<Bool>& array,
+                            const IPosition& collapseAxes);
+Array<Bool> partialAnyTrue (const Array<Bool>& array,
+                            const IPosition& collapseAxes);
+
```
juliantaylor: wouldn't this make more sense as an overloaded variant of anyTrue?
partial does not sound axis reduction to me, more like slicing/windowing

gervandiepen:
As such it it is a good idea to use the names allTrue and anyTrue. However, there are many more such functions in ArrayLogical.h and ArrayPartMath.h which are called partialXXX. So I prefer to keep the name. Renaming them all is not backward compatible.

juliantaylor:
ah I wasn't aware of that, if its existing convention its of course better to keep it

PR #381, opened 2016-03-31
Added StorageOption argument to Table::copy

### LINKED ISSUE #374: Let Table::copy have an optional StorageOption argument
The copy and deepCopy function in class Table should have a StorageOption argument to make changing the storage format possible.

gervandiepen:
Copying an table can use the StorageOption argument.
showKeywords has been moved to class Table to be commonly available.
Close #374

### DIFF
--- a/tables/Tables/BaseTable.cc
+++ b/tables/Tables/BaseTable.cc
@@ -402,14 +402,15 @@ void BaseTable::renameSubTables (const String&, const String&)
 {}
 
 void BaseTable::deepCopy (const String& newName,
-			  const Record& dataManagerInfo,
+                          const Record& dataManagerInfo,
+                          const StorageOption& stopt,
 			  int tableOption,
 			  Bool valueCopy,
 			  int endianFormat,
 			  Bool noRows) const
 {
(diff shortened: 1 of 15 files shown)


juliantaylor:
note that this change breaks binary compatiblity

gervandiepen:
Changes in the BaseTable, etc. API/ABI are internal, thus do not break compatibility.
I do agree that addition of a parameter to the TableCopy copy function breaks the ABI. However, a few more non ABI-compatible changes have been checked in these days.
It is the intention that these changes (with the upcoming TaQL changes) become a next major release.

juliantaylor:
its an installed header and a public function.
but its just an observation I think abi compatibility is not a worthwhile goal for something like casacore, but so far I know casacore is binary packaged in some places so it needs to at least be noted.

gervandiepen:
Maybe we should add an issue to go through the CMakeLists files and only install the necessary header files.

juliantaylor:
I guess the basetable header is needed at least a casa components includes it (but probably unnecessarily)
one could make it non-inheritable and make the public functions protected

PR #383, opened 2016-03-31
CAS-8447 enable parallel processing using omp

dmehring:
I had made some methods protected in a previous commit that was merged, but now these can be re-made private since I was able to provide a multi-threaded implementation using openmp in MSMetaData, obviating the need for a derived class.

### DIFF
--- a/ms/MSOper/MSMetaData.cc
+++ b/ms/MSOper/MSMetaData.cc
@@ -3382,178 +3382,6 @@ void MSMetaData::_computeScanAndSubScanProperties(
     SHARED_PTR<std::map<SubScanKey, MSMetaData::SubScanProperties> >& subScanProps,
     Bool showProgress
 ) const {
-    SHARED_PTR<Vector<Int> > scans = _getScans();
-    SHARED_PTR<Vector<Int> > fields = _getFieldIDs();
-    SHARED_PTR<Vector<Int> > ddIDs = _getDataDescIDs();
-    SHARED_PTR<Vector<Int> > states = _getStateIDs();
-    SHARED_PTR<Vector<Double> > times = _getTimes();
-    SHARED_PTR<Vector<Int> > arrays = _getArrayIDs();
-    SHARED_PTR<Vector<Int> > observations = _getObservationIDs();
-    SHARED_PTR<Vector<Int> > ant1, ant2;
-    SHARED_PTR<QVD> exposureTimes = _getExposureTimes();
-    SHARED_PTR<QVD> intervalTimes = _getIntervals();
-    vector<uInt> ddIDToSpw = getDataDescIDToSpwMap();
-    _getAntennas(ant1, ant2);
-    Vector<Int>::const_iterator scanIter = scans->begin();
-    Vector<Int>::const_iterator scanEnd = scans->end();
-    Vector<Int>::const_iterator a1Iter = ant1->begin();
-    Vector<Int>::const_iterator a2Iter = ant2->begin();
-    Vector<Int>::const_iterator fIter = fields->begin();
-    Vector<Int>::const_iterator dIter = ddIDs->begin();
-    Vector<Int>::const_iterator stateIter = states->begin();
-    Vector<Int>::const_iterator oIter = observations->begin();
-    Vector<Int>::const_iterator arIter = arrays->begin();
-    Vector<Double>::const_iterator tIter = times->begin();
-    const Vector<Double>& exposures = exposureTimes->getValue();
-    Vector<Double>::const_iterator eiter = exposures.begin();
-    const Vector<Double>& intervals = intervalTimes->getValue();
-    Vector<Double>::const_iterator iIter = intervals.begin();
-    SubScanKey subScanKey;
-    ScanKey scanKey;
-    uInt count = 0;
-    subScanProps.reset(
-        new std::map<SubScanKey, SubScanProperties>()
-    );
-    scanProps.reset(
-        new std::map<ScanKey, ScanProperties>()
-    );
-    map<std::pair<ScanKey, uInt>, uInt> counts;
-    map<pair<SubScanKey, uInt>, Double> intervalSum;
-    map<SubScanKey, Double> exposureSum;
-    pair<SubScanKey, uInt> subScanSpw;
-    SHARED_PTR<ProgressMeter> pm;
-    if (showProgress) {
-        LogIO log;
-        log << LogOrigin("MSMetaData", __func__, WHERE)
-            << LogIO::NORMAL << "Compute subscan properties"
-            << LogIO::POST;
-        pm.reset(new ProgressMeter(0, scans->size(), "Compute subscan info"));
-    }
-    while (scanIter != scanEnd) {
-        scanKey.obsID = *oIter;
-        scanKey.arrayID = *arIter;
-        scanKey.scan = *scanIter;
-        Double half = *iIter/2;
-        uInt spw = ddIDToSpw[*dIter];
-        if (scanProps->find(scanKey) == scanProps->end()) {
-            (*scanProps)[scanKey].times[spw] = std::set<Double>();
-            (*scanProps)[scanKey].timeRange = std::pair<Double, Double>(*tIter-half, *tIter+half);
-        }
-        else {
-            (*scanProps)[scanKey].timeRange.first = min((*scanProps)[scanKey].timeRange.first, *tIter-half);
-            (*scanProps)[scanKey].timeRange.second = max((*scanProps)[scanKey].timeRange.second, *tIter+half);
-        }
-        (*scanProps)[scanKey].times[spw].insert(*tIter);
-        subScanKey.obsID = *oIter;
-        subScanKey.arrayID = *arIter;
-        subScanKey.scan = *scanIter;
-        subScanKey.fieldID = *fIter;
-        subScanSpw.first = subScanKey;
-        subScanSpw.second = spw;
-        if (subScanProps->find(subScanKey) == subScanProps->end()) {
-            SubScanProperties props;
-            props.beginTime = *tIter;
-            props.endTime = *tIter;
-            props.nrows = 1;
-            exposureSum[subScanKey] = *eiter;
-            intervalSum[subScanSpw] = *iIter;
-            props.spwNRows[spw] = 1;
-            (*subScanProps)[subScanKey] = props;
-        }
-        else {
-            (*subScanProps)[subScanKey].beginTime = min(*tIter, (*subScanProps)[subScanKey].beginTime);
-            (*subScanProps)[subScanKey].endTime = max(*tIter, (*subScanProps)[subScanKey].endTime);
-            ++(*subScanProps)[subScanKey].nrows;
-            exposureSum[subScanKey] += *eiter;
-            if (intervalSum.find(subScanSpw) == intervalSum.end()) {
-                (*subScanProps)[subScanKey].spwNRows[spw] = 1;
-                intervalSum[subScanSpw] = *iIter;
-            }
-            else {
-                ++(*subScanProps)[subScanKey].spwNRows[spw];
-                intervalSum[subScanSpw] += *iIter;
-            }
-        }
-        (*subScanProps)[subScanKey].antennas.insert(*a1Iter);
-        (*subScanProps)[subScanKey].antennas.insert(*a2Iter);
-        (*subScanProps)[subScanKey].ddIDs.insert(*dIter);
-        (*subScanProps)[subScanKey].spws.insert(spw);
-        (*subScanProps)[subScanKey].stateIDs.insert(*stateIter);
-        std::map<Double, TimeStampProperties>& timeProps = (*subScanProps)[subScanKey].timeProps;
-        if (timeProps.find(*tIter) == timeProps.end()) {
-            timeProps[*tIter].nrows = 1;
-        }
-        else {
-            ++timeProps[*tIter].nrows;
-        }
-        timeProps[*tIter].ddIDs.insert(*dIter);
-        ++tIter;
-        ++scanIter;
-        ++fIter;
-        ++dIter;
-        ++stateIter;
-        ++oIter;
-        ++arIter;
-        ++a1Iter;
-        ++a2Iter;
-        ++eiter;
-        ++iIter;
-        if (pm) {
-            ++count;
-            if (count % 1000000 == 0) {
-                 pm->update(count);
-            }
-        }
-    }
-    map<pair<SubScanKey, uInt>, Double>::const_iterator scIter = intervalSum.begin();
-    map<pair<SubScanKey, uInt>, Double>::const_iterator scEnd = intervalSum.end();
-    Unit unit = intervalTimes->getFullUnit();
-    map<pair<ScanKey, uInt>, Double> scanIntervalSum;
-    pair<ScanKey, uInt> scanSpw;
-    for (; scIter != scEnd; ++scIter) {
-        const SubScanKey& subScan = scIter->first.first;
-        const uInt spw = scIter->first.second;
-        (*subScanProps)[subScan].meanInterval[spw] = Quantity(
-            scIter->second/((*subScanProps)[subScan].spwNRows[spw]), unit
-        );
-        scanSpw.first = casa::scanKey(subScan);
-        scanSpw.second = spw;
-        if ((*scanProps)[scanKey].spwNRows.find(spw) == (*scanProps)[scanKey].spwNRows.end()) {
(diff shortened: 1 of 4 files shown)


ms/MSOper/MSMetaData.cc
```diff
@@ -3546,6 +3553,345 @@ void MSMetaData::_computeScanAndSubScanProperties(
...
+    _pmCount = 0;
+    SHARED_PTR<ProgressMeter> pm;
+    if (showProgress) {
+        LogIO log;
+        log << LogOrigin("MSMetaDataThreaded", __func__, WHERE)
+            << LogIO::NORMAL << "Compute subscan properties " << LogIO::POST;
+        pm.reset(new ProgressMeter(0, nrows, "Compute subscan info"));
+    }
+    pair<map<ScanKey, ScanProperties>, map<SubScanKey, SubScanProperties> > *fut =
+        new pair<map<ScanKey, ScanProperties>, map<SubScanKey, SubScanProperties> >[nthreads];
```
juliantaylor: variable sized arrays are a gnu extensions, better use a heap array

ms/MSOper/MSMetaData.cc
```diff
@@ -3546,6 +3553,345 @@ void MSMetaData::_computeScanAndSubScanProperties(
...
+#pragma omp parallel for
+    for (uInt i=0; i<nthreads; ++i) {
```
juliantaylor: this doesn't look right, first nthreads ~= nrows so the name is wrong, then why does this loop on the rows but the begin and row are larger chunks, so most rows are processed multiple times

ms/MSOper/MSMetaData.cc
```diff
@@ -3546,6 +3553,345 @@ void MSMetaData::_computeScanAndSubScanProperties(
...
+        fut[i] = _getChunkSubScanProperties(
+            scans, fields, ddIDs, states, times, arrays,
+            observations, ant1, ant2, exposureTimes,
+            intervalTimes,  ddIDToSpw,  i * chunkSize,
+            std::min((i + 1) * chunkSize - 1, nrows - 1)
+        );
+#pragma omp critical(progresslock)
+        {
+            if (pm) {
+#pragma omp flush(_pmCount)
```
juliantaylor: why is this a shared variable?
if the iteration is in chunks this should just be the iteration counter

juliantaylor: possibly the api of progressmeter is not to well suited for undefined parallization ordering, so one should either make it ordered or update the api to allow increasing not based on certain counts

juliantaylor:
the code adds a lot of functions guarded with ifdefs but I see no indication in the code for why this is needed.
openmp in most cases can be turned on and off without semantical change. this looks like such a case.
why are special functions needed when openmp is enabled?

ms/MSOper/MSMetaData.cc
dmehring: I thought that's what new() did. Can you please provide details of how to do what you suggest?

juliantaylor: ups I'm sorry I misread the code, this is fine

ms/MSOper/MSMetaData.cc
juliantaylor: actually keeping it shared is probably simplest, but it should be updated here inside the loop based on the processed chunksize instead of for each single row in another function.

ms/MSOper/MSMetaData.cc
dmehring: I'm not following your objection. The loop is over the threads, not over rows. nthreads can be equal to nrows if nrows is small (<= 1000), but that's only for small data sets, in most cases nthreads, which maxes out at 1,000, will be significantly less than nrows. 

I don't see how the same row gets processed more than once. Consider the case where nrows = 1e7, for example. In that case nthreads = 1000, so chunkSize = 10,000. One then has the following relations:

i       begnRow(= i*chunkSize)      endRow(=std::min((i + 1) * chunkSize - 1, nrows - 1))
0             0                                    9999
1      10000                                  19999
2      20000                                  29999
...
999   9,990,000                         9,999,999
So each row is processed exactly once.

juliantaylor: ah you are taking the minimum and setting the chunksize appropriately
still the variable should be named nchunks not nthreads

dmehring: OK, I will rename the variable.

ms/MSOper/MSMetaData.cc
dmehring: OK, will change so progress meter is updated using chunks processed, rather than updating the counter for each row processed.

dmehring:
Re: ifdef'ed methods: I wasn't clear on the convention. I'll remove the ifdefs.

dmehring:
pushed. Hopefully I've reasonably addressed all the issues you raised. If not, let me know.
Thanks!

ms/MSOper/MSMetaData.cc
```diff
@@ -3571,9 +3913,22 @@ void MSMetaData::_getScanAndSubScanProperties(
...
+    else {
+        _computeScanAndSubScanProperties(
+            myscanprops, myssprops, showProgress
+        );
+    }
+#else
     _computeScanAndSubScanProperties(
         myscanprops, myssprops, showProgress
     );
+#endif
```
juliantaylor: is this branch necessary? is there a reason to not always take the parallel code path and simply limit the openmp threads to 1 if `_parallel` is false

this is done via

```
#pragma omp parallel for if (_parallel)
...
```

ms/MSOper/MSMetaData.cc
```diff
@@ -3546,6 +3553,341 @@ void MSMetaData::_computeScanAndSubScanProperties(
...
+    uInt doneChunks = 0;
+#pragma omp parallel for
+    for (uInt i=0; i<nchunks; ++i) {
+        fut[i] = _getChunkSubScanProperties(
+            scans, fields, ddIDs, states, times, arrays,
+            observations, ant1, ant2, exposureTimes,
+            intervalTimes,  ddIDToSpw,  i * chunkSize,
+            std::min((i + 1) * chunkSize - 1, nrows - 1)
+        );
+#pragma omp atomic
```
juliantaylor: now this can be moved into in the critical section, no need for the atomic anymore.

ms/MSOper/MSMetaData.cc
```diff
@@ -40,6 +41,12 @@
 #include <casacore/tables/TaQL/TableParse.h>
 #include <casacore/tables/Tables/TableProxy.h>
 
+#include <utility>
+
+#ifdef _OPENMP
+#include <omp.h>
+#endif
```
juliantaylor: this is not needed when you are not using the openmp C-api (function calls starting with ´omp_`)

ms/MSOper/MSMetaData.cc
```diff
@@ -3546,6 +3553,341 @@ void MSMetaData::_computeScanAndSubScanProperties(
...
+    }
+    pair<map<ScanKey, ScanProperties>, map<SubScanKey, SubScanProperties> > *fut =
+        new pair<map<ScanKey, ScanProperties>, map<SubScanKey, SubScanProperties> >[nchunks];
+    uInt doneChunks = 0;
+#pragma omp parallel for
+    for (uInt i=0; i<nchunks; ++i) {
+        fut[i] = _getChunkSubScanProperties(
+            scans, fields, ddIDs, states, times, arrays,
+            observations, ant1, ant2, exposureTimes,
+            intervalTimes,  ddIDToSpw,  i * chunkSize,
```
juliantaylor: does this need to be `i*chunksize + 1` as the function below has a while smaller *equal* loop?

ms/MSOper/MSMetaData.cc
dmehring: The progress meter will be useless in the non-parallel case if nchunks = 1. It behaves properly in the serial-specific method.

juliantaylor:
do you have numbers on how effective this parallelization is?

ms/MSOper/MSMetaData.cc
juliantaylor: `nchunks` 1 is a one row table, is the progressmeter relevant there?

dmehring:
I can get them, The speed up is qualitatively quite significant.

ms/MSOper/MSMetaData.cc
juliantaylor: and if it is, is the progressmeter worth 200 lines of extra code which is almost a duplicate?
if yes there are certainly ways to restructure it to make the progress meter more useful for tables with few rows.

dmehring: Ah, I was thinking about it incorrectly, so I think you are right there. But there is still a lot of collating of data that is done in the loops following that loop, which isn't necessary to do in the serial-specific version. This may not be all that significant, I'm not sure, but since the serial-specific version already exists, it seems this unnecessary overhead can be avoided simply by using it when necessary.

ms/MSOper/MSMetaData.cc
dmehring: i*chunkSize is the beginning row, so I think i*chunkSize is correct. If you add 1, then when i = 0 the beginning row is 1, so you do not process row 0.

ms/MSOper/MSMetaData.cc
juliantaylor: I have a hard time understanding how the collating really works (some more in code comments would help the less experienced understanding it faster) but naively I would think that the overhead is either relatively low or a way could be found to skip the collation and work directly on the result if it is expensive.
Removing one redundant code path might be worthwhile in terms of maintainability even if it does imply a slight performance hit for the small workload case.

dmehring: I understand the maintainability issue. I'll run a test on a large dataset to see if the overhead is significant.

ms/MSOper/MSMetaData.cc
juliantaylor: yes, not +1 for the first iteration... but I'm blind I actually missed that you are subtracting 1 from the end of the chunk in the line below ...

juliantaylor: maybe the code would be a bit clearer if the while loop was just a smaller than and not subtract one here?
but thats probably just my preference of having while loops similar to c-style for loops, if you like it like that its fine.

gervandiepen:
I haven't looked at what the code is actually doing. But if it is accessing the same table from multiple threads, I have to warn you that the table internals are not thread-safe.
I'm thinking of making it possible to access the same table from multiple threads, but it is not straightforward due to the internal buffering done.

dmehring:
The data structures that are being simultaneously accessed (read-only) are casa::Vectors which are copies of table columns; the table itself is not being accessed by multiple threads.

juliantaylor:
it looks correct in how it works but it might still be a good idea to check with a race detector to see if there are some nasty surprises in some used datastructures. I can do that next week, it would help if you can provide a testcase but I assume just calling listobs on some alma ms should do it?

dmehring:
That would be great, thanks. The committed listobs interface doesn't support the parallel version yet, although the version in my sandbox does. So, I can provide you with that if you'd like, or you can just create a small test program that essentially does:

MeasurementSet ms("my.ms")
// for this test, the 100 doesn't really matter
MSMetaData md(&ms, 100);
// configure parallel processing
md.setParallel(True);
// this will run the parallelized method
// True means show a progress meter
md.getSubScanProperties(True);

ms/MSOper/MSMetaData.cc
dmehring: For reasons I don't immediately understand, the original serial method takes about 8 minutes to run on my machine, whereas running the multi-threaded method in serial mode takes about 2.5 minutes. So I think I can safely discard the original method.

dmehring:
My timings indicate that on a data set with 250 million rows, running the new method serially takes about 150 seconds, whereas running it in parellel takes 20 seconds on my 8 core Xeon box, so the speed up is substantial.

dmehring:
pushed. Hopefully this covers everything we've discussed thus far. If I've missed something, let me know.

dmehring:
@juliantaylor Hi Julian, it's been a couple of days since my most recent commit, so I wanted to check if you have been able to make any progress on this?

ms/MSOper/MSMetaData.cc
```diff
@@ -3399,6 +3399,8 @@ void MSMetaData::_computeScanAndSubScanProperties(
     uInt chunkSize = nrows/nchunks;
     if (nrows % nchunks > 0) {
         ++chunkSize;
+        // integer division
+        nchunks = nrows/chunkSize + 1;
     }
```
juliantaylor: you would have less special cases to if you remove this `if` completely and change the lower loop to

    for (i=0; i < nrows; i+= chunkSize)
    ...
      i , std::min(i +chunkSize, nrows)

dmehring: If I discard that if block, then there are cases for which the fut array wouldn't be sized correctly when it is declared before that loop. It is sized to nchunks.

juliantaylor: isn't the maximum size `nchunks + 1` for when nrows is not divisible by nchunks?

    fut[nchunks + 1] or fut[nchunks + (nrows & nchunks == 0 ? 0 : 1)]

juliantaylor: but thats just the same special case written differently, so yeah not so useful

is the `++chunkSize` still neeeded? I read that as the nchunks++ original so I didn't notice the nchunk issue

dmehring: If chunkSize is not incremented for that case, as currently written, there will be rows at the end of the table which are not processed.

Consider the case for nrows = 10,001 

currently then, chunkSize = 11 and nchunks = 910. In this case, the loop covers all rows.

Without the if block for this case, nchunks is 1000 and chunkSize is 10. In this case, the last row will not be processed.

juliantaylor: no, keep the if but remove only ´++chunkSize`
in that case nchunks is 1001 and chunksize 10, covering all rows

dmehring: ok, that seems to work as far as I can tell.

In that case, the line

nchunks = nrows/chunkSize + 1;

can simply be rewritten as

++nchunks;

do you agree?

dmehring: Actually, the unit test fails when I try that substitution, so it appears I wasn't correct :) .

dmehring: OK I removed the ++chunkSize line, committed, and pushed.

juliantaylor: easy mistake to make with the nchunks++, doesn't work due to the maximum number of 1000 chunks. Good the testsuite catches that.
One could also write it like this:

```
uInt chunkSize = max(1u, nrows / 1000u);
uInt nchunks = nrows / chunkSize;
if (nrows % chunkSize != 0) {
    nchunks++;
}
```

ms/MSOper/MSMetaData.h
```diff
@@ -564,7 +564,23 @@ class MSMetaData {
 
     const MeasurementSet* getMS() const { return _ms; }
 
-protected:
+    // If set to true, enable parallel computations where possible,
+    // else, always do serial computations.
+    void setParallel(Bool b) {
```
juliantaylor: this needs to be implemented in the .cc file, when its in the header the user of the header has to enable openmp regardless on how casacore itself is built.

juliantaylor:
currently the default CASA casacore configuration is without openmp.
With this change it probably makes sense to change that, for that add

    set(USE_OPENMP ON)

somewhere around line 87 in CMakeList.txt

dmehring:
Done and pushed. Also added note re: enabling parallel processing at top of .h file.

gervandiepen:
Note that the implementation of setParallel does not catch the case where
only 1 thread is available (e.g. if OMP_NUM_THREADS was set to 1).
I'm wondering why setParallel is needed and why the OMP pragmas are not
always used.

juliantaylor:
in the one thread case `setParallel` does nothing as the openmp pragma will do always do nothing.

but I do agree the function seems redundant, it could be useful to allow users to turn off parallelization independent of how its actually implemented (openmp or not) but I'm not sure that if that is really needed.

dmehring:
Removed. It made more sense when there were separate implementations for parallel and non-parallel processing.

juliantaylor:
thanks, threadsanitizer reports no issues and it also shows nice speedup on already moderate sized data, very nice.

dmehring:
Thanks for your help and suggestions.

PR #387, opened 2016-04-01
revert CASA_BUILD defaults to no threads

schiebel:
difficulties encountered with RHEL5 + fftw + threads (back out threads)... for CASA_BUILD

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,7 +84,7 @@ if( CASA_BUILD )
       set(DATA_DIR "%CASAROOT%/data")
    endif( )
    set(USE_FFTW3 ON)
-   set(USE_THREADS ON)
+   set(FFTW3_DISABLE_THREADS ON)
    ###
    ### after CASA switches from ASAP to libsakura, CASA will no longer
    ### use casacore's python module and it will no longer use boost...


juliantaylor:
this will essentially just disable the Mutex class? I don't think that is a good idea.
actually casabuild does not enable openmp in casacore right? that is USE_OPENMP

schiebel:
No... you said openmp was not needed w/ casacore (for our purposes)... as far as I know, this is the default settings (i.e. no openmp, no threads, fftw) we had before we switched to github casacore...

juliantaylor:
yes, openmp is not (yet) needed but grepping the source `USE_THREAD` only does one thing, turn Mutex into a no-op.
That does not seem like a good idea to me so I think having threads on is better.

schiebel:
Yes, I see what you mean... I assumed that setting USE_THREADS=OFF would turn off all threading. I think you're right that it does not. I think we also need to add FFTW3_DISABLE_THREADS=OFF. With these two set to OFF, libpthread is not explicitly linked into any casacore library. There are still dependencies upon libpthread, but these dependencies are the result of implicit dependencies of other libraries.

There is no need for mutexes if threads are not used.

schiebel:
I pushed up the addition of FFTW3_DISABLE_THREADS=OFF, but it's possible that we still want threaded fftw libraries even if casacore is not threaded...

JimNrao:
I think the USE_THREADS is more to allow CasaCore clients to be able to 
use threads than it is for CC itself to use threads.  CasaCore uses lots 
of static variables and without the mutexes the odds of using CC 
successfully in a multithreads app are somewhat less than 1.0.

On 04/01/2016 01:12 PM, schiebel wrote:
>
> I pushed up the addition of FFTW3_DISABLE_THREADS=OFF, but it's 
> possible that we still want threaded fftw libraries even if casacore 
> is not threaded...
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/casacore/casacore/pull/387#issuecomment-204528025>Web 
> Bug from 
> https://github.com/notifications/beacon/AGPGtBLeSj25Oxxq_oFduIDI94ZRA3rxks5pzW4OgaJpZM4H92f6.gif
>

-- 
Jim Jacobs

phone: 835-7235
office: 301

schiebel:
Perhaps @JimNrao , but by default, USE_THREADS is OFF, and until I submitted my last pull request, we had never used it (to my knowledge).

gervandiepen:
USE_THREADS=YES if USE_OPENMP=YES.
As Jim says, mutexes are needed to ensure proper initialization of statics
in a multi-threaded environment (thus also with OPENMP).

schiebel:
Thanks @gervandiepen , we set USE_OPENMP=NO so I guess we can set USE_THREADS=NO too

schiebel:
One last comment, if we understand 'USE_THREADS=YES' to mean "protect our shared state if threads are used", then 'USE_OPENMP=YES' should automatically imply 'USE_THREADS=YES'. In fact, it seems like with this understanding of 'USE_THREADS' (protect shared state), its value is always completely determined by whether the user actually wants threads or not (I do not know if openmp is the only way to introduce threads into casacore or not).

juliantaylor:
enabling openmp (USE_OPENMP) also enables the mutex (USE_THREADS). There is basically no reason to disable it unless you fear a performance impact when you are not using threads. A measurable performance impact is very unlikely, especially as modern hardware can do lock elision.

schiebel:
@juliantaylor based upon your argument, there should be no USE_THREADS command line switch. However, a single-threaded build of casacore seems like the one case where no mutex is required.

juliantaylor:
as el5 support will be dropped soon from CASA I guess we can close this

PR #390, opened 2016-04-07
Improve FindNUMPY.cmake when headers are not installed (fixes #382)

### LINKED ISSUE #382: Build failure: missing arrayobject.h
I tried building casacore and received the following error:
```
In file included from /Users/adam/repos/casacore/casacore/python/Converters/PycArray.tcc:32:0,
                 from /Users/adam/repos/casacore/python/Converters/PycArray.cc:30:
/Users/adam/repos/casacore/casacore/python/Converters/PycArrayNP.h:38:31: fatal error: numpy/arrayobject.h: No such file or directory
 #include <numpy/arrayobject.h>
                               ^
compilation terminated.
make[2]: *** [python/CMakeFiles/casa_python.dir/Converters/PycArray.cc.o] Error 1
make[1]: *** [python/CMakeFiles/casa_python.dir/all] Error 2
make: *** [all] Error 2
```

Any suggestions?  I may not have flex or bison, but this looks like a numpy issue and I certainly have numpy installed.

tammojan:


juliantaylor:
very weird, you have to patch numpy in order for np.get_include() to return none, but apparently someone though that was a good idea so we might as well detect it.

PR #392, opened 2016-04-08
don't fail when openmp is not supported

juliantaylor:
Older macs have no openmp, just warn in this case instead of failing as it is optional and only used in very few parts. This also helps keeping installation instructions more uniform.


juliantaylor:
actual fix coming soon, just testing the failure on travis now.

juliantaylor:
fix posted, merging after tests to fix casa mac builds
if wished I can follow-up and only make it non-fatal in CASA_BUILD  configuration

PR #395, opened 2016-04-12
CAS-8509, increase effeciency of getScanKeys() in certain instances

dmehring:
This prevents an unnecessary table scan to recompute scan keys when they have already been computed and cached in the _scanProperties map.

### DIFF
--- a/ms/MSOper/MSMetaData.cc
+++ b/ms/MSOper/MSMetaData.cc
@@ -1247,12 +1247,23 @@ std::set<ScanKey> MSMetaData::getScanKeys() const {
         return _scanKeys;
     }
     std::set<ScanKey> scanKeys;
-    std::set<SubScanKey> subScanKeys = _getSubScanKeys();
-    std::set<SubScanKey>::const_iterator iter = subScanKeys.begin();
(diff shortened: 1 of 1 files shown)


ms/MSOper/MSMetaData.cc
```diff
@@ -1247,12 +1247,23 @@ std::set<ScanKey> MSMetaData::getScanKeys() const {
...
-        scanKeys.insert(scanKey(*iter));
-        ++iter;
+    if (_scanProperties) {
+        // scan keys already exist and are cached, they just need to be
+        // harvested
+        map<ScanKey, ScanProperties>::const_iterator iter = _scanProperties->begin();
+        map<ScanKey, ScanProperties>::const_iterator end = _scanProperties->end();
+        for (; iter!=end; ++iter) {
+            scanKeys.insert(iter->first);
+        }
```
juliantaylor: I see no reason this needs to be once a for loop and once a while loop, please choose one.

dmehring: I've been migrating to for loops.

juliantaylor:
thanks

PR #397, opened 2016-04-13
direction cosine output in MeasUDF

### LINKED ISSUE #396: Support direction cosine output in TaQL's measure conversions
It would be nice if TaQL's meas UDF can return the result of direction conversions as direction cosines instead of angles.

gervandiepen:
MeasUDF (used in TaQL) has function DIRCOS to get results as direction cosines
Close #396


tammojan:
I cancelled the PR build, since a push build is also running.

PR #398, opened 2016-04-13
CAS-8511: Use subscan props in getNRowMap() if possible

dmehring:
If _subScanProperties has been computed already, the necessary map can be gotten from that structure, avoiding a full table scan.


dmehring:
@juliantaylor would you please review this small change? Thanks.

PR #399, opened 2016-04-13
CAS-6733 fix infinite loop bug

dmehring:
Fix bug by switching from while to for loop. Also remove useless verbose parameter. Migrate some calls to use appropriate MSMetaData methods. postfix ++ -> prefix ++ in loops.


juliantaylor:
thanks

PR #400, opened 2016-04-15
CAS-8497 implement modifications to uvfits AN read/write rules

dmehring:
New UVFITS AN table writing rules:

1. Antenna positions are ITRF, with geocentric origin, no subtraction of a fiducial "array center" is done
   => Array position ARRAYX, ARRAYY, ARRAYZ are always written as 0.
2. Antenna coordinate systems are right-handed, => XYZHAND is always written as "RIGHT" 
   (which I believe is what AIPS hopes for if I've read the uvfits spec correctly). 

Reading rules:
Antenna positions are not implicitly rotated or inverted unless user explicitly specifies that they should be rotated and other conditions are met (the added configuration parameter allows for backward compatibility with UVFITS files written prior to this change).

### DIFF
--- a/msfits/MSFits/MSFitsInput.cc
+++ b/msfits/MSFits/MSFitsInput.cc
@@ -1764,6 +1764,24 @@ void MSFitsInput::fillAntennaTable(BinaryTable& bt) {
     arrayXYZ(1) = bt.getKeywords().asdouble("ARRAYY");
     arrayXYZ(2) = bt.getKeywords().asdouble("ARRAYZ");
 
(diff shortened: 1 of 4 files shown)


JimNrao:
Whoe, if anyone, coordinate the UVFITS area?  I gather its distinct from the regular FITS region of the code?

dmehring:
If you're asking me, I don't know. Yes it's different from fits image code.

JimNrao:
I'll see if I can reopen this one so it can be merged back in.

PR #402, opened 2016-04-18
improve performance of MSSummary and MSMetaData

dmehring:
1. create MSMetaData::getFieldCodes()
2. avoid unnecessary table scan in MSMetaData::_getSubScanKeys() if _subScanProperties data structure has previously been cached
3. avoid unnecessary table scan in MSMetaData::getUniqueFiedIDs() if _subScanProperties data structure has previously been cached 

Performance increase of 15-20% realized.

### DIFF
--- a/ms/MSOper/MSSummary.cc
+++ b/ms/MSOper/MSSummary.cc
@@ -159,12 +159,10 @@ void MSSummary::list (LogIO& os, Record& outRec, Bool verbose,
 {
     // List a title for the Summary
     listTitle (os);
-
     // List the main table as well as the subtables in a useful order and format
     listWhere (os,verbose);
     listWhat (os,outRec, verbose, fillRecord);
     listHow (os,verbose, oneBased);
-
     // These aren't really useful (yet?)
     //  listSource (os,verbose);
     //  listSysCal (os,verbose);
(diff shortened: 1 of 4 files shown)


ms/MSOper/MSSummary.cc
```diff
@@ -1048,41 +1023,52 @@ void MSSummary::listField (LogIO& os, Record& outrec,  Bool verbose, Bool fillRe
...
-            uInt fld=fieldId(i);
-            if (fld<msFC.phaseDir().nrow()) {
-                MDirection mRaDec=msFC.phaseDirMeas(fld);
+        vector<MDirection> phaseDirs = _msmd->getPhaseDirs();
+        vector<String> fieldNames = _msmd->getFieldNames();
+        vector<String> codes = _msmd->getFieldCodes();
+        std::set<Int>::const_iterator fiter = uniqueFields.begin();
+        std::set<Int>::const_iterator fend = uniqueFields.end();
+        vector<Int> sourceIDs = _msmd->getFieldTableSourceIDs();
+        for (; fiter!=fend; ++fiter) {
```
juliantaylor: this changes iteration order from as stored to sorted by field name right?
I assume this is intended?

juliantaylor: actually nevermind, fiter is the uniquefield int set not the names ... so its sorted the same way

juliantaylor:
looks good, thanks

dmehring:
Thanks!

PR #403, opened 2016-04-18
Mod to allow setting tile cache using hypercube number instead of row.

JimNrao:
The tile cache for an StMan is set on a per hypercube basis.  The methods
available from the TiledStManAccessor class use a row number that the
StMan then uses to determine the appropriate hypercube.  For virtually
concatenated databases, it's handy to be able to walk through the
hypercubes without regard to the row number.  TiledStManAccessor already
provides the number of tiles, so the new feature allows looping over the
hypercubes in the StMan and setting each hypercube to have the appropriate
cache size.

I don't fell that this mod does not appear to significantly expose any
additional abstractions from the StMan beyond what TiledStManAccessor
already exposes.

### DIFF
--- a/tables/DataMan/TiledStMan.cc
+++ b/tables/DataMan/TiledStMan.cc
@@ -466,6 +466,14 @@ void TiledStMan::setCacheSize (uInt rownr, uInt nbuckets, Bool forceSmaller)
     hypercube->setCacheSize (nbuckets, forceSmaller, True);
 }
 
+void TiledStMan::setHypercubeCacheSize (uInt hypercube, uInt nbuckets, Bool forceSmaller)
+{
+    // Set the cache size (in buckets) for the given hypercube.
+    TSMCube* tsmCube = getTSMCube (hypercube);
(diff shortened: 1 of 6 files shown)


gervandiepen:
It would be good if the new functionality is tested in e.g. tTiledShapeStMan.

JimNrao:
That is a good idea.  Looks like I can piggyback on the code in that 
without having to figure out for myself how to create a test table. 
Thanks for the suggestion.

On 04/18/2016 11:52 PM, Ger van Diepen wrote:
>
> It would be good if the new functionality is tested in e.g. 
> tTiledShapeStMan.
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/casacore/casacore/pull/403#issuecomment-211740400>Web 
> Bug from 
> https://github.com/notifications/beacon/AGPGtGiIJOrx2205AKl6RUUB6uOsq81wks5p5G2ggaJpZM4IKIs3.gif
>

-- 
Jim Jacobs

phone: 835-7235
office: 301

PR #404, opened 2016-04-19
CAS-8919 modify getRowStats() to use _subScanProps if it is cached

dmehring:
Also add correct spelling of getUniqueFieldIDs() and deprecate getUniqueFiedIDs().

### DIFF
--- a/ms/MSOper/MSMetaData.h
+++ b/ms/MSOper/MSMetaData.h
@@ -85,12 +85,13 @@ public:
         // the subscan and that spwID
         map<uInt, Quantity> meanInterval;
         Quantity meanExposureTime;
-        uInt nrows;
         std::set<uInt> spws;
         // number of rows for each spectral window
         map<uInt, uInt> spwNRows;
         std::set<Int> stateIDs;
-        std::map<Double, TimeStampProperties> timeProps;
+        map<Double, TimeStampProperties> timeProps;
+        // number of cross-correlation rows.
+        uInt xcRows;
     };
 
     // construct an object which stores a pointer to the MS and queries the MS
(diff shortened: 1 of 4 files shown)


ms/MSOper/MSMetaData.h
```diff
@@ -85,12 +85,13 @@ class MSMetaData {
         // the subscan and that spwID
         map<uInt, Quantity> meanInterval;
         Quantity meanExposureTime;
-        uInt nrows;
         std::set<uInt> spws;
         // number of rows for each spectral window
         map<uInt, uInt> spwNRows;
         std::set<Int> stateIDs;
-        std::map<Double, TimeStampProperties> timeProps;
+        map<Double, TimeStampProperties> timeProps;
```
juliantaylor: it needs to be std::map, that you can use it without is an error, std namespace should not be set default in headers

ms/MSOper/MSMetaData.h
```diff
@@ -85,12 +85,13 @@ class MSMetaData {
...
+        // number of cross-correlation rows.
+        uInt xcRows;
```
juliantaylor: this should maybe be placed next to acRows as it is related
as you are breaking binary compat anyway be removing nrows you don't need to add at the bottom.

ms/MSOper/MSMetaData.h
dmehring: This is the first I'm hearing about this. Why is it not permitted to use using namespace std; ?

juliantaylor: because this is a header, using namespace cannot be reverted so you are forcing it on everybody including the header.

juliantaylor: the actual culprit is in some other header, but that is no reason to stick to bad style, also it avoids an error in for the case when the cause is fixed.

dmehring: OK, fair enough. Thanks.

dmehring:
moved xcRows as requested. Changed all map to std::map in header file.

dmehring:
Thanks!

PR #406, opened 2016-04-21
CAS-8523 modify MSSummary::listSpectralAndPolInfo() to be more effecient

dmehring:
Also add MSMetaData::getUniqueDataDescIDs()

### DIFF
--- a/ms/MSOper/MSMetaData.cc
+++ b/ms/MSOper/MSMetaData.cc
@@ -4549,6 +4549,25 @@ Bool MSMetaData::_hasFieldID(const Int fieldID) const {
     return uniqueFields.find(fieldID) != uniqueFields.end();
 }
 
+std::set<Int> MSMetaData::getUniqueDataDescIDs() const {
+    // this method is responsible for setting _uniqueDataDescIDs
+    if (_uniqueDataDescIDs.empty()) {
+        if (_subScanProperties) {
+            map<SubScanKey, SubScanProperties>::const_iterator iter = _subScanProperties->begin();
+            map<SubScanKey, SubScanProperties>::const_iterator end = _subScanProperties->end();
+            for (; iter!=end; ++iter) {
+                const std::set<uInt>& ddIDs = iter->second.ddIDs; 
+                _uniqueDataDescIDs.insert(ddIDs.begin(), ddIDs.end());
+            }
+        }
+        else {
+            SHARED_PTR<Vector<Int> > allDDIDs = _getDataDescIDs();
+            _uniqueDataDescIDs.insert(allDDIDs->begin(), allDDIDs->end());
+        }
+    }
+    return _uniqueDataDescIDs;
+}
+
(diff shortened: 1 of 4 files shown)


ms/MSOper/MSSummary.cc
```diff
@@ -1558,11 +1551,11 @@ void MSSummary::listSpectralAndPolInfo (
...
         os.output().precision(9);
         // order by spwid, not ddid, CAS-7376
-        Vector<Int>::const_iterator iter = spwIds.begin();
-        Vector<Int>::const_iterator end = spwIds.end();
+        Vector<uInt>::const_iterator iter = spwIds.begin();
+        Vector<uInt>::const_iterator end = spwIds.end();
         std::vector<std::set<uInt> > spwToDDID = _msmd->getSpwToDataDescriptionIDMap();
-        std::set<uInt> uDDIDSet(uddId.begin(), uddId.end());
-        vector<uInt> ddToPolID = _msmd->getDataDescIDToPolIDMap();
+        std::set<uInt> uDDIDSet(ddId.begin(), ddId.end());
```
juliantaylor: this copy seems unnecessary as ddId now already is a set

ms/MSOper/MSMetaData.cc
```diff
@@ -4549,6 +4549,25 @@ Bool MSMetaData::_hasFieldID(const Int fieldID) const {
...
+                const std::set<uInt>& ddIDs = iter->second.ddIDs; 
+                _uniqueDataDescIDs.insert(ddIDs.begin(), ddIDs.end());
+            }
+        }
+        else {
+            SHARED_PTR<Vector<Int> > allDDIDs = _getDataDescIDs();
+            _uniqueDataDescIDs.insert(allDDIDs->begin(), allDDIDs->end());
+        }
+    }
+    return _uniqueDataDescIDs;
```
juliantaylor: the ddi is once a uInt and once a Int, can this be cleaned up?

dmehring:
Thanks. Changes made as requested. Shortened tMSMetaData test.

juliantaylor:
why is ut uInt from the subscanproperties and Int from `getDataDescIds()`

also please squash the commits if you do not provide a useful commit message

dmehring:
History mostly. The table column type is int not uint. However fillers are supposed to put nonnegative values in this column. There has been a history of other such columns with fillers putting in negative values which has caused issues. However, it seems unlikely this would happen for data description id and it would be wrong if it did, although we have allowed such things with the field id column. I can change the return of getDataDescIds, but this may cause a domino effect of other things needing to be changed. 

Re your comment re "squash". I'm not sure I understand.

dmehring:
correction, I think I was wrong, it was the STATE_ID, not the FIELD_ID column in which at least one filler was putting values of -1, and was not writing a STATE subtable at all. As I recall, code such as MSMetaData::nStates() was modified to allow this.

PR #408, opened 2016-04-25
CAS-8529 improve performance of MSMetaData::_getExposureTimes()

dmehring:
This decreases the run time of MSMetaData::_getExposureTimes()
    by a factor of 4 to 5 on large data sets.
Also, add a unit test for MSMetaData::getFirstExposureTimeMap()
    which verifies this change

### DIFF
--- a/measures/TableMeasures/ScalarQuantColumn.h
+++ b/measures/TableMeasures/ScalarQuantColumn.h
@@ -196,11 +196,10 @@ public:
 
   // get the column as a Quantum<Vector<T> >. If <src>unit</src> is
   // not empty, the returned Quantum will have that unit. Else if
-  // a unit was specified at construction of this object, the returned
-  // Quantum will have that unit. Else if the units are variable, the
-  // values in the returned Vector have been converted to the
-  // unit of the 0th row entry. Otherwise, the units of the returned
-  // Quantum are the units specified in the column descriptor.
+  // the units are variable, the values in the returned Vector
+  // have been converted to the unit of the 0th row entry. Otherwise,
+  // the units of the returned Quantum are the units specified in
+  // the column descriptor.
   SHARED_PTR<Quantum<Vector<T> > > getColumn(const Unit& unit="") const;
 
 protected:
--- a/measures/TableMeasures/ScalarQuantColumn.h
+++ b/measures/TableMeasures/ScalarQuantColumn.h
@@ -194,13 +194,14 @@ public:
   // Throw an exception if the object is null.
   void throwIfNull() const;
 
-  // get the column as a Quantum<Vector<T> >. If the units are variable, the
+  // get the column as a Quantum<Vector<T> >. If <src>unit</src> is
+  // not null, the returned Quantum will have that unit. Else if
+  // itsConv is True, the returned Quantum will have itsUnitOut as
+  // the unit. Else if the units are variable, the
   // values in the returned Vector have been converted to the
   // unit of the 0th row entry. Otherwise, the units of the returned
-  // Quantum are itsUnit (no conversion to itsUnitOut is attempted if
-  // itsConvOut is True, because of compile issues if the column type
-  // is Complex).
-  SHARED_PTR<Quantum<Vector<T> > > getColumn() const;
+  // Quantum are itsUnit.
+  SHARED_PTR<Quantum<Vector<T> > > getColumn(const Unit* unit=NULL) const;
 
 protected:
   //# Quantum column's units (if units not variable)
--- a/measures/TableMeasures/ScalarQuantColumn.h
+++ b/measures/TableMeasures/ScalarQuantColumn.h
@@ -195,13 +195,13 @@ public:
   void throwIfNull() const;
 
   // get the column as a Quantum<Vector<T> >. If <src>unit</src> is
-  // not null, the returned Quantum will have that unit. Else if
-  // itsConv is True, the returned Quantum will have itsUnitOut as
-  // the unit. Else if the units are variable, the
+  // not empty, the returned Quantum will have that unit. Else if
+  // a unit was specified at construction of this object, the returned
+  // Quantum will have that unit. Else if the units are variable, the
   // values in the returned Vector have been converted to the
   // unit of the 0th row entry. Otherwise, the units of the returned
-  // Quantum are itsUnit.
-  SHARED_PTR<Quantum<Vector<T> > > getColumn(const Unit* unit=NULL) const;
+  // Quantum are the units specified in the column descriptor.
+  SHARED_PTR<Quantum<Vector<T> > > getColumn(const Unit& unit="") const;
 
 protected:
   //# Quantum column's units (if units not variable)
--- a/measures/TableMeasures/ScalarQuantColumn.h
+++ b/measures/TableMeasures/ScalarQuantColumn.h
@@ -194,6 +194,14 @@ public:
   // Throw an exception if the object is null.
   void throwIfNull() const;
 
+  // get the column as a Quantum<Vector<T> >. If the units are variable, the
+  // values in the returned Vector have been converted to the
+  // unit of the 0th row entry. Otherwise, the units of the returned
+  // Quantum are itsUnit (no conversion to itsUnitOut is attempted if
+  // itsConvOut is True, because of compile issues if the column type
+  // is Complex).
+  SHARED_PTR<Quantum<Vector<T> > > getColumn() const;
+
 protected:
   //# Quantum column's units (if units not variable)
   Unit itsUnit;
--- a/measures/TableMeasures/ScalarQuantColumn.tcc
+++ b/measures/TableMeasures/ScalarQuantColumn.tcc
@@ -235,18 +235,20 @@ void ScalarQuantColumn<T>::put (uInt rownr, const Quantum<T>& q)
 }
 
 template<class T>
-SHARED_PTR<Quantum<Vector<T> > > ScalarQuantColumn<T>::getColumn(const Unit* unit) const {
+SHARED_PTR<Quantum<Vector<T> > > ScalarQuantColumn<T>::getColumn(const Unit& unit) const {
     SHARED_PTR<Quantum<Vector<T> > > qv;
-    const Unit* unitOut = unit ? unit
-        : itsConvOut ? &itsUnitOut : NULL;
+    Unit unitOut;
+    Bool convert = ! unit.empty() || itsConvOut;
+    if (convert) {
+        unitOut = unit.empty() ? itsUnitOut : unit;
+    }
     if (itsUnitsCol) {
         Vector<String> units = itsUnitsCol->getColumn();
         Vector<String>::const_iterator uiter = units.begin();
-        Unit first(*uiter);
-        if (! unitOut) {
-            unitOut = &first;
+        if (! convert) {
+            unitOut = Unit(*uiter);
         }
-        qv.reset(new Quantum<Vector<T> >(Vector<T>(), *unitOut));
+        qv.reset(new Quantum<Vector<T> >(Vector<T>(), unitOut));
         Vector<T>& val = qv->getValue();
         itsDataCol->getColumn(val);
         typename Vector<T>::iterator viter = val.begin();
@@ -255,16 +257,17 @@ SHARED_PTR<Quantum<Vector<T> > > ScalarQuantColumn<T>::getColumn(const Unit* uni
         for (; viter != vend; ++viter, ++uiter) {
             q.setValue(*viter);
             q.setUnit(*uiter);
-            *viter = q.getValue(*unitOut);
+            *viter = q.getValue(unitOut);
         }
     }
     else {
         qv.reset(new Quantum<Vector<T> >(Vector<T>(), itsUnit));
         Vector<T>& val = qv->getValue();
         itsDataCol->getColumn(val);
-        if (unitOut) {
-            val = qv->getValue(*unitOut);
-            qv->setUnit(*unitOut);
+        if (convert) {
+            qv->convert(unitOut);
+            //val = qv->getValue(unitOut);
+            //qv->setUnit(unitOut);
         }
     }
     return qv;
--- a/measures/TableMeasures/ScalarQuantColumn.tcc
+++ b/measures/TableMeasures/ScalarQuantColumn.tcc
@@ -234,7 +234,34 @@ void ScalarQuantColumn<T>::put (uInt rownr, const Quantum<T>& q)
   }
 }
 
-} //# NAMESPACE CASACORE - END
+template<class T>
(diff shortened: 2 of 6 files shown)


ms/MSOper/MSMetaData.cc
```diff
@@ -2174,18 +2174,15 @@ SHARED_PTR<QVD> MSMetaData::_getExposureTimes() const {
...
-    for (; iter!=end; ++iter, ++i) {
-        *iter = exposure(i).getValue();
-    }
-    SHARED_PTR<QVD> ex(new QVD(v, exposure.getUnits()));
+    Unit units = ScalarQuantColumn<Double>(
+        *_ms, MeasurementSet::columnName(MSMainEnums::EXPOSURE)
+    ).getUnits();
+    SHARED_PTR<QVD> ex(new QVD(Vector<Double>(), units));
+    Vector<Double>& val = ex->getValue();
+    col.getColumn(val);
```
juliantaylor: don't you need to check `isUnitsVariable()` before doing this?

juliantaylor:
it seems to me that it would be a better idea to add a getColumn method to `ScalarQuantColumn` which does this directly.

ms/MSOper/MSMetaData.cc
dmehring: I don't think so. AFAIK, there isn't a separate units column associated with this column.

dmehring:
@gervandiepen, @juliantaylor says "it seems to me that it would be a better idea to add a getColumn method to ScalarQuantColumn which does this directly."

Ger, do you have any objections to that?

dmehring:
@juliantaylor We seem to have a conflict between casacore's QVector and qt's QVector

When I add 

#include <casacore/casa/Quanta/QVector.h>

to casacore/measures/TableMeasures/ScalarQuantColumn.h

and try and compile our code/ tree, I get the error listed below. Any idea on how to avoid that?

In file included from /usr/lib64/casa/01/include/QtCore/qpair.h:45:0,
                 from /usr/lib64/casa/01/include/QtCore/qhash.h:49,
                 from /usr/lib64/casa/01/include/QtCore/qdebug.h:46,
                 from /usr/lib64/casa/01/include/QtCore/QDebug:1,
                 from /export/data_1/dmehring/casa_eclipse/projects/CAS-8529/code/guitools/Feather/FeatherThread.cc:34:
/usr/lib64/casa/01/include/QtCore/qdatastream.h:297:41: error: reference to 'QVector' is ambiguous
 QDataStream& operator>>(QDataStream& s, QVector<T>& v)
                                         ^
In file included from /usr/lib64/casa/01/include/QtCore/qobject.h:48:0,
                 from /usr/lib64/casa/01/include/QtCore/qthread.h:45,
                 from /usr/lib64/casa/01/include/QtCore/QThread:1,
                 from /export/data_1/dmehring/casa_eclipse/projects/CAS-8529/code/guitools/Feather/FeatherThread.h:30,
                 from /export/data_1/dmehring/casa_eclipse/projects/CAS-8529/code/guitools/Feather/FeatherThread.cc:27:
/usr/lib64/casa/01/include/QtCore/qstring.h:81:29: note: candidates are: template<class T> class QVector
 template <typename T> class QVector;
                             ^
In file included from /export/data_1/dmehring/casa_eclipse/projects/CAS-8529/linux/include/casacore/measures/TableMeasures/ScalarQuantColumn.h:34:0,
                 from /export/data_1/dmehring/casa_eclipse/projects/CAS-8529/linux/include/casacore/ms/MeasurementSets/MSAntennaColumns.h:36,
                 from /export/data_1/dmehring/casa_eclipse/projects/CAS-8529/linux/include/casacore/ms/MeasurementSets/MSColumns.h:34,
                 from /export/data_1/dmehring/casa_eclipse/projects/CAS-8529/code/synthesis/TransformMachines/Utils.h:32,
                 from /export/data_1/dmehring/casa_eclipse/projects/CAS-8529/code/synthesis/TransformMachines/CFBuffer.h:34,
                 from /export/data_1/dmehring/casa_eclipse/projects/CAS-8529/code/synthesis/TransformMachines/CFStore2.h:31,
                 from /export/data_1/dmehring/casa_eclipse/projects/CAS-8529/code/synthesis/TransformMachines/CFCache.h:47,
                 from /export/data_1/dmehring/casa_eclipse/projects/CAS-8529/code/synthesis/TransformMachines/FTMachine.h:46,
                 from /export/data_1/dmehring/casa_eclipse/projects/CAS-8529/code/synthesis/MeasurementEquations/Imager.h:44,
                 from /export/data_1/dmehring/casa_eclipse/projects/CAS-8529/code/guitools/Feather/FeatherThread.cc:31:
/export/data_1/dmehring/casa_eclipse/projects/CAS-8529/linux/include/casacore/casa/Quanta/QVector.h:37:26: note:                 template<class T> class casa::QVector
 template <class T> class QVector;

dmehring:
The include line I added is

`#include <casacore/casa/Quanta/QVector.h>`

I seem to have confused github's parser by not escaping the brackets.

juliantaylor:
some or many headers are clobbering the namespace with `using` declarations ... I'll have a look if I can fix it.
can you make due with a forward declaration?

dmehring:
I think it would still be a problem since the tcc would have to include it, wouldn't it? And the tcc is include by the .h file

`#ifndef CASACORE_NO_AUTO_TEMPLATES
#include <casacore/measures/TableMeasures/ScalarQuantColumn.tcc>
#endif //# CASACORE_NO_AUTO_TEMPLATES
#endif
`

But I probably don't need the QVector specialization, so I can probably make due with its base class `Quantum<Vector<Double>>` and avoid this issue.

juliantaylor:
the issue should be fixed in casa r36858
recall I nitpicked about `using namespace` before, this is why

dmehring:
I recall :) . I don't have an immediate need to switch back to QVector, so feel free to review what I've just pushed.

measures/TableMeasures/ScalarQuantColumn.h
```diff
@@ -194,6 +194,14 @@ template<class T> class ScalarQuantColumn
   // Throw an exception if the object is null.
   void throwIfNull() const;
 
+  // get the column as a Quantum<Vector<T> >. If the units are variable, the
+  // values in the returned Vector have been converted to the
+  // unit of the 0th row entry. Otherwise, the units of the returned
+  // Quantum are itsUnit (no conversion to itsUnitOut is attempted if
+  // itsConvOut is True, because of compile issues if the column type
+  // is Complex).
```
juliantaylor: what compile issue exactly?
I think an optional unit argument to this function would be useful

measures/TableMeasures/ScalarQuantColumn.tcc
```diff
@@ -234,7 +234,34 @@ void ScalarQuantColumn<T>::put (uInt rownr, const Quantum<T>& q)
...
+        Vector<T>& val = qv->getValue();
+        itsDataCol->getColumn(val);
+        typename Vector<T>::iterator viter = val.begin();
+        typename Vector<T>::iterator vend = val.end();
+        Quantum<T> q;
+        for (; viter != vend; ++viter, ++uiter) {
+            q.setValue(*viter);
+            q.setUnit(*uiter);
+            q.convert(unitOut);
+            *viter = q.getValue();
```
juliantaylor: is there a reason `getvalue(Unit)` is not used?

measures/TableMeasures/ScalarQuantColumn.h
dmehring: Quantum.tcc

This will not compile if Qtype is `Vector<Complex>`. Perhaps there is a way around it, but on the other hand, this is starting to go significantly beyond the scope of what I need this for.

`322       qVal = Qtype (qVal *
323               (qUnit.getValue().getFac()/s.getValue().getFac()));
`

measures/TableMeasures/ScalarQuantColumn.tcc
dmehring: There is no need. the previous convert() call assures it's in the correct units.

tammojan:
> @gervandiepen, @juliantaylor says "it seems to me that it would be a better idea to add a getColumn method to ScalarQuantColumn which does this directly."
> 
> Ger, do you have any objections to that?

@gervandiepen is on vacation until end of next week. I don't see any objections to add the getColumn function to ScalarQuantColumn.

dmehring:
The previous commit combines convert() and getValue() into a single call that has the same result.

measures/TableMeasures/ScalarQuantColumn.h
juliantaylor: can you be more specific, this compiles casa fine for me:

```patch
--- a/measures/TableMeasures/ScalarQuantColumn.tcc
+++ b/measures/TableMeasures/ScalarQuantColumn.tcc
@@ -240,7 +240,7 @@ SHARED_PTR<Quantum<Vector<T> > > ScalarQuantColumn<T>::getCo
     if (itsUnitsCol) {
         Vector<String> units = itsUnitsCol->getColumn();
         Vector<String>::const_iterator uiter = units.begin();
-        const Unit unitOut = *uiter;
+        const Unit unitOut = itsConvOut ? itsUnitOut : *uiter;
         qv.reset(new Quantum<Vector<T> >(Vector<T>(), unitOut));
         Vector<T>& val = qv->getValue();
         itsDataCol->getColumn(val);
```

dmehring:
I can't immediately reproduce the compile issue. The latest commit avoids the call to convert, allows user to specify units, and if not specified, will convert to itsUnitOut if it exists.

measures/TableMeasures/ScalarQuantColumn.h
```diff
@@ -194,6 +194,15 @@ template<class T> class ScalarQuantColumn
...
   void throwIfNull() const;
 
+  // get the column as a Quantum<Vector<T> >. If <src>unit</src> is
+  // not null, the returned Quantum will have that unit. Else if
+  // itsConv is True, the returned Quantum will have itsUnitOut as
+  // the unit. Else if the units are variable, the
+  // values in the returned Vector have been converted to the
+  // unit of the 0th row entry. Otherwise, the units of the returned
+  // Quantum are itsUnit.
+  SHARED_PTR<Quantum<Vector<T> > > getColumn(const Unit* unit=NULL) const;
```
juliantaylor: using a empty unit as the default is nicer than a pointer to a unit

the documentation should not talk in terms of private variables, better would be for example "if unit is empty it will be converted to the unit associated to the column or the unit of the 0th row if `isUnitVariable()` is true"

measures/TableMeasures/ScalarQuantColumn.tcc
```diff
@@ -234,7 +234,42 @@ void ScalarQuantColumn<T>::put (uInt rownr, const Quantum<T>& q)
...
+        }
+    }
+    else {
+        qv.reset(new Quantum<Vector<T> >(Vector<T>(), itsUnit));
+        Vector<T>& val = qv->getValue();
+        itsDataCol->getColumn(val);
+        if (unitOut) {
+            val = qv->getValue(*unitOut);
+            qv->setUnit(*unitOut);
+        }
```
juliantaylor: is this correct? setUnit does not convert, if the branch above is correct
this case needs to be handled in the other branch

dmehring: Correct, setUnit() does not convert. Line 266 does the necessary conversion of the numerical values to unitOut. Line 267 just sets the unit to be consistent with the values. I don't understand your assertion that it needs to be handled in the other branch.

juliantaylor: just though it would be more consistent, but thats slower and unit identity comparison looks clumsy so this is fine too

though couldn't it just be `qv->convert(unitOut)`?

dmehring: It could be (and was), but that's what produced the compiler error I mentioned. I'll change it back and see if I can reproduce the compiler error.

dmehring:
This version won't compile because of issues with Quantum<Vector<Complex> >:: convert().

dmehring:
More specifically, tTableQuantum.cc won't compile.

juliantaylor:
weird I don't understand why getValue would work, its basically the same code.

dmehring:
Hmm, well it appears I didn't try compiling with getValue() and Quantum<Vector<Complex> >. That does indeed fail as well. So... I can use convert() and just (again) remove the Quantum<Vector<Complex> > call in the test, but the issue will lurk (which is why I initially didn't want to do any conversions). What do you suggest?

juliantaylor:
There is still the option of doing the conversion in the other branch where we have Quantums of scalars instead of Quantums of collections.

juliantaylor:
this would be my suggestion.

```patch
--- a/measures/TableMeasures/ScalarQuantColumn.tcc
+++ b/measures/TableMeasures/ScalarQuantColumn.tcc
@@ -237,38 +237,27 @@ void ScalarQuantColumn<T>::put (uInt rownr, const Quantum<T>& q)
 template<class T>
 SHARED_PTR<Quantum<Vector<T> > > ScalarQuantColumn<T>::getColumn(const Unit& unit) const {
     SHARED_PTR<Quantum<Vector<T> > > qv;
-    Unit unitOut;
-    Bool convert = ! unit.empty() || itsConvOut;
-    if (convert) {
-        unitOut = unit.empty() ? itsUnitOut : unit;
-    }
-    if (itsUnitsCol) {
-        Vector<String> units = itsUnitsCol->getColumn();
-        Vector<String>::const_iterator uiter = units.begin();
-        if (! convert) {
-            unitOut = Unit(*uiter);
+    if ((itsUnitsCol && itsUnitCol->nrow() > 0) || ! unit.empty()) {
+        Unit unitOut;
+        if (! unit.empty()) {
+            unitOut = unit;
+        }
+        else {
+            unitOut = itsUnitsCol->get(0);
         }
         qv.reset(new Quantum<Vector<T> >(Vector<T>(), unitOut));
         Vector<T>& val = qv->getValue();
         itsDataCol->getColumn(val);
-        typename Vector<T>::iterator viter = val.begin();
-        typename Vector<T>::iterator vend = val.end();
         Quantum<T> q;
-        for (; viter != vend; ++viter, ++uiter) {
-            q.setValue(*viter);
-            q.setUnit(*uiter);
-            *viter = q.getValue(unitOut);
+        for (uInt i = 0; i < val.size(); ++i) {
+            get(i, q, unitOut);
+            val[i] = q.getValue();
         }
     }
     else {
         qv.reset(new Quantum<Vector<T> >(Vector<T>(), itsUnit));
         Vector<T>& val = qv->getValue();
         itsDataCol->getColumn(val);
-        if (convert) {
-            qv->convert(unitOut);
-            //val = qv->getValue(unitOut);
-            //qv->setUnit(unitOut);
-        }
     }
     return qv;
 }

```

juliantaylor:
thanks

PR #409, opened 2016-04-26
Cas 8525  Fix bug in which min/max pos strings were being determined if min/max pixel location was empty

dmehring:
Two commits:
1. Fix bug in which min/max pos strings were being determined if min/max pixel location was empty.
2. substitute spaces for tabs for indentation.


juliantaylor:
`formatCoordinate` should also raise an error on mismatching input, this doesn't look like a function where one needs to skip boundary checks for performance

dmehring:
Check added as requested, unit test added to verify it.

juliantaylor:
excellent, thank you

dmehring:
Thanks!

PR #410, opened 2016-04-29
CAS-8569 add _getMainScalarColumn() for common way of retrieving col data

dmehring:
Also using col.getColumn(Vector<T>&) provides a noticeable performance
increase over Vector<T> = col.getColumn() for large tables.


juliantaylor:
thanks

dmehring:
Thanks!

PR #412, opened 2016-05-08
Adding dependency on CentOS

shaoguangleo:
Adding all the dependency for casacore on **CentOS**.
And all had test properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,14 @@ $ sudo apt-get install libhdf5-serial-dev libfftw3-dev python-numpy \
     libboost-python-dev
 ```
 
+On CentOS you can install these with:
+ ```bash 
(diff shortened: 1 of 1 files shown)


alxames:
Happen to use CentOS 7, so I can say this list is legit. Not all listed pkgs are required though. Especially cmake-gui is not a dep, but really useful. Bound to save people time, so let's merge.

Crazy idea: Perhaps we should one day have cmake print such a command on error with the missing pkgs for Debian and RH/CentOS based distros.

tammojan:
Thanks both. Indeed better feedback would be nice. Let's leave that on the wishlist for now.

shaoguangleo:
Yes, @alxames **Great** idea to install casacore with dependency hint.

PR #414, opened 2016-05-10
cache brew installation on osx

juliantaylor:


juliantaylor:
turns out ccache was working all along, so I added caching of the brew installation which saves some minutes too

tammojan:
So this is good to go?

tammojan:
Thanks for this, this will speed up testing a lot.

PR #413, opened 2016-05-09
Revert "CAS-8497 implement modifications to uvfits AN read/write rules"

JimNrao:
Reverts casacore/casacore#400

PR #415, opened 2016-05-10
Revert "Revert "CAS-8497 implement modifications to uvfits AN read/write rules""

JimNrao:
Reverts casacore/casacore#413


JimNrao:
Maybe two wrongs will make a right?

PR #416, opened 2016-05-11
Fix typo and add centos version 7

shaoguangleo:
- lapck-devel -> lapack-devel
- All the dependency had been test on CentOS7, **but** on CentOS6, you need compile some source code, for example CentOS6 have the WCSLIB version 4.3, but casacore need version 4.7 at least.


tammojan:
thanks.

PR #422, opened 2016-05-24
CAS-8645 improve MSSummary performance for MS with many fields

dmehring:
Account for ephemeris objects in getPhaseDirs() and improve performance of MSSummary by retrieving entire phaseDir data structure in one go rather than making multiple calls to msmd method.

### DIFF
--- a/ms/MSOper/MSMetaData.cc
+++ b/ms/MSOper/MSMetaData.cc
@@ -1093,7 +1093,6 @@ std::set<String> MSMetaData::getIntentsForField(Int fieldID) {
     return fieldToIntentsMap[fieldID];
 }
 
-
 uInt MSMetaData::nFields() const {
     if (_nFields > 0) {
         return _nFields;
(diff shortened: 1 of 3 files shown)


ms/MSOper/MSMetaData.cc
```diff
@@ -1103,6 +1102,32 @@ uInt MSMetaData::nFields() const {
...
+    ROMSFieldColumns msfc(_ms->field());
+    ROScalarColumn<Int> ephemCol = msfc.ephemerisId();
+    _ephemFields.reset(new std::set<Int>());
+    if (ephemCol.isNull()) {
+        return _ephemFields;
+    }
+    Vector<Int> colData = ephemCol.getColumn();
+    if (! anyTrue(colData >= 0)) {
+        return _ephemFields;
+    }
```
juliantaylor: this fastpath does not seem necessary, it might even be slower than the other loop as it creates a temporary.

dmehring: Are you referring to the test in line 1117?

dmehring:
Assuming the answer is yes to my previous question, I removed that block in my most recent commit.

ms/MSOper/MSMetaData.cc
juliantaylor: yes

dmehring:
Thanks.

PR #425, opened 2016-05-25
CAS-8642: Make fits reader more robust w.r.t. invalid binary tables

dpetry:
Improved code to intercept possible segfault.
Made BinaryTable check the HDU error status before proceeding
with reading the HDU.

### DIFF
--- a/fits/FITS/BinTable.cc
+++ b/fits/FITS/BinTable.cc
@@ -28,6 +28,7 @@
 //# Includes
 #include <casacore/fits/FITS/BinTable.h>
 #include <casacore/fits/FITS/fits.h>
+#include <casacore/fits/FITS/hdu.h>
 #include <casacore/tables/Tables/Table.h>
 #include <casacore/tables/Tables/SetupNewTab.h>
 #include <casacore/tables/DataMan/IncrementalStMan.h>
--- a/fits/FITS/BinTable.cc
+++ b/fits/FITS/BinTable.cc
@@ -72,7 +72,7 @@ BinaryTable::BinaryTable(FitsInput& fitsin, FITSErrorHandler errhandler,
     colNames(0), vatypes_p(0), vaptr_p(0), va_p(0), theheap_p(0)
 {
 
-    AlwaysAssert(err() == HeaderDataUnit::HDUErrs::OK, AipsError);
+    AlwaysAssert(err() == HeaderDataUnit::OK, AipsError);
 
     // is there a heap
     if (pcount()>0) {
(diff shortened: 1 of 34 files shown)


juliantaylor:
c++98 build failed, Bintable.cc is missing a `#include "fits/FITS/hdu.h"` (no clue why c++11 works)

juliantaylor:
ah my bad, it actually needs to be:

    HeaderDataUnit::OK

This is what changed in in c++11, there `HeaderDataUnit::HDUErrs::OK` is allowed

I guess the include can be kept as it is used directly

juliantaylor:
urg please rebase before merging this mess in future

juliantaylor:
I'm actually going to reset this, I have no idea what you even merged

dpetry:
This change was just a few bytes. How can it be "a mess"?
Could we maybe discuss this in person?

juliantaylor:
just look at the difference, 1500 lines changed and 13 commits to change a few bytes?
you have mixed in lots of different prs into this branch somehow

dpetry:
Sorry. No idea how this happened. It was certainly not intentional.
I was just following git instructions.
If you can repair this, please do.

dpetry:
Well, please let me know what you are going to do and if I still need to do anything.
Also, can I delete this branch now or do you need it?

dpetry:
This may be a mess, but the final outcome seems to be the right one.
I just did an update on the NRAO svn copy of casacore and my changes are the only
ones which came through. The other 1500 lines did not.
So at least nothing was actually damaged.

I think what I probably inadvertently did is to recommit a number of already tested and merged pull-requests. So the final outcome was correct. It was just not exactly elegant.

Julian, I tried to contact you by phone to discuss this but didn't reach you.
Please confirm if all is fine from your point of view.

tammojan:
If @juliantaylor has time, I actually prefer him to redo this merge. The commit history on the master should not have commits like "Darn: misread Julian's comment.". It shouldn't be too hard to redo this merge. @juliantaylor , are you still around?

juliantaylor:
I am a bit concerned what happens with the svn export when one resets the branch.
The cli interface log does look reasonable, the criss cross merges just seem to confuse the github gui.

Its probably best to leave it as is now.

tammojan:
Ok, if the only remaining problem is the wording of some commits on master, I can live with that.

dpetry:
If the problem is really only the "Darn", I apologize. It was meant as an entertainment for you.
Sorry if it came across differently.
Anyway, I learned something today and will make sure this kind of merging problem
doesn't happen again. And in casacore comments, I will just say clean, simple sentences which cannot be misunderstood.

tammojan:
Nothing wrong with entertainment of course! It's just that this type of commits should be squashed together to make one neat commit on the master branch.

PR #426, opened 2016-05-25
Add functionality to clone a table column

### LINKED ISSUE #354: Create a function to clone a table column
It is very useful if a column like DATA can be cloned to, say, CORRECTED_DATA using the same column description and data manager properties. The column should be filled from an existing column, either by copying its data or by creating an array with the same shape filled with a constant value.

gervandiepen:
CASA needed functionality to clone a table column, which has been done by adding a few functions to class TableCopy.
- cloneColumn clones a column using the same description and data manager type
- cloneColumnTyped does the same, but can change the data type
- copyColumnData copies the data from the given column to another (e.g., the new one)
- fillArrayColumn fills an array column with the given array
- fillColumnData fills a column from another column or with a given value

The cloning/copying functionality can preserve the tile shape of the original column, for which a preserveTileShape argument has been added to some table functions.

Close #354

### DIFF
--- a/tables/DataMan/VirtualTaQLColumn.cc
+++ b/tables/DataMan/VirtualTaQLColumn.cc
@@ -157,7 +157,7 @@ void VirtualTaQLColumn::prepare()
   TableColumn tabcol (table(), itsColumnName);
   itsExpr = tabcol.keywordSet().asString ("_VirtualTaQLEngine_CalcExpr");
   // Compile the expression.
-  TaQLResult res = tableCommand ("calc from $1 calc " + itsExpr, table());
+  TaQLResult res = tableCommand ("calc " + itsExpr + " from $1" , table());
   itsNode = new TableExprNode(res.node());
   // Check if the expression type matches the column type.
   if (itsNode->isScalar() == itsIsArray) {
--- a/tables/DataMan/VirtualTaQLColumn.cc
+++ b/tables/DataMan/VirtualTaQLColumn.cc
@@ -157,7 +157,7 @@ void VirtualTaQLColumn::prepare()
   TableColumn tabcol (table(), itsColumnName);
   itsExpr = tabcol.keywordSet().asString ("_VirtualTaQLEngine_CalcExpr");
   // Compile the expression.
-  TaQLResult res = tableCommand ("calc " + itsExpr + " from $1" , table());
+  TaQLResult res = tableCommand ("calc from $1 calc " + itsExpr, table());
   itsNode = new TableExprNode(res.node());
   // Check if the expression type matches the column type.
   if (itsNode->isScalar() == itsIsArray) {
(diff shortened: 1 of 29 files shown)


tables/DataMan/VirtualTaQLColumn.cc
```diff
@@ -157,7 +157,7 @@ void VirtualTaQLColumn::prepare()
   TableColumn tabcol (table(), itsColumnName);
   itsExpr = tabcol.keywordSet().asString ("_VirtualTaQLEngine_CalcExpr");
   // Compile the expression.
-  TaQLResult res = tableCommand ("calc from $1 calc " + itsExpr, table());
+  TaQLResult res = tableCommand ("calc " + itsExpr + " from $1" , table());
```
juliantaylor: can you explain how this relates to the column cloning?

gervandiepen: You are right. That change accidently became part of the change set. Very
sloppy of me. I'll remove it.
Thanks.

juliantaylor:
this merge caused a 30% performance regression in the CASA calibration tasks.
I assume this is due to the change from `asArray` to `toArray` in `TableRow.cc`. The extra array copy and involved memory allocations seem to be very expensive.
Can this be avoided?

gervandiepen:
I had asked Jim to confirm the change was fine before committing it.
Please feel free to change it to asArray again. I'm afraid I won't have
sufficient time in the coming weeks; I'm in Malta now for an SDP meeting
and will be 5 weeks on vacation thereafter.

juliantaylor:
can you say why it was changed?
I am not familiar with how this code works.

gervandiepen:
That change was made to make data type promotion possible for the clone
column functionality if the data type of the new column differs from the
old one (as requested by Jim).

It could be changed like below which I did for TpArrayBool only. Maybe you
make the change for the datatype which is used in your case (or maybe you
need them for more/all datatypes). I have to prepare something for tomorrow
now.

--- a/tables/Tables/TableRow.cc
+++ b/tables/Tables/TableRow.cc
@@ -623,11 +623,16 @@ void ROTableRow::putField (uInt rownr, const
TableRecord& record,
     case TpBool:
        (*(ScalarColumn<Bool>*)(itsColumns[whichColumn])).put
                                (rownr, record.asBool (whichField));
        break;
     case TpArrayBool:
+      if (record.dataType(whichField) == TpArrayBool) {
+       (*(ArrayColumn<Bool>*)(itsColumns[whichColumn])).put
+                               (rownr, record.asArrayBool (whichField));
+      } else {
        (*(ArrayColumn<Bool>*)(itsColumns[whichColumn])).put
                                (rownr, record.toArrayBool (whichField));
-       break;
+      }
+      break;
     case TpUChar:
        (*(ScalarColumn<uChar>*)(itsColumns[whichColumn])).put
                                (rownr, record.asuChar (whichField));

PR #430, opened 2016-06-01
CAS 8681 changes:

dmehring:
expose MSMetaData::getScanToStatesMap() in public interface
add private MSMetaData::_generateScanPropsIfWanted() to allow methods to generate _scanProperties
    data structure if it will be needed later.
Add filter() to filter scan keys based on an array key
Change tabs to spaces in some files.

### DIFF
--- a/ms/MSOper/MSKeys.cc
+++ b/ms/MSOper/MSKeys.cc
@@ -161,6 +161,24 @@ std::set<ArrayKey> uniqueArrayKeys(const std::set<ScanKey>& scanKeys) {
     return arrayKeys;
 }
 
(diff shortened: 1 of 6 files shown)


dmehring:
I had started another branch for this, so it was easiest to stick with that.

I looked at set_intersection() in place of filter(), but in set_intersection() when there is a match, both iterators are incremented, which means that the array key iterator is incremented, which is not always what should happen in this case. If there is another way to do that that I'm not understanding, we can talk about it, but the user needs this feature asap, so I'd like to get it merged if possible.

dmehring:
Thanks!

PR #431, opened 2016-06-01
Mutexfixes: large change set to improve thread-safety, lazy initialization, CMake config defaults

alxames:
On top of Ger van Diepen's work over the last year to make casacore thread-safe, fix some more thread-safety issues.
In particular, replace the double checked locking pattern with call_once/pthread_once calls that are thread-safe.

See the individual commits for change details and here and there hints for reviewing.

Ger agreed to review, but any other review (partial or full) is welcome.
In particular, can somebody build & test it on a system where __APPLE__ is defined?


alxames:
Opened discussion topic #432 related to this pull request in case there is a need to discuss broader issues than small parts of the pull request code.

tammojan:
The OSX build fails with:
```
[  1%] Building CXX object casa/CMakeFiles/casa_casa.dir/Arrays/IPosition2.cc.o
In file included from /Users/travis/build/casacore/casacore/casa/Arrays/IPosition2.cc:35:
In file included from /Users/travis/build/casacore/casacore/casacore/casa/Logging/LogIO.h:35:
In file included from /Users/travis/build/casacore/casacore/casacore/casa/Logging/LogSink.h:37:
/Users/travis/build/casacore/casacore/casacore/casa/OS/Mutex.h:234:15: error: expected expression
    : itsFlag(PTHREAD_ONCE_INIT)
              ^
/usr/include/pthread.h:201:27: note: expanded from macro 'PTHREAD_ONCE_INIT'
#define PTHREAD_ONCE_INIT {_PTHREAD_ONCE_SIG_init, {0}}
                          ^
1 error generated.
```

https://travis-ci.org/casacore/casacore/jobs/134475545#L646

juliantaylor:
linux clang also fails as its missing atomic_thread_fence, though that seems to be only used as a workaround (of which the why of it should be better documented in the code)

at first glance this looks quite ok. But the amount of duplication makes me wonder if we should consider dropping < c++11 at some point.

the PR does break some CASA tests, but its not obvious why as it doesn't really use threads.

alxames:
Julian,

Thanks for you swift response.

I had done a Clang (3.4) build + test on Linux, also with libc++, so I was surprised it ended up broken in Travis. Today, Tammo could reproduce it on his MacOS laptop.

Any idea why Clang on some systems is missing all C++11 atomics? It's a std lib issue, but on Linux Clang uses GNU's stdlibc++ by default. Strange thing is, I've also built with clang++ -stdlib=libc++ and that works fine on my Linux system. But the libc++ atomic header, also on my sys, has just about the whole file inside /* */... (?)

I did notice (your?) thread_local (TLS) use in Quanta/{Euler,MVPosition}, ifdef-ing it away on \_\_APPLE\_\_. I tried to enable the cache on more configs, e.g. USE_THREADS=OFF and on pre-C++11 using \_\_thread, which had varying levels of success.
I searched around, and it seems that on even recent (Xcode?) Clang versions, TLS is N/A, even for C++11 builds. Do you have any idea about this apparent non-conformance of TLS and atomics, how to reliably detect it, or how to not grab locks all the time without std::atomic?

In CMakeLists.txt I read:
>    ### after CASA switches from ASAP to libsakura, CASA will no longer
>    ### use casacore's python module and it will no longer use boost...
hence I tried to avoid using Boost. But is that still actual? Is boost sort of banned from casacore? I hope I can fix that pthread_once init issue without resorting to boost::call_once for pre-C++11 (which is nicer anyway).

> [...] consider dropping < c++11 at some point
Same feeling here. Atm, we do have some very old systems in use. They're on the way out, but that could still take months. (Casacore performance is irrelevant there though.) Essentially, we'd require GCC 4.8(.1) or Clang 3.1 (formerly 3.3).
I have the same feeling about USE_THREADS=OFF, but for the moment, we can deal.
I tried to keep code dup at a minimum. Please let me know if you see other ways to improve.

> the PR does break some CASA tests, but its not obvious why as it doesn't really use threads.
That was something I was interested in to hear. Could you provide the logs? For CASA regression test failures, it's going to be hard for me alone, but CASA build and casacore build and regression errors I may be able to fix it. I'm aware of one API breakage (noted in a commit log) that I probably want to undo anyway.

I will provide an updated (rebased) branch (but not in a few days) to:
* fix Clang/MacOS builds
* fix for CASA breakage (if cause or easy solution is on this side)
* improve commit logs (they were written much hastier than the code :)) and bisectability
* fix more thread-safe casacore statics. I noticed last night :(. I'm not sure if/when I can cover it all, esp. measures was really a lot of work.

In the meantime, any feedback is welcome. For the moment, we've marked this branch as 'dontmerge'.

alxames:
Saw it a few days ago. Wanted to say s'thing, but I didn't know exactly what :).
Maybe a combination of: thanks, hope the changes are all fine, if only I had any time left at the time to improve this PR..., surprising it merges cleanly after 2+ yrs, congrats with the opening of APERTIF and with DUPPLO, have fun at the upcoming open day, how do you do, and good luck. :)

PR #434, opened 2016-06-02
Cas 8713

dmehring:
* Implemented MSMetaData::_generateSubScanPropsIfWanted() as analog to _generateScanPropsIfWanted()  and modified implementation of _getScansAndSpwMaps() to use it.
* Implemented MSMetaData::getScanToSpwsMap() to return complete scan -> spws map so callers do not have to call MSMetaData::getSpwsForScan() in a loop which is inefficient when the number of scans is large.


dmehring:
Thanks!

PR #436, opened 2016-06-09
Cas 8731  	Implemented MSMetaData::getSpwToScansMap()

dmehring:
Also  	Place test for getScansForSpw() in its own block in tMSMetaData.

### DIFF
--- a/ms/MSOper/MSMetaData.cc
+++ b/ms/MSOper/MSMetaData.cc
@@ -1428,6 +1428,13 @@ std::map<ScanKey, std::set<uInt> > MSMetaData::getScanToSpwsMap() const {
     return scanToSpwMap;
 }
 
+vector<std::set<ScanKey> > MSMetaData::getSpwToScansMap() const {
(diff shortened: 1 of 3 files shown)


ms/MSOper/MSMetaData.cc
```diff
@@ -1428,6 +1428,13 @@ std::map<ScanKey, std::set<uInt> > MSMetaData::getScanToSpwsMap() const {
     return scanToSpwMap;
 }
 
+vector<std::set<ScanKey> > MSMetaData::getSpwToScansMap() const {
```
juliantaylor: spw's here are always indexed 0 to vector.size()?
I would have expected a map int -> set (or multi_map)

dmehring: Yes. Values of spw are simply the row numbers in the SPECTRAL_WINDOW subtable, so a vector seemed like the most natural choice.

juliantaylor: ok assuming this is consistent in the api
thanks

dmehring:
Thanks.

PR #437, opened 2016-06-09
Fixed valid interval calculation for pointing when interval is 0.0

kgolap:


ms/MeasurementSets/MSPointingColumns.cc
```diff
@@ -129,11 +129,11 @@ MDirection ROMSPointingColumns::sourceOffsetMeas(Int row,
     for (Int i=start; i<end; i++) {
       if (antennaId()(i)==antenna) {
 	Double halfInt=0.0;  
-	if(interval()(i)==0){
+	if(interval()(i)==0.0){
 	  Int counter=0;
 	  Int adder=1;
 	  
-	  while(time()(i+counter)==time()(i)){
+	  while(!( (time()(i+counter)!=time()(i))&& (antennaId()(i+counter) == antennaId()(i)))){
```
juliantaylor: `antennaId()(i)` can just be `antenna`

it will endless loop (or crash?) when the antenna id does not appear anymore (in addition to all times being equal), though I assume that does not happen?

kgolap: In theory you are right... in practice an antenna will appear more than 
once in a pointing table.
Fixed it for that...just in  case

Thanks
Kumar

PR #443, opened 2016-06-14
CAS-8741 bug fix for improper handling of single channel spectral window

dmehring:


juliantaylor:
I don't like this code, but the fix does look correct
thanks

dmehring:
Thanks. I doubt you dislike this code more than I do :) ..

PR #444, opened 2016-06-14
CAS-8743 remove WSRT specific code, since writing of WSRT MSes

dmehring:
is delegated completely to MSFitsOutputAstron


dpetry:
Should be reviewed by somebody from ASTRON.

gervandiepen:
I've taken a look at MSFitsOutput; it uses the MSFitsOutputAstron version
for WSRT and LOFAR. So the changes are fine.
It can be merged.
Ger

PR #447, opened 2016-06-15
Use omp_get_max_threads to determine fftw thread number

juliantaylor:
Allows controlling openmp based applications completely via the openmp
env variables.

### DIFF
--- a/scimath/Mathematics/FFTW.cc
+++ b/scimath/Mathematics/FFTW.cc
@@ -37,6 +37,9 @@
 #ifdef HAVE_FFTW3
 # include <fftw3.h>
 #endif
+#ifdef _OPENMP
+# include <omp.h>
+#endif
 
 #include <iostream>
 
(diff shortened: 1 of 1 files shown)


gervandiepen:
Wouldn't it be nicer to put this in a function maxNumThreads() in HostInfo?
Some people might argue that FFTW is not using OpenMP, so its nr of threads should not depend on OMP_NUM_THREADS. However, I think it is fine to use it as a general limit.

BTW. I noticed that HostInfo is not thread-safe when testing and filling its info member.

juliantaylor:
if casacore is compiled with openmp it makes sense that the fft respects its limits, fftw can also use openmp internally.

Currently there is no way to control at all how many threads the fft will use which is a problem when the parallelization occurs at a higher level. With this one can at least control it via one common method.

I think you can technically already chance the cpu number in HostInfo as a user but a function for it would also make sense.
Then numCPUs should respect openmp settings by default (if compiled in) get a `force` option to allow overriding it.
But then changing global state is also ugly, maybe here just adding a nthread parameter to the FFT class is nicer?

juliantaylor:
not the best currently but there must to be a way to control the threads of fft for casa so merging.

Issue #446, opened 2016-06-15
Linux HostInfo does not account for cgroups settings


juliantaylor:
the memory reported as available by it just reads /proc/meminfo and returns MemTotal
This does not account for per process constraints on the usable memory imposed by the standard linux resource management system cgroups.

The function should be adapted to read the real limits and reports these to allow proper resource usage on flexible clusters.

This involves:

* read `/proc/self/cgroup` to obtain the cgroup hierachy
* try cgget on that hiearchy to get the set memory and swap limits
* if unavailable read it from `/sys/fs/cgroup/<hiearchy>` directly

PR #450, opened 2016-06-16
consider cgroup limits for memory in HostInfo

juliantaylor:
You usually want to know the available memory that can be used when
using these functions but it returns all memory of a machine instead.
This causes problems when a process is resource limited via cgroups.

Update the function to consider the cgroup limits when returning its
results.
Currently only supports v1 cgroups, but currently there are no
standard distributions using v2 yet.

Currently hardcodes common cgroup mount point, technically one has to
search the location in the mount list.
It might be useful to change it to use libcgroup/cgget to be more
generic but the current layout works for ubuntu/debian/fedora/el7.

Closes gh-446

### DIFF
--- a/casa/OS/HostInfoLinux.h
+++ b/casa/OS/HostInfoLinux.h
@@ -49,6 +49,12 @@
 #include <fcntl.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <fstream>
+#include <iostream>
+#include <string>
+#include <sstream>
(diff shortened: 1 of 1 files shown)


juliantaylor:
an issue might be that the returned free memory is now limit - used in the hierarchy, if there is no limit it might be less than is actually used.

could be changed by only using cgroups to determine used memory when there is a limit set for the process

PR #451, opened 2016-06-17
CAS-8477 Ensure that receptor angles for specified spectral windows

dmehring:
do not vary on an antenna by antenna basis

implemented MSMetaData::getUniqueSpwIDs()

modify MSFitsOutput::writeFitsFile() not to consume thrown exception just to return False, but rather allow thrown exceptions to continue to propogate up the stack. Method return type remains Bool.


juliantaylor:
thanks, though I'd have prefer this split into three commits.

PR #455, opened 2016-06-28
CAS-8783 Parallel stats

dmehring:
* Refactor, create ClassicalStatistics::_increment() method for updating loop counters
* Implement StatisticsUtilities::combine() for combining statistics derived from different data sets to get a set of statistics that applies to all those data sets.
* implementation of ParallelClassicalStatistics for parallel support in _getStatistics(). There are other avenues to investigate, but this implementation does improve performance on computing stats on large images by a factor of 1.5 to 3 depending on the use case.
* add support for parallel computation of stats to LatticeStatistics (default behavior is unchanged; objects must be explicitly configured to use parallel computation).
* reformat files, changing tabs to spaces.


dmehring:
Hmm looks like the travis build failed. Maybe I forgot to include an omp file. Anyway, I'm currently investigating a hopefully better way to do this, so hold off reviewing this for now. I'll give a heads up when it's ready to review.

juliantaylor:
openmp is disabled in travis, so you have to put ifdefs around the function calls (replace result with 1) and includes, but the pragmas don't need them.

dmehring:
This can now be reviewed and merged. Apologies for the roundabout path to get to the final solution (which was complicated by the annoyance of a non-fast-forward push issue that I had to resolve with a pull, merge, and a couple of additional commits before I could push my second batch of changes).

juliantaylor:
looks quite good

on a simple ia.statistics() on a large image in page cache it stops scaling when the threads receive less that ~5000 elements to process, so instead of using `ciCount/100` to determine nthreads I think 5000 is a better value. It avoids wasting cycles in the barriers.

I would prefer when the factor 8 for padding would be named via a define or static, e.g. CACHE_PADDING or CACHELINE_OFFSET

dmehring:
Thanks, updated as per your suggestions.

lattices/LatticeMath/LatticeStatistics.h
```diff
@@ -199,23 +199,26 @@ template <class T> class LatticeStatistics : public LatticeStatsBase
...
+        FitToHalfStatisticsData::USE_DATA ud;
+        // fit to half center value
+        AccumType cv;
+        // Chauvenet zscore
+        Double zs;
+        // Chauvenet max iterations
+        Int mi;
+        // maximum number of threads to use for
+        // parallel support
+        uInt maxThreads;
```
juliantaylor: this is appears to be unused

scimath/Mathematics/StatisticsData.h
```diff
@@ -45,35 +45,37 @@ class String;
...
-		HINGESFENCES
-	};
+    // implemented algorithms
+    enum ALGORITHM {
+        CHAUVENETCRITERION,
+        CLASSICAL,
+        FITTOHALF,
+        HINGESFENCES,
+        // classical with parallel support
+        PCLASSICAL
```
juliantaylor: is this still needed?

juliantaylor:
there are now iterators per thread, they do not share any state right?
then their increments could be moved into the parallel loop, most of the setup code of `_getStatistics` can probably be put in there.

dmehring:
The two bits of code you noted are no longer needed. I'll remove them. The increment code cannot go in the parallel loop, because if one uses a data provider (which is used eg by LatticeStatstics), it is the object from which the iterators are formed. Basically, it provides a chunk of data, and the threads each use a piece of that chunk. Once they have all processed their portion of that chunk, the data provider is then incremented to provide the next chunk. So, the increment of the data provider has to occur outside the parallel loop.

juliantaylor:
I don't follow.
The nthread DataIterators are passed to the _computeStats function, inside there they are incremented in each thread via the StatisticsIncrementer
So when the ++ operator is not threadsafe for this object the parallel loop would not be threadsafe either

juliantaylor:
just to clarify I only mean the first `for i in nthreads` inside the while loop, something like this:

```patch
--- a/scimath/Mathematics/ClassicalStatistics.tcc
+++ b/scimath/Mathematics/ClassicalStatistics.tcc
@@ -481,13 +481,22 @@ StatsData<AccumType> ClassicalStatistics<CASA_STATP>::_getStatistics() {
         uInt extra = _myCount % nthreads;
         uInt ciCount = _myCount/nthreads;
         vector<uInt> initialOffset(nthreads);
+       uInt dataStride = _myStride*nthreads;
+	 uInt maskStride = _maskStride*nthreads;
+       if (_hasWeights) {
+		stats.weighted = True;
+	 }
+	 if (_hasMask) {
+	 	stats.weighted = True;
+	 }
+
+#pragma omp parallel for
         for (uInt i=0; i<nthreads; ++i) {
             uInt idx8 = CACHE_PADDING*i;
             ngood[idx8] = 0;
             dataCount[i] = ciCount;
-            if (extra > 0) {
+            if (i % nthreads < extra) {
                 ++dataCount[i];
-                --extra;
             }
             initialOffset[i] = i*_myStride;
             dataIter[idx8] = _myData;
@@ -507,17 +516,7 @@ StatsData<AccumType> ClassicalStatistics<CASA_STATP>::_getStatistics() {
                     ++maskIter[idx8];
                 }
             }
-        }
-        uInt dataStride = _myStride*nthreads;
-        uInt maskStride = _maskStride*nthreads;
-        if (_hasWeights) {
-            stats.weighted = True;
-        }
-        if (_hasMask) {
-            stats.weighted = True;
-        }
-#pragma omp parallel for
-        for (uInt i=0; i<nthreads; ++i) {
+
             uInt idx8 = CACHE_PADDING*i;
             _computeStats(
                 tStats[idx8], ngood[idx8], mymin[idx8], mymax[idx8],
```

dmehring:
Ah ok, I misunderstood what you meant. Yes, working on that bit now.

juliantaylor:
actually the way you are iterating is not ideal for parallel processing, if I understand it correctly say for stride == 1 and 4 threads, the first thread reads the points 1, 5, 9, ... and thread 2 reads 2,6,10, ... etc.. Is this correct?

this is not good as this will imply that each cpu will load some memory another cpu has already loaded as they load memory in blocks of 64 bytes (a cache line). Also data access per cpu will never be stride == 1 which means it cannot be vectorized.

It would be better when thread 1 reads a whole block of stride 1 data (if the data layout allows that) and then jumps to the next block of contiguous data.
A problem with this approach the lack of the += operator which means that the threads need to spend quite some time just incrementing the iterator without doing useful work.

dmehring:
Yes, that seemed like it would be the better way to me as well, so I did try it. But at least for my test case, it resulted in repeatably poorer performance (by about 10-20% as I recall).

juliantaylor:
the lack of += also affects the current iteration schema, as to get data 1,5,9, .. it has to increment 4 times in each step, so maybe it would be better to change the method to improve memory load efficiency and allow future vectorization

for this instead of having a loop from 1..nthread, have a loop 1...nrdatablocks but still only accumulate into the nthread stat objects. You select the stat object to use in the loop via `omp_get_thread_num` which will gie you a number 0..nthread-1

juliantaylor:
do you still have the code? did you still have only a 1..nthread loop in parallel or did you use a loop over a larger number of smaller datablocks?
in the former case you would have a asymmetric workload as some threads need to iterate more than others so the parallel efficiency is lower.

dmehring:
How would you suggest choosing a reasonable value for nrdatablocks?

I don't have the code, but it should be relatively simple for me to recreate and commit if you want to see it. It still used 1..nthreads, it just incremented the start iterators of each thread to a different location and used the default data stride (rather than nthreads*datastride for the stride as it currently does), so that rather than "interleaving" how the threads read, each block from which a thread read was isolated from the others.

dmehring:
This commit still uses interleaving of iterators, but resolves the obsolete and thread initialization issues we discussed earlier.

juliantaylor:
Usually you would want to give each thread data which fits into the higher level cpu caches, so L2 or L3 sized datablocks which can be from 500kb to several mb, to squeeze out max performance you also need to be aware of which cpus might share caches e.g. L3 is often shared over a whole socket.

But here we are just iterating once over the memory once so the caches don't matter. static scheduled openmp for loops which you want to use for this symmetric problem have almost no block specific overhead, in the end each thread just gets a fixed starting point at the beginning and they start working without any additional synchronization until they are done.
So the blocksize does not matter much and can be pretty small. I'd say 1000 elements is a good starting point, but we will want to measure how it scales. The current code stopped scaling at 5000 per thread so a too small blocksize might start too many threads. But with a better memory access scaling behaviour could be better.

dmehring:
The version I just committed does not use interleaved thread reads and is about 40% slower than the interleaving version on my test case.

juliantaylor:
thanks, in this version for say 4 threads, the first thread starts working on its quarter immediately while the 4 thread will first have to iterate over three quarter of the entries to even start working. So the first few threads will have a significant idle time waiting for the last one to finish.
By using smaller blocks you will keep all threads busy the whole time like in the interleaved case which should then have similar or better performance.

dmehring:
OK, yes I can understand the delay the incrementing would incur in the current non-interleaved case. Essentially, (one of) the last threads is likely the bottleneck since its iterator must increment most of the way to end of the block before it can begin its computational work. I'll try to implement a version that uses smaller blocks.

juliantaylor:
note that for the simple case where the iterator is just a pointer you could also get this fast by using `std::advance` instead of the loop.
It even seems the the compiler is able to do that on its own for the StatisticsIncrementer where the loop is simpler.

dmehring:
Multiple blocks now processed per thread. This really didn't improve performace that much (maybe 5-10%). The optimal block size for my image stats test case was 2000 (+/-500), without much variance between 1,000 and 3,000. I was able to remove a lot of inputs to associated methods though, which is a good thing.

juliantaylor:
instead of using a critical section you can probably just increase the per thread iterators by `omp_get_num_threads() * blocksize`

I didn't expect it to increase performance much, but this way of iterating keeps the option of vectorization open which an be significant.

dmehring:
Redoing the iterator advancement without the critical block has improved performance by about 15%.

One concern I have is the number of times a specific thread ID will be assigned in the nBlocks loop. For example, say nBlocks = 33 and there are 16 threads. Will thread ID 0 always be assigned for three loop iterations and all the other thread IDs assigned to two iterations each? The code assumes implicitly that is true in how it handles advancing of the iterators, and testing seems to bear it out (at least I'm not seeing test failures after running tests repeatedly).

juliantaylor:
the standard guarantees that the chunks of a parallel for with static scheduling are handed out in a round robin fashion, to me this means each thread gets a block when there are as many or more blocks than threads.
It is not very specific on how large a chunksize is when it is not specified, so here in theory an implementation could choose a chunksize causing less chunks than threads. To avoid this hypothetical issue you could specify schedule(static, 1) to set the chunksize to 1 explicitly (without it you will usually get nBlocks / nthreads chunks)

note you could also write it like this:

```
#pragma omp parallel
{
// stagger initial local iter
localdataiter = maindataiter +  omp_get_thread_num()  * blocksize * omp_get_num_threads()
#pragma omp for
for i in blocks {
   handle_datiter(localdataiter)
   localdataiter += omp_get_num_threads() * blocksize
}
}
```

the regular parallel section is executed by all threads setting up their local variables and the go into the for loop where each thread handles its chunk
but this is effectively the same thing if std::advance is constant complexcity so there is no need to change the code.

I think this looks mergeable now, do you still have more changes in mind?

dmehring:
I don't have any additional changes in mind at the moment. I'll be watching the daily test reports for unexpected failures. A lot of tests depend on this code, so I think it should be pretty obvious if anything is amiss. Thanks for your help.

Issue #453, opened 2016-06-23
excessive cache flushing with Lattices


juliantaylor:
In casa we have an issue that we want to update a Lattice in an order which does match too well with the tileshape.
For this purpose the backing PaggedArray has the bucketcache which is supposed to collect partial tile writes and write them out when they are fuller.
Unfortunately the Lattice infrastructure seems to flush the caches everywhere making the cache moot.

E.g. consider this code:

```c++
#include <casacore/lattices/Lattices/SubLattice.h>
#include <casacore/lattices/Lattices/PagedArray.h>
#include <casacore/casa/Arrays/AxesSpecifier.h>
#include <casacore/casa/Arrays/Vector.h>
#include <casacore/casa/Arrays/IPosition.h>
#include <casacore/casa/Inputs/Input.h>

#include <casacore/casa/namespace.h>

int main (int argc, const char* argv[])
{
    int nx = 400, ny = 400;
    PagedArray<Float> pa(IPosition(3,nx, ny, 100), "tLattice_tmp.pa");
    PagedArray<Float> pb(IPosition(3, nx, ny, 1), "input.pa");
    pa.set(1.5);
    pb.set(2.5);
    for (int i = 0; i < 100; i++) {
        Slicer slicer(IPosition(3, 0, 0, i), IPosition(3, nx, ny, 1));
        SubLattice<Float> sl(pa, slicer, True);
        int r = sl.advisedMaxPixels();
        cout << sl.shape() << " " << pb.shape() << " " << sl.advisedMaxPixels() << " " << sl.niceCursorShape() << endl;
        sl.copyData(pb);
    }
    return 0;
}
```

here it correcty chooses 64 buckets as cachesize so it should just writes 62mb of data one full tile at a time. But copyData creates a dummy temporary LatticeIterator for some reason. The destruction of this iterator triggers a cache flush and it actually writes out 10 times more data than necessary (the tileshape is 50,50,10).
This goes against the documentation saying iterators should be short lived, when the flush the cache on destruction the must be long lived.

Similarly the LatticeExpr also flushes the cache on destruction which is a major problem for these types of iterations that do not match the tile ordering.

Is there something I am missing or is the cache just actively harmful in combination with Lattices?

juliantaylor:
Assuming I made no mistakes, how to best fix this?

My first idea would be to add a `setFlushOnDelete` function and member to the `LatticeBase` and have the `PagedArray` check this to decide whether it should flush.
[...]

PR #457, opened 2016-07-06
Remove table flush on destruction of PagedArray

juliantaylor:
The PagedArray destructor flushes its underlying Table which essentially
makes the write Tile cache useless as many short lived objects do
contain PagedArray objects. When iterating over a Lattice in an
order that does not match the tileshape it will flush the cache on
destructions of for example LatticeIterators, LatticeExpressions or
SubImages and it while thus constantly rewrite partially modified tiles.
In some CASA tasks this leads to unnecessary very large write
amplifications.

To remove the all code that explicitly looks at table files on disk must
be adapted to respect the TableCache as it the flush frequency is
massively reduced and the files might not exist on disk at the time the
functions try to read them.
This commit updates Table::isReadable and TableInfo which is enough to
make most of the CASA regression suite succeed.
This also requires removing an unnecessary lock from
TableCache::lookCache which is during the reopenRW call may otherwise
recursively lock the TableCache.

IsWritable is not changed as it does not appear necessary currently and
if would need additional logic to handle read-only tables in cache that
are supposed to be opened in write mode.

closes gh-453

### DIFF
--- a/tables/Tables/TableCache.cc
+++ b/tables/Tables/TableCache.cc
@@ -170,9 +170,8 @@ void TableCache::flushTable (const String& name,
 PlainTable* TableCache::lookCache (const String& name, int tableOption,
                                    const TableLock& lockOptions)
 {
-    ScopedMutexLock sc(itsMutex);
     //# Exit if table is not in cache yet.
-    PlainTable* btp = getTable(name);
+    PlainTable* btp = this->operator()(name);
     if (btp == 0) {
 	return btp;
     }
--- a/lattices/Lattices/PagedArray.tcc
+++ b/lattices/Lattices/PagedArray.tcc
@@ -182,15 +182,6 @@ PagedArray<T>::~PagedArray()
   if (itsMarkDelete) {
     tempReopen();
   }
-  // Only need to do something if really constructed.
-  if (! itsTable.isNull()) {
-    // Table may not be written if ref count > 1 - here we force a write.
-    // (but only if it is not a scratch table).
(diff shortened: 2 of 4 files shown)


juliantaylor:
@tammojan can you please test this branch, it is a large behaviour change and is quite likely to break stuff.

tables/Tables/TableCache.cc
```diff
@@ -170,9 +170,8 @@ void TableCache::flushTable (const String& name,
 PlainTable* TableCache::lookCache (const String& name, int tableOption,
                                    const TableLock& lockOptions)
 {
-    ScopedMutexLock sc(itsMutex);
```
tammojan: Why did you delete the mutexes?

juliantaylor: the mutex is not necessary for the full function, we only need it to access the shared state, the `tableMap_p` which is done in the operator call following it.
if we keep this mutex we would need recursive locking as the reopenRW will call isReadable which calls back into the cache.

juliantaylor: or at least it appears to be the only state, you would assume that the synchronization of the calls into the shared table object would be handled by the lock of he table.
But it doesn't seem to have a lock so maybe we need to make it recursive ..

juliantaylor:
Only part of the casa test suite is capable to run from a git branch, so to test the rest a merge is simplest way.
Should problems show up please let me know, I will follow up with fixes or a revert if necessary. I am also open to changing the locking scheme to recursive as it might be a practicality over purity case.

PR #458, opened 2016-07-11
remove unnecessary String constructions

juliantaylor:
the conversion from const char to String implies unnecessary memory
allocation in some relatively commonly called functions. Avoid this via
a const char interface


juliantaylor:
updated to use a const char interface instead of static const strings, with the former you might end up with lots of static string objects from the template in non c++11 mode. Not sure if the compiler can deduplicate these objects.

PR #461, opened 2016-07-14
Import FITS-IDI WEATHER tables

kettenis:
Import FITS-IDI WEATHER tables and store the information from those tables in an MS WEATHER table. Tested using a VLBA data set that includes a WEATHER table.  Also tested using EVN data that doesn't have weather information.


dpetry:
Need to think about extending the unit tests for this.

PR #464, opened 2016-07-19
Measures data: reduce update frequency for IERS tables

tammojan:
The update frequency for IERS EOP-files was reduced in December 1 2011, see http://hpiers.obspm.fr/eop-pc/index.php?index=C04 . This has cause the script that generates the measures data tables to go into an infinite loop, downloading eopc04.16 , causing unnecessary load on the hpiers.obspm.fr server. 
The attached fix was implemented at ASTRON in december 2011 but apparently didn't make it upstream...


tammojan:
Confirmed with Wim Brouw and @apschoenmakers .

PR #467, opened 2016-07-19
Justoagv aipsrc

justoagv:
Prologue: I have not changed any existing functionality / interface, just added some new features.

Following requirements from the CASA HPC project (parallelization framework) I've added some set/gets to the casacore HostInfo class to override the aipsrc file values. In this way we can set different values per process for the memory/memoryFraction/numCores settings. 

When the custom memory/memoryFraction/numCores values are <= 0 they are ignored, so it is possible to set them to some custom value and then go back to the .aipsrc file defaults by setting them to 0.

See CAS-8159 and CAS-8158.

### DIFF
--- a/casa/OS/HostInfo.cc
+++ b/casa/OS/HostInfo.cc
@@ -124,7 +124,9 @@ Int HostInfo::numCPUs(bool use_aipsrc)				\
     static const String keyword("system.resources.cores");	\
     if ( use_aipsrc ) {						\
 	String value;						\
-	if ( Aipsrc::find(value, keyword) ) {			\
+	if (resources_numCPUs > 0) {				\
+		return resources_numCPUs;			\
+	} else if ( Aipsrc::find(value, keyword) ) {			\
 	    int result;						\
 	    if ( sscanf( value.c_str( ), "%d", &result ) == 1 )	\
 		return (Int) result;				\
@@ -144,11 +146,15 @@ ptrdiff_t HostInfo::memoryTotal(bool use_aipsrc) 		\
     /** returns the memory in kilobytes...         **/		\
     if ( use_aipsrc ) {						\
 	String value;						\
-	if ( Aipsrc::find(value, memory) ) {			\
+    if (resources_memory > 0) {				\
+        return resources_memory;				\
+    } else if ( Aipsrc::find(value, memory) ) {		\
 	    int result;						\
 	    if ( sscanf( value.c_str( ), "%d", &result ) == 1 )	\
 		return (ptrdiff_t) result * 1024;		\
-	} else if ( Aipsrc::find(value,	fraction) ) {		\
+    } else if (resources_memfrac > 0) {			\
+        frac = resources_memfrac;				\
+    } else if ( Aipsrc::find(value,	fraction) ) {		\
 	    int result;						\
 	    if ( sscanf( value.c_str( ), "%d", &result ) == 1 )	\
 		frac = result;					\
@@ -200,6 +206,26 @@ ptrdiff_t HostInfo::swapFree( )					\
     if ( ! info ) info = new HostMachineInfo( );		\
     info->update_info( );					\
     return info->valid ? info->swap_free : -1;			\
+}								\
+								\
+ptrdiff_t HostInfo::setMemoryTotal(ptrdiff_t memory)		\
+{								\
+    ptrdiff_t old_memory = resources_memory;			\
+    resources_memory = 1024*memory;				\
+    return old_memory;						\
+}								\
+								\
+Int HostInfo::setMemoryFraction(Int memfrac)			\
+{								\
+    Int old_memfrac = resources_memfrac;			\
+    resources_memfrac = memfrac;				\
+    return old_memfrac;					\
+}								\
+Int HostInfo::setNumCPUs(Int numCPUs)				\
+{								\
+    Int old_numCPUs = resources_numCPUs;			\
+    resources_numCPUs = numCPUs;				\
+    return old_numCPUs;					\
 }
 
 
@@ -271,5 +297,8 @@ ptrdiff_t HostInfo::swapFree( )    { return -1; }
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
 HostMachineInfo *HostInfo::info = 0;
+ptrdiff_t HostInfo::resources_memory = 0;
+Int HostInfo::resources_memfrac = 0;
+Int HostInfo::resources_numCPUs = 0;
 
 } //# NAMESPACE CASACORE - END
--- a/casa/OS/HostInfo.h
+++ b/casa/OS/HostInfo.h
@@ -131,12 +131,26 @@ public:
     static ptrdiff_t swapFree();
     // </group>
 
+    // Allows to set custom resource values overriding the
+    // existing .apisrc setting and/or system defaults
+    // Expected unit are KiB for memory total
+    // and percentage (%) for memory fraction
+    // Returns the value previously stored
+    // <group>
+    static ptrdiff_t setMemoryTotal(ptrdiff_t memory);
+    static Int setMemoryFraction(Int memfrac);
+    static Int setNumCPUs(Int numCPUs);
+    // </group>
+
 private:
     // we don't want folks creating these...
     HostInfo( );
     const HostInfo &operator=( const HostInfo & );
 
     static HostMachineInfo *info;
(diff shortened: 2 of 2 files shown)


casa/OS/HostInfo.cc
```diff
@@ -124,7 +124,9 @@ Int HostInfo::numCPUs(bool use_aipsrc)				\
     static const String keyword("system.resources.cores");	\
     if ( use_aipsrc ) {						\
 	String value;						\
-	if ( Aipsrc::find(value, keyword) ) {			\
+	if (resources_cores > 0) {			\
+		return resources_cores;			\
```
juliantaylor: indentation is wrong

casa/OS/HostInfo.cc
```diff
@@ -271,5 +291,8 @@ ptrdiff_t HostInfo::swapFree( )    { return -1; }
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
 HostMachineInfo *HostInfo::info = 0;
+int HostInfo::resources_memory = 0;
+int HostInfo::resources_memfrac = 0;
+int HostInfo::resources_cores = 0;
```
juliantaylor: memory needs to be a size_t or ptrdiff_t

casa/OS/HostInfo.h
```diff
@@ -131,12 +131,22 @@ class HostInfo
     static ptrdiff_t swapFree();
     // </group>
 
+    // Introduce run-time memory limit re-specification capabilities
+    // <group>
+    static void setMemoryTotal(int memory);
+    static void setMemoryFraction(int memfrac);
+    static void setCores(int cores);
```
juliantaylor: the function to get is is named CPU this should use the same name

casa/OS/HostInfo.h
```diff
@@ -131,12 +131,22 @@ class HostInfo
     static ptrdiff_t swapFree();
     // </group>
 
+    // Introduce run-time memory limit re-specification capabilities
```
juliantaylor: missing documentation that this is only used when aipsrc is used too

also why is that so?

casa/OS/HostInfo.h
```diff
@@ -131,12 +131,22 @@ class HostInfo
...
+    // <group>
+    static void setMemoryTotal(int memory);
```
juliantaylor: its probably useful to return the old value in the setter to simplify reverting the setting again

casa/OS/HostInfo.cc
justoagv: indentation seemed a bit chaotic, is there a standard for that?

casa/OS/HostInfo.cc
justoagv: When the value is read from .aipsrc is comes as int, and then is casted to ptrdiff_t so I keep that pattern

casa/OS/HostInfo.h
justoagv: I followed the naming convention of .aipsrc for the setters

casa/OS/HostInfo.h
justoagv: I think is not needed, because the original values are re-read from .aipsrc for each call (which by the way does not seem very efficient but I didn't change it)

casa/OS/HostInfo.cc
justoagv: this is a test

casa/OS/HostInfo.h
justoagv: My reasoning was that HostInfo calls with use_aiprs=False really want to rely on defaults, and only the ones with use_aipsrc=True really want something customized.

casa/OS/HostInfo.cc
juliantaylor: 4 spaces should be the standard but it is very inconsistent, so either reindent the file correctly or use same indentation as the rest of the file

casa/OS/HostInfo.cc
juliantaylor: the reading from aipsrc is wrong, rather fix that than repeating the error

casa/OS/HostInfo.h
juliantaylor: true this is inconsistent, but consistency of the class is more important imo

casa/OS/HostInfo.h
juliantaylor: if you have changed them they are not reread from file
currently it is not possible to sanely restore the defaults as the aipsrc gets in the way of that, so this is almost mandatory

justoagv: But if the custom values are set to zero, then .aipsrc is read again, is not that enough?

juliantaylor: technically yes but you cannot set them back to their original zero value as you will set them to the aipsrc value instead. this is probably fine in practice as the aipsrc should not change during the execution. But returning the old value in the setter is still more convenient

casa/OS/HostInfo.cc
justoagv: and what about the slashes?

casa/OS/HostInfo.h
justoagv: and what about the input arguments, should they be ptrdiff_t  too? Isn't that inconvenient?

casa/OS/HostInfo.cc
juliantaylor: I'm not a big fan of these aligned slashes but if the original author went through the trouble to do so one might as well keep them aligned

casa/OS/HostInfo.h
juliantaylor: of course the input argument needs to be ptrdiff_t for the memory, how is that inconvenient?

justoagv: what type should be use to interface with python?

juliantaylor: Py_ssize_t is the equivalent in the python c-api

justoagv: can one use Py_ssize_t in the swig xml files?

juliantaylor: swig so far I know does not use xml files, so I guess you mean casas homebrew'd gcwrap. There probably no, but can you even bind casacore functions with that?

juliantaylor: also casa only supports 64 bit linux, so in bindings you can also just use long or longlong which is the same on that platform

casa/OS/HostInfo.cc
```diff
@@ -144,11 +146,15 @@ ptrdiff_t HostInfo::memoryTotal(bool use_aipsrc) 		\
     /** returns the memory in kilobytes...         **/		\
     if ( use_aipsrc ) {						\
 	String value;						\
-	if ( Aipsrc::find(value, memory) ) {			\
+    if (resources_memory > 0) {				\
+        return resources_memory * 1024;			\
```
juliantaylor: the conversion to kb should be done in the setter and not here, but more importantly it is inconsistent again. The getters return kb, but the setter wants mb.
yes the aipsrc is different but as before consistency of the class is more important.

(btw. it might be a good idea to add units to the aipsrc to make this unambiguous)

casa/OS/HostInfo.cc
```diff
@@ -200,6 +206,26 @@ ptrdiff_t HostInfo::swapFree( )					\
...
+}								\
+								\
+ptrdiff_t HostInfo::setMemoryTotal(ptrdiff_t memory)		\
+{								\
+    ptrdiff_t old_memory = resources_memory;			\
+    resources_memory = memory;					\
+    return old_memory;						\
+}								\
+								\
+ptrdiff_t HostInfo::setMemoryFraction(Int memfrac)		\
```
juliantaylor: the input and return type are inconsistent (int <-> ptrdiff_t)

casa/OS/HostInfo.h
```diff
@@ -131,12 +131,24 @@ class HostInfo
     static ptrdiff_t swapFree();
     // </group>
 
+    // Allows to set custom resource values overriding the
+    // existing .apisrc setting and/or system defaults
+    // Returns the value previously stored
+    // <group>
+    static ptrdiff_t setMemoryTotal(ptrdiff_t memory);
```
juliantaylor: missing documentation on the expected units which should be KiB for memtotal and % for memfrac

casa/OS/HostInfo.cc
justoagv: yes, good that u noticed

PR #468, opened 2016-07-20
Cas 8825 implement multi-threading for ClassicalStatistics::_binCounts()

dmehring:


scimath/Mathematics/ClassicalStatistics.tcc
```diff
@@ -940,78 +917,168 @@ vector<vector<uInt64> > ClassicalStatistics<CASA_STATP>::_binCounts(
...
+        }
+    }
+    vector<vector<uInt64> >::iterator begin = bins.begin();
+    vector<vector<uInt64> >::iterator iter = begin;
+    vector<vector<uInt64> >::iterator end = bins.end();
+    typename vector<CountedPtr<AccumType> >::iterator sbegin = sameVal.begin();
+    typename vector<CountedPtr<AccumType> >::iterator siter = sbegin;
+    typename vector<CountedPtr<AccumType> >::iterator send = sameVal.end();
+    vector<Bool>::iterator abegin = allSame.begin();
+    vector<Bool>::iterator aiter = abegin;
```
juliantaylor: these iterators do not need to be in the function global scope, vector.begin just maps to an inline &data[0] and this is not a performance relevant loop
also a short comment at the top on what you are doing helps readability. Its probably merging the thread results into the final results but I should not be required to read 40 lines of code to figure that out.

juliantaylor:
while it is consistent with the existing design of the class and can be kept as is, I do not like the addition of the loop iteration and blocking state to the parent object state very much.
The objects necessary to the computation loops should be kept local to the function performing the computation instead of having the required objects being part of the parent object and initialized.

It is usually better design to keep side effects of function calls minimal as the effects of functions more predictable. In particular when considering parallization, the less state and side effects your code has the easier and also often the more efficient the parallelization is.

The exception to this is when the side effect is very expensive and should not be repeated on each function call, but this does not appear to be the case here.

dmehring:
Assuming you are talking about the code I've added to _initLoopVars() and _incrementThreadIters(), I can modify so that all this code is local, but that's a lot of repeated code, especially considering I'm planning on implementing multi-threading in most of the methods that use this pattern, and there are around five more of those. So that's a lot of repeated code. I would argue on the basis of maintainability this common code should be factored as I've done here.

juliantaylor:
you do not need to repeat code, `initLoopVars` can instead of modifying the state of the object, read the state of the object and create the necessary new objects for the iteration. Just moving from object local storage to function local storage.
But as said the existing object already worked like that and its not really necessary to change that now so it is fine.

dmehring:
OK, I'll ponder that but leave as is for now. I understand your point that these variables need not be in object scope.

dmehring:
I've removed the per-thread iterators etc from object scope on another branch which I'll issue a PR for once the current branch has been merged.

dmehring:
Thanks!

PR #469, opened 2016-07-20
turn quicksort into introsort

juliantaylor:
limit worst case runtime to O(NlogN) by aborting to heapsort when the
recursion level goes above the expected depth of log2(len(data))

This avoids O(N^2) performance worst case and stack overflow on the not
so uncommon case where the median of 3 pivot selects the last element.
E.g. [1, 2, 3, 4, 5, 0] which can happen on indirect sorting of MS index
columns like SCAN_NUMBER.

### DIFF
--- a/casa/Utilities/GenSort.h
+++ b/casa/Utilities/GenSort.h
@@ -171,7 +171,7 @@ private:
                      uInt nparts);
 
     // Quicksort in ascending order.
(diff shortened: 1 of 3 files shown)


juliantaylor:
the same thing should also be applied to the compound key quicksorts in Sort.cc, but I'll probably not get to that soon.

PR #475, opened 2016-07-29
Cas 8855 implement multithreading in min/max computation

dmehring:


scimath/Mathematics/ClassicalStatistics.tcc
```diff
@@ -2352,10 +2403,10 @@ void ClassicalStatistics<CASA_STATP>::_makeBins(
 #define _minMaxCode \
     if (! mymin.null()) { \
         if (*datum < *mymin) { \
-            mymin = new AccumType(*datum); \
+            *mymin = *datum; \
         } \
         else if (*datum > *mymax) { \
```
juliantaylor: one should remove the else here as that prevents using branchless instructions and vectorization.
Its the classic example of the difference in what you think is the most efficient code but is not how a x86 cpu actually works.
x86 has min and max instructions that do not need a branch, by placing a else if here instead of unconditionally assigning the maximum you force the compiler to put a branch there as in the (possible) presence of threads it cannot remove conditional assignments.

but then these are shared pointers so I don't actually know what the compiler will do with it :/

juliantaylor: hm being shared pointers breaks the compilers ability to use the branchless instructions and vectorize
why are they shared pointers?
If they have to be a double and float specialization is probably worthwhile
also pull out the null check can be pulled out of the loops and replaced with first entry assignment.

dmehring:
The instruction on line 2408 is "else if" not "else". If *datum is between *mymax and *mymin, neither block is executed, that is the implicit "else" block in this conditional.

They are shared pointers because I use the null state to indicate that they haven't been set. They are not required to be set (eg when all data are flagged) and there needs to be something to indicate that. However, if shared pointers pose an issue, I can use another data structure, such as a std::pair<AccumType,Bool>, with the Bool flag indicating if they have been set or not.

juliantaylor:
yes and that else if is bad, you are assuming that executing less code is results in faster runtime. This is not the case here. Always executing the code is going to be faster because you can avoid a branch which can detrimental on performance as it might imply cpu pipeline invalidation (depending on the predictability of your branch) and more significantly prevents the compiler from auto-vectorizing the code.

I see why you want the null check, but it can still be pulled out of the loop as the code is written now.
When the accumulator is null you are setting it to the value currently encountered, you can just set the accumulator to the first entry in the data before running the loop.

dmehring:
OK, so then you are suggesting to replace with eg

`mymin = min(datum, mymin);
mymax = max(datum, mymax);`

yes? rather than having conditionals?

juliantaylor:
this has the same effect as removing the else, so yes

juliantaylor:
wouldn't it be much simpler to just add a flag to the _minMax functions which is set when a value has been set instead of adding all these special find first functions?

```
_minMax(bool & beenSet)
{
  while (c < nr) {
    if (*mask) {
      _minMaxCode
      beenSet=True;
    }
  }
}
```

dmehring:
I'm on vacation for two weeks so will have to look in detail when I return. But perhaps I misunderstood your point before. I thought you had wanted to factor conditionals out of _minMaxCode. With your suggestion I would still need to check the value of beenSet in _minMaxCode, so for each iteration. If that's ok to do from a performance stand point, then yes that's easier and requires less code.

juliantaylor:
you only have the conditional in the masked code, the non-masked code does not care about the minmax being set as its always the case

dmehring:
It is not the case when there are weights, it is not the case when there are ranges (which includes subclasses). In most cases, there is no gaurantee the first point is the first "good" point. Only when there are no weights, no mask, and no ranges is it the case.

juliantaylor:
the same `if ... else ...` `_minMaxCode` is used for all of these cases.
All of these cases don't care whether the first point is good or not it assigned regardless after the mask weight or range check

juliantaylor:
to clarify, I do not mean remove the mask, weight or range checks, I mean remove the null check with an boolean flag
you can probably even do that directly in the _minMaxCode macro, the compiler should be able to move the invariant of of the loop by itself in the simple case. But if not you can use a compile time constant argument.

```
#define minMaxCode(set)
 if (set)
  boolhasbeenset = True
...


unmasked_min_max()
{
for x in data:
 _minMaxCode(0);
}

masked_minmax()
{
for x in data:
if (!masked)
  _minMaxCode(1);
}
}
```

juliantaylor:
but this is a minor optimization,we can merge now without the findfirst commit and do the optimization it later

dmehring:
Ah ok I think I understand what you mean now. Thanks for the clarification. It's fine by me if you want to merge without the last commit.

juliantaylor:
ok I'll merge it skipping the last commit

just here the complete thing I actually meant, instead of some vague tidbits ._.
```C
_minMax(CountedPtr mymin)
{
    AccumType lmin;
    bool valueFound = False
   while (!valueFound && count < nr) {
       if (*mask) {
           lmin = *datum;
           valueFound = True
       }
       increment();
   }
   while (count < nr) {
       if (*mask) {
           lmin = min(lmin, *datum)
       }
       increment();
    }
    if (valueFound) {
        *mymin = lmin;
    }
}
```
It implies duplicating a couple small loops but all other logic stays intact.

juliantaylor:
merged without the last commit

PR #478, opened 2016-08-02
update travis boost-python bottle and tune cache size

juliantaylor:


juliantaylor:
annoying building the boost pushes the build above the timelimit so we can't fill the cache :/ probably needs bootstraping via a force successful build.
But better does someone have the link to a boost 1.61 binary bottle like for the 1.60 which this branch replaced? the 1.60 won't install anymore due to being outdated.

juliantaylor:
@gijzelaerr I think you created the py3 binaries we used, can you update them?
else we could just disable py3 on mac

gijzelaerr:
yeah, that is reason why we do a manual build, homebrew doesn't build python3 bindings out of the box which was causing timeouts. I'll update the boost library.

gijzelaerr:
uploaded new bottle: https://bintray.com/artifact/download/casacore/homebrew-bottles/boost-python-1.61.0.el_capitan.bottle.1.tar.gz

please try with this one.

Would be great to automate this somehow, for now wrote notes down here

https://github.com/casacore/casacore/wiki/How-to-create-a-python-boost-bottle-with-python3-enabled

juliantaylor:
thanks it now just about makes it in time

PR #479, opened 2016-08-04
Use IGRF12 instead of IGRF10 in measuresdata

olebole:
IGRF12 has data for ten more years (until 2015 instead of 2005) and is the [latest IGRF model](http://www.ngdc.noaa.gov/IAGA/vmod/igrf.html).

Since the model is unchanged (for a change, a new model is created), it does not make sense to update the table from the same data again, so it makes sense to disable the update here.

### DIFF
--- a/measures/apps/measuresdata/measuresdata.cc
+++ b/measures/apps/measuresdata/measuresdata.cc
@@ -244,8 +244,8 @@ const String &IERSpredict2000Format = IERSpredictFormat;
 
 // IGRF
 const String IGRFFormat = String("") +
-"a1 x2 i3 i3 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9" +
-" f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9";  // 22 coefficients and SV
+"a1 i3 i3 f7 f7 f7 f7 f7 f7 f7 f7 f7 f7 f7" +
+" f7 f7 f7 f7 f7 f7 f7 f7 f7 f9 f10 f10 f9 f8"; // 24 coefficients and SV
 
 //*************************************************************************//
 
@@ -482,10 +482,10 @@ const tableProperties allProperties[] = {
     "geodetic/IGRF",		"http",  		"ascii",
     &IGRF,			IGRFCol,		0,
     vector<String>(),		vector<uInt>(),		vector<TableColumn*>(),
-    "IGRF10 reference magnetic field",			"earthField",
+    "IGRF12 reference magnetic field",			"earthField",
     False,			"",			vector<String>(),
     &IGRFFormat,		vector<formatDescr>(),
-    { "www.ngdc.noaa.gov", "IAGA/vmod", "igrf10coeffs.txt" } },
+    { "www.ngdc.noaa.gov", "IAGA/vmod", "igrf12coeffs.txt" } },
 
   //**********************************************************************//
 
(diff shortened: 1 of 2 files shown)


tammojan:
Thanks @olebole . We'll have a look at it after summer vacation (mid August).

tammojan:
```
The following tests FAILED:
	  3 - tEarthField (Failed)
	  4 - tEarthMagneticMachine (Failed)
	  7 - tMEarthMagnetic (Failed)
```
This does not show up in the travis tests, because these use the measures data from the Astron FTP. I have an idea what is wrong (the order of coefficients may have changed), and will have a look at it. But for now, don't merge please.

measures/apps/measuresdata/measuresdata.cc
```diff
@@ -244,8 +244,8 @@ const String &IERSpredict2000Format = IERSpredictFormat;
 
 // IGRF
 const String IGRFFormat = String("") +
-"a1 x2 i3 i3 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9" +
-" f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9 f9";  // 22 coefficients and SV
+"a1 i3 i3 f7 f7 f7 f7 f7 f7 f7 f7 f7 f7 f7" +
+" f7 f7 f7 f7 f7 f7 f7 f7 f7 f7 f7 f7 f7 f7";  // 24 coefficients and SV
```
tammojan: The second line should be `" f7 f7 f7 f7 f7 f7 f7 f7 f7 f9 f10 f10 f9 f8";  // 24 coefficients and SV`

olebole:
I just updated the patch according to your comment.

tammojan:
Ok thanks @olebole. That passes the checks. Wim Brouw is currently looking into whether the model in casacore accurately follows the model in http://www.ngdc.noaa.gov/IAGA/vmod/igrf12.f , so I suggest waiting for his findings before merging this.

dpetry:
This is now also a CASA issue (CAS-8564). Please keep me in the loop. Thanks!

tammojan:
@olebole, could you change `const Int PQ_LEN = 65;` in measures/Measures/EarthField.h to `const Int PQ_LEN = 104;` ?

olebole:
@tammojan -- done.

tammojan:
I finally convinced myself that casacore with the IGRF12 coefficients produces the same results as the [igrf12.f reference program](https://www.ngdc.noaa.gov/IAGA/vmod/igrf12.f), accurate to about 1 nT.

I'm currently evaluating what happens with (old code with new coefficients) and (new code with old coefficients), to see if that would lead to upgrade problems.

tammojan:
I've compared the result of the magnetic field computations with those from the reference program. To be specific, for all observatories for years 1998 to 2020, I computed the difference (in 2-norm) between the casacore outcome and the reference outcome. Because the reference implementation is in fortran, a relative difference of 10<sup>-5</sup> can be expected (single precision).

Ideally, one should update both casacore and the measures data at the same time. However, if only one gets updated, the error in the magnetic field does not become worse than it already was. I think this means any update path is fine.

 | Casacore before merge | Casacore after merge
--- | ------------ | -------------
**ITRF10 data** | rel. error 10<sup>-2</sup> | rel. error 10<sup>-2</sup>
**ITRF12 data** | rel. error 10<sup>-3</sup> | rel. error 10<sup>-5</sup>

dpetry:
Great work, Tammo! Have you heard from Wim again w.r.t. this or is your comparison essentially
what he wanted to do?

tammojan:
> Have you heard from Wim again w.r.t. this or is your comparison essentially
> what he wanted to do?

This is what Wim suggested to do.

tammojan:
@dpetry , do you think this is ready to merge?

dpetry:
I would like to merge this but it seems the Travis checks have failed.
I can test the changes in my local build but would still like to understand what problems Travis had ...

tammojan:
Yuk, indeed the IGRF-specific tests fail. I'll look into it.

tammojan:
This PR is superseded by #584

PR #481, opened 2016-08-10
Fix wrong loop condition and minimize scope

juliantaylor:
The blanked value loop did not use the loop counter in the condition.
Minimize scope of counter variables to avoid this type of mistake.


dpetry:
Interesting. How did you find this? I am wondering how a developer could have arrived at this strange setup. Maybe for debugging something?

juliantaylor:
clang-3.8 throws a warning for this problem. It looks like it was just a typo (+ a later warning fix that didn't realize the issue). 
I assume this code path is never used as it resulted in an infinite loop. I should have a look at creating a testcase for it.

PR #487, opened 2016-08-26
CAS-8918. Add first exposure times data structure to scans and subscans data structures

dmehring:


ms/MSOper/MSMetaData.cc
```diff
@@ -3857,6 +3907,23 @@ MSMetaData::_getChunkSubScanProperties(
     return make_pair(scanProps, mysubscans);
 }
 
+void MSMetaData::_modifyFirstExposureTimeIfNecessary(
+    FirstExposureTimeMap& current, Int dataDescID,
+    Double time, Double exposure, const Unit& eunit
```
juliantaylor: please move this overload next to the other one in the file
you could probably also implement the Map, Map overload in terms of this one to remove a little duplicated logic.

ms/MSOper/MSMetaData.h
```diff
@@ -562,10 +569,14 @@ class MSMetaData {
 
     uInt nPol();
 
+    // DEPRECATED
     // get a map of data desc ID, scan number pair to exposure time for the first time
     // for that data desc ID, scan number pair
     vector<std::map<Int, Quantity> > getFirstExposureTimeMap();
 
+    // get map of scans to first exposure times
+    map<ScanKey, FirstExposureTimeMap> getScanToFirstExposureTimeMap(Bool showProgress) const;
```
juliantaylor: std::map, also std::vector in getFirstExposureTimeMap return above

dmehring:
I opted not to implement one in terms of the other because of possible performance issues due to necessary additional method calls to do so. It gets called for every row iteration in the main table, so in theory, for large data sets, there could be an (admittedly probably relatively small) performance hit.

juliantaylor:
there should be pretty much no overhead as the Quantum is a template class so the get functions are in a header and can be inlined. But its a minor complaint so I can let it slide.

dmehring:
Thanks!

PR #488, opened 2016-08-29
Fix possible uInt overflows and tighten scopes

juliantaylor:
Several loops iterating arrays used uInt as counter, this could overflow
for large arrays.
Also minimize the scope of the accumulators in some of the computation
loops to avoid unnecessary stack spills.


juliantaylor:
a concern is that to get the compiler to emit decent code the summation order has to be changed, so it does influence results. On average the results should be better as its now does one recursion of a pairwise summation, but I'm not sure if the very minor performance improvement is worth a change in result, even if it is better.

PR #490, opened 2016-09-02
Further mods supporting changeover to casacore namespace.

JimNrao:
There were apparently some omissions in the previous commit as well as updates
to the codebase where people reintroduced the "casa" namespace.


gervandiepen:
It looks as if the erroneous use of casa:: instead of casacore:: was mainly done by CASA people. Please tell them to use namespace casacore.

PR #491, opened 2016-09-08
Cas 8783 1 bug fixes

dmehring:
ensure _centerValue is computed in FitToHalfStatistics
FitToHalf stats extrema computation bug fixes
remove trailing character from string stream used for output in ImageStatistics.

### DIFF
--- a/scimath/Mathematics/FitToHalfStatistics.tcc
+++ b/scimath/Mathematics/FitToHalfStatistics.tcc
@@ -128,6 +128,7 @@ void FitToHalfStatistics<CASA_STATP>::_getRealMinMax(
         CountedPtr<AccumType>& realMin, CountedPtr<AccumType>& realMax,
     CountedPtr<AccumType> knownMin, CountedPtr<AccumType> knownMax
 ) {
+    _setRange();
     realMin = new AccumType(_centerValue);
     realMax = new AccumType(_centerValue);
     if (knownMin.null() || knownMax.null()) {
@@ -322,7 +323,7 @@ void FitToHalfStatistics<CASA_STATP>::setCalculateAsAdded(
 ) {
     ThrowIf(
         c, "FitToHalfStatistics does not support calculating statistics "
-            "incrementally as data sets are added"
+        "incrementally as data sets are added"
     );
 }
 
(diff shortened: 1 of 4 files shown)


scimath/Mathematics/FitToHalfStatistics.tcc
```diff
@@ -128,6 +128,7 @@ void FitToHalfStatistics<CASA_STATP>::_getRealMinMax(
         CountedPtr<AccumType>& realMin, CountedPtr<AccumType>& realMax,
     CountedPtr<AccumType> knownMin, CountedPtr<AccumType> knownMax
 ) {
+    _setRange();
```
juliantaylor: if its needed here, shouldn't it also be needed in `getMinMax` and `getQuantiles`?

dmehring: I don't think so. getMinMax() calls ConstrainedRangeStatistics::getMinMax() as necessary. That method calls _setRange() as defined in the appropriate object.  getQuantiles() calls getMinMax() as necessary, which as discussed ultimately calls _setRange() on the appropriate object. It also calls getNPts() as necessary, which will call the ConstrainedRangeStatistics::getNPtps() which calls _setRange(). It then always calls _getRealMinMax(), which is now guaranteed to called _setRange().

The basic idea is that when the corresponding method in ConstrainedRangeStatistics is called, _setRange() will be called from there. The problem was that there were paths in getQuantiles() in which _getRealMinMax() was being called prior to ConstrainedRangeStatistics::getQuantiles() (which calls _setRange()) being called, so that _centerValue had not yet been set.

It is a bit convoluted I admit, but the goal was to minimize unnecessarily repeating code in derived classes and minimizing unnecessary repeated calls. In general, ConstrainedRangeStatistics methods deal with calling _setRange(), but in this case there were paths in which _setRange() needed to be called prior to a call to the corresponding method in the base class.

juliantaylor: is it ever wrong to call _setRange in these functions?
it only actually executes once as it has a fast path if it was already called
there is the chance that there is some other (maybe future) way of executing this stuff where it might again not have been called.
If the call is always correct as its either needed or a no-op, it might be safer to just call it everywhere its results are needed.

dmehring:
OK, I've added _setRange() calls to the relevant places in the relevant public methods.

juliantaylor:
thanks
still seems the way it works could be improved, but I don't really know these classes well enough to make a concrete suggestion.

PR #496, opened 2016-09-16
Cas 8984 LatticeStats reimplement zero-ing out of quantile stats when they are not computed, et al.

dmehring:
CAS-8984 Reimplement inefficient method for zero-ing out quantile stats when they are not to be computed.
Catch noncontiguous array exceptions when old TiledApply method is used so that the new Stats framework method can be tried. Avoid that exception for smallish lattices by always using the new
framework method for them.

### DIFF
--- a/lattices/LatticeMath/LatticeStatistics.tcc
+++ b/lattices/LatticeMath/LatticeStatistics.tcc
@@ -841,8 +841,10 @@ Bool LatticeStatistics<T>::generateStorageLattice() {
         _doStatsLoop(nsets, pProgressMeter);
     }
     pProgressMeter = NULL;
-    // Do robust statistics separately as required.
-    generateRobust();
+    if (doRobust_p) {
+        // Do robust statistics separately as required.
+        generateRobust();
+    }
     needStorageLattice_p = False;
     doneSomeGoodPoints_p = False;
     return True;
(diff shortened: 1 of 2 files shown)


lattices/LatticeMath/LatticeStatistics.tcc
```diff
@@ -987,11 +1008,36 @@ void LatticeStatistics<T>::_doStatsLoop(
...
+        pStoreLattice_p->putSlice(zeros, start);
+        start[ndim - 1] = LatticeStatsBase::QUARTILE;
+        pStoreLattice_p->putSlice(zeros, start);
+    }
 }
 
 template <class T>
 void LatticeStatistics<T>::generateRobust () {
-    Bool showMsg = haveLogger_p && doRobust_p && displayAxes_p.nelements()==0;
+    if (! doRobust_p) {
```
juliantaylor: might be nicer to just not call the function when its not needed instead of having some extra state here

the function used to just zero out stuff when this was false before?

dmehring:
Yes. Previously if doRobust_p was False, it just zero-ed out the relevant stats in a very inefficient way. So I removed the inefficient code and now the zeroing out is done elsewhere in a more efficient manner.

PR #507, opened 2016-10-11
Travis: Install gfortran on OSX, fixes #503

### LINKED ISSUE #503: Travis tests fail on OSX: no gfortran
As can be seen in e.g. https://travis-ci.org/casacore/casacore/jobs/166159149 , the test fails on OSX because gfortran is not found. Also, the test script could be improved to handle this a bit cleaner.

tammojan:


juliantaylor:
this doesn't take into account the cache
also instead of caching gcc it probably makes more sense to cache the compiled stuff so gcc is not needed in the normal case

tammojan:
Agreed. However, I don't want to change the cache now because then the boost-python binary will complain about a wrong version (and I don't have the right Mac to create a new binary).

tammojan:
Also, we do need gfortran also just for compiling casacore, so I think it makes sense to cache gcc.

juliantaylor:
I assume the actual problem is travis removed gfortran from the default image?
We should probably just ask why and if that can be changed. No gfortran in travis is going to break a lot of stuff in other projects. I can ask around if you haven't yet.

tammojan:
I actually think gfortran was never in the travis image, and was installed as a dependency of one of the other brew packages. But I'm not sure.

juliantaylor:
I don't think so, gcc/gfortran was never cached so it should have not been able to work
I think its more related to the new xcode that has been rolled out a few days ago.
we could check that by rolling back to that via `osx_image: beta-xcode6.1`

juliantaylor:
when will you have time to create a new boost bottle?
As this will happen often I would recommend just removing the python3 build on mac, its not very important anyway as linux covers it.

gijzelaerr:
agreed. I just pushed to a branch and then discovered this discussion.

https://github.com/casacore/casacore/pull/518

PR #509, opened 2016-10-19
taql constant integer set optimization

juliantaylor:
Optimize `WHERE x in [integers]`

Replace the range limited constant lookup table with a set, much simpler code and consistent performance regardless of range. This avoids degrading to quadratic performance when the set range is too large.
It still seems like the wrong place to me to do this optimization, to me it would be more logical to have the implemented in the right hand operand instead of the operator, but I don't know the design of the taql very well so I left the current structure.
A possible improvement could be to use an c++11 `unordered_set` to kept he constant time, but its is not really relevant enough in practice do bother.

Also removed a quadratic performance issue in `TableExprNodeSet::add` when you do have a very long input taql set. Its cause is `casacore::Block::resize` not overprovisioning so it is very expensive for the `+1` resize done there. `std::vector` does this so and is sufficient for the simple pointer container.

If merged also needs an update to the 199 casa note which documents the old optimization.

### DIFF
--- a/tables/TaQL/ExprLogicNode.cc
+++ b/tables/TaQL/ExprLogicNode.cc
@@ -290,8 +290,7 @@ TableExprNodeINInt::TableExprNodeINInt (const TableExprNodeRep& node,
 void TableExprNodeINInt::convertConstChild()
 {
   if (rnode_p->isConstant()  &&  rnode_p->valueType() == VTArray) {
-    // Convert set/array to a Bool array (for direct lookup) if the range
-    // is sufficiently small.
+    // Convert array to a set for lookup
     MArray<Int64> values = rnode_p->getArrayInt(0);
     Array<Int64> arr(values.array());
     if (values.hasMask()) {
@@ -299,21 +298,8 @@ void TableExprNodeINInt::convertConstChild()
       arr.reference (values.flatten());
     }
     if (! arr.empty()) {
-      minMax (itsMin, itsMax, arr);
-      Int64 sz = itsMax - itsMin + 1;
-      if (sz <= 1024*1024) {
-        itsIndex.resize (sz);
-        itsIndex = False;
-        Array<Int64>::const_iterator arrend = arr.end();
-        for (Array<Int64>::const_iterator iter=arr.begin();
-             iter!=arrend; ++iter) {
-          itsIndex[*iter - itsMin] = True;
-        }
-        if (itsDoTracing) {
-          cout << "  created IN index of size " << sz
-               <<" offset=" << itsMin << endl;
-        }
-      }
+      itsIndexSet.clear();
+      itsIndexSet.insert(arr.begin(), arr.end());
     }
   }
 }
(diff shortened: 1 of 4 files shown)


alxames:
FYI: Ger is out-of-office. He returns next Thu (Oct 27th).

gervandiepen:
Nice change. I do agree that using a set makes the code simpler, but a find in a set has O(logN) complexity, while the Bool array used now uses a single lookup. Don't you fear a degradation in performance for often used cases like antenna selection?

juliantaylor:
the lookup is faster though that doesn't really matter that much, what is important is avoiding the square complexity.
On warm cache selecting 50000 rows via a set from a 61 million row ms takes around 2.8 seconds with the constant lookup and 3.1 seconds with the set. Faster but imo insignificant.
On the other hand when the constant lookup fails (due to the range being larger than 2014*^2) selecting ten times less rows takes 5 minutes, I didn't want to wait how long 50000 rows take.
As this pathologically bad there is likely no code out there that hits this problem so there should not be any noticeable performance regressions in practice.

But I can still add the constant lookup back if you still want it.

gervandiepen:
I agree that in almost all cases the set implementation is fine. Only if the set is very large (millions of elements), it takes much more time (already in creating the set).
I've experimented with an unordered_set which is faster, but is also slow in creating the unordered_set for many, many elements.
I think the case of millions of elements is only theoretical, so I think the change is fine.
Note that in something like 'COLUMN in [0:10000000]', the set is turned into an array with 10000000 elements. Another optimisation could be to avoid that for the IN operator, but who would write something like that?

Shall I do the merge or do you want to make other changes?
I'll update note 199.

Regarding your remark about the wrong place of optimisation. In an expression like
'date() + [0:31]' the sequence needs to be a vector, not a set. So the operator determines if it can be turned into a set which is only the case for the IN operator.

juliantaylor:
Is there a practical case where the IN operator is used on M rows with a N set elements where M < log(N)?
Only then I would consider the set to be a problem and we need something with even better complexity (e.g. the unorderd_set). But I can't imagine this to currently exist given that right now this has M * N complexity which is so bad people would have complained about it already.
I'd say lets see if this shows up in anybodies performance profiles before we try to optimize further.
This code right now has nice and predictable performance, always log(N) with no surprises on the size of N.

Concerning the place of the optimization, I was thinking about having it more like python where the IN operation is part of the container (the `__contains__` dunder). So date() + [1:31] would optimize itself to a set if possible so it is only stored once in the appropriate structure instead of the IN operator optimizing its operand into more appropriate copies of the actual operand.
This would for example allow [1:10000] to optimize itself into a range object which only needs start stop and stride as storage and has constant time IN operations.

juliantaylor:
I do not plan on making any changes, so you can merge.
But if you do want the small set optimization set back first I can do so, that is no issue.

gervandiepen:
If ever needed, we can later add back the small set optimisation.

PR #510, opened 2016-10-27
CAS-9146 increase tolerance for determining if direction pixels are square

dmehring:


JimNrao:
The change was minimal, passed two builds and failed in a CMAKE command on the other two; it appears to be a Travis-related problem to me.

PR #512, opened 2016-10-31
CMakeLists.txt mod to fix PR #506

JimNrao:
The newly added Json*.h files weren't installed into the include directory
during build and would break casacore client builds.


JimNrao:
Roll back of PR 506 which was causing client builds to break.

PR #515, opened 2016-11-02
Allow alternate spelling of SOURCE_ID random parameter that is used in

kettenis:
FITS-IDI files generated by difx2fits.


kettenis:
This small change makes sure field IDs are properly assigned for data correlated by the DiFX correlator, such as VLBA and GMVA data.

Issue #505, opened 2016-10-11
Add Json support


gervandiepen:
Currently, ImageConcat and ImageExpr save their data as an AipsIO file.
It would be better to save it in text form to be readable (and possibly editable); the best form to use is Json.
Besides supporting standard Json, it should have support for conversion of Array objects to/from Json as well as conversion of (D)Complex numbers.

PR #517, opened 2016-11-02
Add Json support for ImageConcat and ImageExpr

gervandiepen:
Issue #505
CASA requested the creation of the ImageConcat and ImageExpr output file in a named directory instead of a named file to be able to move them more easily.
For that purpose it is also more convenient to have the output in text format instead of AipsIO format, so it can be edited if needed. It was chosen to output it in Json format. Code has been created to easily read and write Json files.

This PR is the same is basically the same as #506 with the addition of the .h files to CMakeLists.txt.
It is known to break the CASA build, which has to be changed asap. Also the CASA pipeline feature of moving an ImageConcat file to a directory has to be removed, because ImageConcat now creates (or uses) a directory as agreed with Kumar Golap.


JimNrao:
Dank u wel, Ger.

PR #518, opened 2016-11-02
disable python3 for osx, install gfortran for osx

gijzelaerr:


tammojan:
The test says `make: gfortran: No such file or directory`, see https://travis-ci.org/casacore/casacore/jobs/172578517#L96

gijzelaerr:
probably because the osx Cellar is cached. not sure how to remove that cache? I didn't set that up.

juliantaylor:
in the travis ui there is a more options tab, under there is the caches folder where you can delete caches. I have removed all caches and restarted, lets hope it finishes within the timeout without a cache.

juliantaylor:
and it fails, apparently it is in gcc instead of gfortran.

tammojan:
Yes it should be gcc. Sorry for not noticing earlier, I did this in #507 (which I'll close in favor of Gijs' better attempt).

juliantaylor:
this PR is also not turning of python3 in `.travis/before_install.sh`

gijzelaerr:
sorry guys, this week i'm moving back to south africa so I can't look into this. @tammojan can you pick this up?

juliantaylor:
looks like the mac non c++11 build is too slow now, maybe the new clang is slower.
Things we could try:

set optimization level to 1 for this build, maybe it is fast enough, disable this build or bootstrap the cache via only partially building casacore.

juliantaylor:
also tConvert is failing in the new mac, I saw it fail on ubuntu machines with stack smashing errors, not sure if mac has the same issue. Verbose builds might be useful.

tammojan:
Apparently a problem with Boost Python on OSX, maybe linking against wrong python version or someting. To be investigated..

tammojan:
Superseded by #654.

PR #521, opened 2016-11-03
Don't build tests by default

juliantaylor:
The autotools behaviour of building tests only on the make check target
is much better as it avoids having to go through the length test
compilation and uncacheable linking during development.
Try to emulate this behaviour with cmake by adding a check target and
only building the test directories on this target.
As you can't add test dependencies on the test target it is
disfunctional when check has not been executed first.

For backward compatibility when BUILD_TESTING is set everything remains
as is, so make builds the test and make test runs them.


juliantaylor:
the effect is that yes, most notably a normal make will never build the tests.
The implementation is that testing is always on but not built in the all target, so only on demand.
The drawback is that without -DBUILD_TESTING=ON make test will not work without a prior make check.
Also the make check is much less featureful than autotools, you can't only build the tests and not run them (make check TESTS=) and you can't execute specific tests via make (though ctest can do that).
cmake is kind of garbage ...

PR #522, opened 2016-11-03
Fix some memory leaks

juliantaylor:
Fix leak in FITS::BinaryTable::fullTable and some minor ones in tests.


juliantaylor:
reduces the number of tests that fail with address sanitation but a bunch of tricky ones remain.

Issue #519, opened 2016-11-02
Dynlib.cc does not find libraries cotaining SOVERSION


olebole:
When only the shared library with a file name according to the `SONAME` is installed (which is  the usual case, at least on Debian), `DynLib::DynLib()` does not find the library. This causes f.e. data conversion fail, like
```
Traceback (most recent call last):
  File "tools/create_casacore_table.py", line 80, in <module>
    row['elevation']))
  File "/usr/lib/python3/dist-packages/casacore/tables/table.py", line 99, in taql
    tab = table(cmd, tabs, _oper=2)
  File "/usr/lib/python3/dist-packages/casacore/tables/table.py", line 271, in __init__
    Table.__init__ (self, tablename, tabledesc);
RuntimeError: Error in TaQL command: 'using style Python calc meas.position('itrf',41.586683deg,43.826167deg,968m,'wgsll')'
  Shared library meas found in CASACORE_LDPATH nor (DY)LD_LIBRARY_PATH
libmeas.so: cannot open shared object file: No such file or directory
libmeas.dylib: cannot open shared object file: No such file or directory
libcasa_meas.so: cannot open shared object file: No such file or directory
libcasa_meas.dylib: cannot open shared object file: No such file or directory
```
The content on Debian's [`libcasa-meas2` package](https://packages.debian.org/sid/amd64/libcasa-meas2/filelist) is only
```
lrwxrwxrwx 1 root root      21 Sep 26 09:52 /usr/lib/x86_64-linux-gnu/libcasa_meas.so.2 -> libcasa_meas.so.2.1.0
-rw-r--r-- 1 root root  556024 Sep 26 09:52 /usr/lib/x86_64-linux-gnu/libcasa_meas.so.2.1.0
```
while the only the [`casacore-dev` file contains](https://packages.debian.org/sid/amd64/casacore-dev/filelist) (among others) the missing link
```
lrwxrwxrwx 1 root root      17 Sep 26 09:52 /usr/lib/x86_64-linux-gnu/libcasa_meas.so -> libcasa_meas.so.2
```
When searching for the library, `Dynlib.cc` should also look for the files with `SOVERSION` attached.

gervandiepen:
So it basically means that the version must be an optional DynLib function argument, otherwise it won't work for packages built outside Casacore which will have different version numbers.
I guess a -DSOVERSION=N has to be added to the compile command to know the version where
N is filled by CMake commands.
[...]

PR #525, opened 2016-11-04
Added support of shared library version

gervandiepen:
Issue #519


gervandiepen:
DynLib got a new constructor to pass a shared library version.
If given, it is tried first before trying it without version. It makes a difference between MacOS and other OSs where the version is inserted in the name.

casa/OS/DynLib.cc
```diff
@@ -152,8 +124,91 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
...
+                       const std::string& version,
+                       const std::string& funcName)
+  {
+    std::string fullName = tryCasacorePath (library, prefix, version);
+    if (fullName.empty()) {
+      fullName = tryOpen (library, string(), prefix, version);
+    }
+    if (itsHandle == 0) {
+      throw AipsError ("Shared library " + library +
+                       " found in CASACORE_LDPATH nor (DY)LD_LIBRARY_PATH\n"
```
juliantaylor: should probably be `not found?
Also the error should state the version and prefix that has not been found

tables/DataMan/DataManager.cc
```diff
@@ -316,7 +317,8 @@ DataManagerCtor DataManager::getCtor (const String& type)
         libname = libname.substr (0, pos);
     }
     // Try to load and initialize the dynamic library.
-    DynLib dl(libname, string("libcasa_"), "register_"+libname, False);
+    DynLib dl(libname, string("libcasa_"), STRING(CASACORE_MAJOR_VERSION),
```
juliantaylor: The major version should not be connected with a binary version, also bugfixes can require changing compatiblity.
If you want to couple it regardless, the next casa version would need to be 3.0.0 not 2.2.0 as there have been breaking changes merged.

gervandiepen: Do you propose to use the SOVERSION specified in CMakeLists.txt, thus passing it in as a compile variable? It should be the same as the versions defined in version.h, but I agree it can be forgotten to update version.h. So what is best?
Ole suggested to only use the major version, not minor and patch, but it is no problem to do so.

I'll improve the error message. The error messages from dlopen show the paths and versions tried (which become part of the error message), thus there is no need to include it in the error message.

juliantaylor: Passing it in from cmake probably makes most sense as there the soversion is defined.
If you couple it with the major version it must be bumped every time there is a breaking release which can be often. It is usually better to keep the two version numbers separate as the describe to related but different things.

Also note that CASA uses the date as the soversion (which is a bit overkill in the other direction), so using the major version like this probably breaks it.

gervandiepen:
Darn, why can't it merge?

juliantaylor:
you have a conflict with the CMakelist.txt in master, you should rebase this branch against it.

Adding SOVERSION to all files as preprocessor define is a bit inconvinient for CASA due to them using the date as soversion, this will cause all files to be uncacheable, at least in one build mode.
Is it possible to only add the define only the the files that use it?

Issue #516, opened 2016-11-02
A C++11 feature is used


gervandiepen:
It must still be possible to build Casacore without C++11 support. However, in StatsTiledCollapser a C++11 feature was introduced in lines 282-283 by using the functions real() and imag() to set a complex value.
It should be replaced by:
     storage[i] = DComplex(realStorage[i], 0);
An ifdef can be used to use the new feature.

PR #526, opened 2016-11-04
Removed use of C++11 feature

gervandiepen:
Issue #516 
Removed use of complex function real and imag to set the complex values as it is only supported in C++11. Until we all agree, no C++11 features should be used (unless ifdef-ed).


juliantaylor:
thanks, on what system did this cause problems?
both clang and gcc support this also outside of c++11. Otherwise we should have seen it in the CI.

juliantaylor:
the measures tarball from astron.nl is not reachable anymore, can someone check what is wrong there?

tammojan:
Checking now.

tammojan:
There was an issue with the measures cron job at ASTRON after an update to the FTP servers. 

The measures data should now be accessible again.

gervandiepen:
It caused troubles on OS-X using an older clang compiler.

PR #527, opened 2016-11-07
new casa third-party path

schiebel:
casacore builds with a casa build should prefer the path for new casa third-party pkgs


juliantaylor:
please rebase the branch against the origins master, there is an unnecessary merge in the branch

PR #528, opened 2016-11-08
fix free memory determination with cgroups

juliantaylor:
Unlimited memory usage can be declared by a very large number but
smaller than size_t.max. Then the free memory is computed using this
wrong very large number.
So instead accept memory limits only when it is smaller than the total
memory of the physical machine determined earlier.

Also cleanup the code and add use of Linux 3.14 MemAvailable for the non cgroup code path. MemAvailable is more accurate than MemFree + MemCached.


tammojan:
I trust you have tested this, nice to see that you also thought of people with linuxes before 2014.

PR #533, opened 2016-11-15
Use c++11 by default

tammojan:
As suggested earlier by @alxames , using C++11 by default is probably good.


alxames:
Hopefully we are not flooded with complaints about build failures from users with pre-C++11 compilers.

tammojan:
> Hopefully we are not flooded with complaints about build failures from users with pre-C++11 compilers.

I'll happily direct all those users to this issue, with as suggested fix to read the documents and add `-DUSE_CXX11=OFF` to the CMake command line.

Issue #511, opened 2016-10-31
Release 2.2.0 ?


tammojan:
If there are no objections, I would like to release version 2.2.0. If no objections appear here, I will make prepare a release on Thursday November 2.

gijzelaerr:
@olebole is cool with this, he suggest we do a 2.0.0RC release. We can package that, see if everything works correctly and then do the final release.

tammojan:
I suggest to wait with the release until the #516 and #503 have been resolved.

tammojan:
Also I'd like to wait for #479.

juliantaylor:
#528 should probably go in too

juliantaylor:
hm you didn't bump the SOVERSION, it is still 2 as in the v2.1.0, you might want to inform packagers.

PR #535, opened 2016-11-15
Prepare for release 2.2

tammojan:
We'd like to release version 2.2 (#511). Comments on the changelog are welcome here.


CHANGES.md
```diff
@@ -1,3 +1,20 @@
+# 2.2.0
+
+## General
+- Lots of bug fixes and improvements
+- Tests are not built by default anymore, use `-DBUILD_TESTING=True`
```
juliantaylor: a bit late, but this was always the case.
the new thing is that the tests can _always_ be built regardless of cmake configuration with `make check`

PR #536, opened 2016-11-17
Cas 8875 MSFITSOutput now writes antenna diameters...

dmehring:
MSFITSInput now uses algorithm supplied by Jen Meyer to write ALMA antenna diameters if no ant diameters are supplied in the uvfits file.
Changed a couple ++ operators from postfix to prefix as loop incrementers
Initialized a object field that wasn't being initialized that was causing a unit test to fail under some circumstances.


dmehring:
@tammojan This failure looks like a travis/CI system issue to me and not related to my code. Could you please check? Thanks

tammojan:
Indeed, only the OSX builds fail. The linux builds are fine, so in this case it's fine to ignore the travis result when merging (I haven't reviewed the change).

PR #544, opened 2016-11-27
Fix minor issue in JsonKVMap.h

tammojan:
In JsonKVMap.h, the namespace was not added to `map`. Not sure why this is allowed in new compilers, but GCC 4.4.3 complains about it, and I agree with it.


tammojan:
Duplicate of #545

PR #548, opened 2016-12-05
Cas 8482 bugfix ImageBeamSet::rotate() now rotates min and max area beams

dmehring:


dmehring:
@gervandiepen Ger, I know you prefer not to review changes made by our team, but this is an image code related bug fix (and a few files in which I converted tabs to spaces), so I think you should review it as you understand this code most. Thanks.

PR #552, opened 2016-12-14
CAS-9358: Necessary fixes for antenna-related methods

dmehring:
because a single antenna name can be listed multiple times in the ANTENNA table. Main changes are to private data structure and associated methods which make use of it. Note that some public interface changes are necessarily not backward-compatible.


tammojan:
Why would an antenna name be listed multiple times in the ANTENNA table?

dmehring:
In concatenated MSes representing multiple observations, the same antenna can be used but also may have multiple stations if it has been moved between observations.

PR #559, opened 2017-01-25
Make TableProxy Record to TableDesc static methods public

sjperkins:
Previously, much of the functionality for constructing a TableDesc
object from the more generic (and python convertible) Record object class was
static private. This meant that it was only possible to perform this
conversion within TableProxy methods.

By contrast, the functionality for constructing a Record
from the internal TableDesc (via TableProxy.getTableDescription) is public.

This commit changes many of the static private methods for constructing
a TableDesc from a Record public. By doing so, it becomes possible to
create more flexible C++ functions for creating various types of tables
from python-casacore. In particular, this makes it possible to create
default MeasurementSet tables from python while supplying additional table
description parameters. For example:

```cpp
TableProxy makems(const String & name, const Record & tabdesc)
{
    TableDesc required_td = MeasurementSet::requiredTableDesc();
    TableDesc user_td;
    String message;

    TableProxy::makeTableDesc(tabdesc, user_td, message);
    required_td.add(user_td);

    SetupNewTable new_table(name, required_td, Table::New);
    MeasurementSet ms(new_table);
    ms.createDefaultSubtables(Table::New);
    return TableProxy(ms);
}
```

It may be better at some point to abstract the Record to/from TableDesc
functionality out of TableProxy at some point, but this simple change at
least makes the functionality accessible.

See also: https://github.com/casacore/python-casacore/issues/61

### DIFF
--- a/tables/Tables/TableProxy.cc
+++ b/tables/Tables/TableProxy.cc
@@ -924,13 +924,24 @@ Record TableProxy::getTableDesc(const TableDesc & tabdesc, Bool cOrder)
 {
     Record rec;
 
+    // Convert columns
     for (uInt i=0; i<tabdesc.ncolumn(); i++) {
       const ColumnDesc& columnDescription = tabdesc.columnDesc(i);
       rec.defineRecord (columnDescription.name(),
                 recordColumnDesc (columnDescription, cOrder));
     }
+    
+    // Convert hypercolumns
+    rec.defineRecord ("_define_hypercolumn_",
+        recordHCDesc (tabdesc));
 
-    rec.defineRecord ("_define_hypercolumn_", recordHCDesc (tabdesc));
+    // Convert keywords
+    rec.defineRecord("_keywords_",
+        tabdesc.keywordSet().toRecord());
+    // Convert private keywords
+    rec.defineRecord("_private_keywords_",
+        tabdesc.privateKeywordSet().toRecord());
+
 
     return rec;
 }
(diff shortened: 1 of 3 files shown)


gervandiepen:
Nice change. I agree that at one time it would be better to move the Record-TableDesc converters elsewhere.

I think that the coding style should not be changed to the preference of the one making a change, otherwise we end up with a mix of coding styles in Casacore. I know there is no formal Casacore coding style guide, but as much as possible the style (e.g., placement of braces) should be kept as is.

sjperkins:
In terms of style, should I be aiming for 2 space indent and braces like

```cpp
for(int i=0; i < 100; ++i) {
  do_stuff();
}
```

as opposed to

```cpp
for(int i=0; i < 100; ++i)
{
    do_stuff();
}
```

?

gervandiepen:
Old code usually has a 4 space indent, but newer code 2 bytes. Some files
(mostly in images) even have 3. I'm not too concerned about that.
Indeed braces are usually on the same line as the for or if. Else is
usually done like
   } else {
   } else if (cond) {

One-liners like
   if (cond) stmt;
are discouraged because it makes it harder to set a breakpoint in
debuggers; furthermore, coverage tools cannot always distinguish cond and
stmt.

gervandiepen:
In the coming time I'll make a tTableProxy to test its functionality. Then I'll also move your test from tTableRecord to that file.

sjperkins:
Great thanks.

PR #566, opened 2017-02-02
add support for builds with Intel compilers

schiebel:
Explicit insertion of the Intel compiler's libraries into the link lines for libraries and executables is necessary for packaging CASA because the Intel libraries do not include a runpath (which is how we make CASA distributions relocatable).


tammojan:
Sorry for not responding to this pull request earlier. Is there no way to accomplish this without modifying every individual CMakeList.txt ?

schiebel:
It's a fair question. I do not know of another way, but perhaps the changes (in the lower-level CMakeList.txt) should be changed from referencing ${INTEL_LIBS} to referencng ${CASACORE_ARCH_LIBS}, i.e. something more generic, so that in the future this hook can be reused.

tammojan:
I don't know another way to accomplish this either. Could you change the variable name to the generic `${CASACORE_ARCH_LIBS}`? I'll do the merge afterwards.

schiebel:
OK I think this pull-request is ready... thanks!

tammojan:
Thanks, it looks ok to me.

Some nitpicking about your git usage: You have three commits here: the initial try, then a merge of the master branch, then the improved version of the first commit. In principle, we only need the one commit for a 'clean history'.
Merging in the master branch (the second commit) is almost never necessary. A good way to accomplish the same is to rebase the feature branch (`intel-compilers`) onto the master.
To merge the two real commits, you can use `git rebase -i HEAD~2` or so, to squash the commits into one.

A nice thing about git is that I can do this for you, so I've done this in #575, which I'll merge when the tests pass.

PR #568, opened 2017-02-07
Added keyChange() method to enable clients to query iteration boundary

gmoellen:
Added keyChange() method to enable clients to query iteration boundary key changes.  Also made MSIter methods virtual and corrected const-ness so derived class are possible.


ms/MeasurementSets/MSIter.cc
```diff
@@ -361,6 +361,7 @@ void MSIter::origin()
 {
   curMS_p=0;
   checkFeed_p=True;
+  if (timeComp_p) timeComp_p->setOffset(0.0);
```
SoftwareRancher: The call to setState already resets timeComp_p's offset, so this addition appears to be redundant.

ms/MeasurementSets/MSIter.h
```diff
@@ -60,11 +60,11 @@ class MSInterval : public BaseCompare
...
     virtual int comp(const void * obj1, const void * obj2) const;
-    Double getOffset() {return offset_p;}
-    void setOffset(Double offset) {offset_p=offset;}
-    Double getInterval() {return interval_p;}
+    Double getOffset() const {return offset_p;}
+    virtual void setOffset(Double offset) {offset_p=offset;}
+    Double getInterval() const {return interval_p;}
     void setInterval(Double interval) {interval_p=interval;}
-private:
+protected:
```
JimNrao: Having instance variable as protected is generally not a good practice.  Unnecessary as their are getters and setters available.

tables/Tables/BaseTabIter.h
```diff
@@ -104,10 +104,14 @@ class BaseTableIterator
...
     virtual BaseTable* next();
 
+    // Report Name of slowest column that changes at end of current iteration
+    String keyChange() const { return keyCh_p; };
+
 protected:
     BaseTable*             sortTab_p;     //# Table sorted in iteration order
     uInt                   lastRow_p;     //# last row used from reftab
     uInt                   nrkeys_p;      //# nr of columns in group
+    String                 keyCh_p;       //# the column name ending the current iteration
```
JimNrao: keyCh_p is a rather cryptic choice for an instance variable name.

tables/Tables/BaseTabIter.h
```diff
@@ -104,10 +104,14 @@ class BaseTableIterator
     // Return the next group.
     virtual BaseTable* next();
 
+    // Report Name of slowest column that changes at end of current iteration
```
JimNrao: Maybe you could expand on this description a bit?

ms/MeasurementSets/MSIter.cc
gmoellen: Sure, but the situation is far more subtle:  The setState() call in origin() comes _after_ a call to TableIterator::reset, which depends upon the internal state of timeComp_p.   If a 2nd pass through the iterator is done---i.e., a 2nd call to MSIter::origin()---the internal state of timeComp_p (seen by TableIterator::reset) is unpredictable, as it depends on the value of the offset_p at the end of the previous pass.  Note that setState() is used both in MSIter::origin() and in MSIter::advance(), mainly to set some meta-info about the current iteration.  IMO, it was quite sloppy to hide the timeComp_p reset in setState(), as timeComp_p is part of managing the behavior of the underlying iterator, not current iteration meta-info.
HOWEVER, this change was a legacy of testing variation in the MSIter specialization I am forming in CASA, and since I do not want to change observable behavior in MSIter with this revision, I should not have included it here.  I will remove this change and re-implement it in my specialization.  This will require MSIter::origin() to be virtual....

ms/MeasurementSets/MSIter.h
gmoellen: I'd agree completely were this a question of making the instance variables public.  My intent here was to make them directly accessible in a derived class that implements a smarter comp() function.  The comp() function is called an enormous number times (~nrow), and I'd hoped to avoid the function calls and obj copies of the getter methods (though perhaps the compiler can be relied upon to inline them).   
Nonetheless, I will make local (alas, redundant) instances of these variables in my derived class.

tables/Tables/BaseTabIter.h
gmoellen: By the standards of other var names in this class, this is a quibble.  OK, I'll change it to keyChangeAtLastNext_p.

tables/Tables/BaseTabIter.h
gmoellen: I will.

gmoellen:
The checks failed because I neglected a necessary revision to tMSIter.  I will fix this and address the 
issues noted in conversation above, and re-submit the pull request on Feb 8.

gmoellen:
OK, this pull request is now working.  I've revised the code as discussed.  
Jim (or Martin?) please reconsider the pull.    Thanks.

gervandiepen:
Why not return lastKeyChange as const String& instead of String? The latter means making a copy, although it is a cheap copy (counted ptr).
Furthermore, I think such a function can be inlined.

Clearing the lastKeyChange by assigning "" to it could be done better by assigning String() to it.
"" means it first has to be converted to an (empty) string.

SoftwareRancher:
Put the two days smallish mods Ger suggested and I'll flush it in; more to the table system are in his bailiwick.  I'm not in today so send me an email as I don't think GitHub notified me.On Feb 8, 2017 14:24, gmoellen <notifications@github.com> wrote:OK, this pull request is now working.  I've revised the code as discussed.
Jim (or Martin?) please reconsider the pull.    Thanks.

gmoellen:
Thanks for the advice, Ger, I've taken it.

Tests still passing, so I think this is ready.

Issue #572, opened 2017-02-16
Convolver: Copy constructor and assignment operator is incomplete


tnakazato:
Both copy constructor and assignment operator doesn't copy one private variable named `valid`. As a result, value of `valid` in copied/assigned object becomes uncertain. This sometimes causes problem because `theXfr`, Fourier transformation of the array `thePsf`, is initialized depending on `valid`.

Here is an example:
```
// dummy array. they should be properly configured in a real code
Array<Float> psf, model, result; 

// create Convolver object
// thePsf is set but theXfr is empty. valid is False.
Convolver<Float> original(psf); 

// copy object
// thePsf and theXfr are copied. valid is uncertain.
Convolver<Float> copied(original);

// convolution using original object should work properly 
// because theXfr will be initialized from thePsf
original.linearConv(result, model);

// convolution using copied object may not work if valid is True 
// because theXfr will not be initialized properly
copied.linearConv(result, model);
```

tnakazato:
I think the fix is simple, just copy `valid` in copy constructor and assignment operator. If it is appropriate fix, I can work on this issue.

PR #574, opened 2017-02-22
Issue #572 Properly copy "valid" when Convolver object is copied.

tnakazato:
This is a possible fix for the issue #572. Please merge it if appropriate.


tammojan:
Looks good, thanks.

PR #577, opened 2017-03-08
Travis: re-run failed test with verbose output

tammojan:
This change re-runs all failed tests on Travis. This should make it easier to spot why a test failed.


tammojan:
Changed original solution with a native ctest flag.

PR #578, opened 2017-03-08
Bug fixes + test update

sanbw:
Bug fixes in 
   - MSSelection::getChanFreqList 
    -MSSpwIndex::convertToChannelIndex

Some cleanup in MSSpwGram.cc

Updated tMSSelection.{out,run}

### DIFF
--- a/ms/MSSel/MSSpwGram.cc
+++ b/ms/MSSel/MSSpwGram.cc
@@ -130,9 +130,7 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
 			     Vector<Int>& selectedDDIDs) 
   {
     MSSpwParse thisParser(spwSubTable, ddSubTable, colAsTEN);
-    Int ret=baseMSSpwGramParseCommand(&thisParser, command, selectedIDs, selectedChans, selectedDDIDs);
-
-    return ret;
+    return baseMSSpwGramParseCommand(&thisParser, command, selectedIDs, selectedChans, selectedDDIDs);
   }
 
   int msSpwGramParseCommand (const MeasurementSet *ms, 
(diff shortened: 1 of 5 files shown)


ms/MSSel/MSSpwGram.cc
```diff
@@ -130,18 +130,16 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
 			     Vector<Int>& selectedDDIDs) 
   {
     Int ret;
-    //    MSSpwParse *thisParser = new MSSpwParse(ms);
-    MSSpwParse *thisParser = new MSSpwParse(spwSubTable, ddSubTable, colAsTEN);
+    MSSpwParse thisParser(spwSubTable, ddSubTable, colAsTEN);
     try
```
gervandiepen: There is no purpose for this try since you only rethrow the exception. The code can simply be:
    MSSpwParse thisParser(spwSubTable, ddSubTable, colAsTEN);
    return baseMSSpwGramParseCommand(&thisParser, command, selectedIDs, selectedChans, selectedDDIDs);

gervandiepen:
Has the swap of the semicolon and comma in the grammar been abandoned?

ms/MSSel/MSSpwGram.cc
sanbw: Agreed.  And checked-in the suggested modifications.

sanbw:
The swap of semicolon and comma has not been abandoned but will need to be staged since the current syntax is used in a number of places in pipelines.  Another idea that I am going to test is to insert a pre-parser which would detect and convert current syntax to the new proposed syntax.  This needs some more testing before the related code can be checked-in.

sanbw:
Thanks.

PR #583, opened 2017-03-27
Make getVersion() callable from plain c

tammojan:
It is useful to have the `getVersion()` method accessible from plain c. For that, I changed the signature of `getVersion()` to return a `char*` instead of a string. 

Also, the number returned by `getVersion()` is now the actual version, not `trunk` as it has been for some years.

I think the consequence of the API-breakage of this change is minimal, since `getVersion` is likely not used (since it hasn't done anything useful). Also, CASA uses `getVersionCASA`.

This change is necessary for casacore/python-casacore#79


gervandiepen:
Why has casacore_assay been changed?

tammojan:
> Why has casacore_assay been changed?

I changed casacore_assay to be able to ignore multiple fragments between `>>>` and `<<<` on one line. This was necessary because the output of tFITSImage is all on one line, and I had to ignore both a date and a version in the test.

The downside of the change is that now the text in between `>>>` and `<<<` cannot contain `<` anymore (to work around the too greedy matching of `.*`). Perhaps this can be worked around by changing the pattern `[^<]*` to something like `[^<]*<\{0,2\}[^<]*` (matches any number of non-`<` followed by 0 to 2 times `<`, followed again by any number of non-`<`). Ideally there would be another `*` behind this, I couldn't get this to work.

Alternatively, I could preprocess the FITS output with e.g. `fold` to insert line breaks every 80th character.

PR #584, opened 2017-04-03
Use IGRF12 in measuresdata

tammojan:
This PR fixes some minor issues in #479 .


dpetry:
Hi,
thanks for taking care of this. So this is all cleared with Wim?

tammojan:
This change has been cleared with Wim. The change to IGRF 12 has been checked against the fortran program that comes with the new coefficients as well. 

The slight change in the test output us also cleared with Wim. 

Wim does suggest to do more testing of the EarthMagnetic code, but that is in principle not related to the update to IGRF12.

PR #588, opened 2017-04-28
Interpolate2d tests, improvement, fixes #587

### LINKED ISSUE #587: Interpolate2D::stringToMethod poorly recognises the lanczos interpolation method
The stringToMethod function in Interpolate2D (scimath/Mathematics/Interpolate2D.cc) does an awkward job of choosing the Lanczos method - the method provided needs to start with a 'Z' (or 'z'), so that providing "lanczos" as the method name actually results in the "linear" method. And this is not documented anywhere.

I have attached a patch (in a zip file - 
[patchfile.zip] https://github.com/casacore/casacore/files/958424/patchfile.zip)
 that allows the method to be "lanczos" and it will return Interpolate2D::LANCZOS - this patch is a minimal fix that preserves the current behaviour for almost all situations, and keeps with the "minimal match" approach).

It could be argued that just doing a string match would be preferable here, which would require a more substantial change. Updating the documentation would be useful as well.

This was found in the casacore-2.0.3 release that is currently used as a 3rdParty package in the ASKAP SDP code tree (see most recent release 0.18.1 at http://doi.org/10.4225/08/5900991d2ae0a), but it appears that the code on the casacore trunk is the same.

tammojan:
Thanks to @mattwhiting for suggesting the fix in `scimath/Mathematics/Interpolate2D.cc`

### DIFF
--- a/scimath/Mathematics/Interpolate2D.cc
+++ b/scimath/Mathematics/Interpolate2D.cc
@@ -278,7 +278,12 @@ Interpolate2D::Method Interpolate2D::stringToMethod (const String &method) {
   if (tmp==String("N")) {
     method2 = Interpolate2D::NEAREST;
   } else if (tmp==String("L")) {
-    method2 = Interpolate2D::LINEAR;
+    String tmp2 = String(typeU.at(1, 1));
+    if (tmp2==String("A")) {
+      method2 = Interpolate2D::LANCZOS;
+    } else {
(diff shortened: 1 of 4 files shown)


gervandiepen:
Very good.

One small remark: in general it's better to catch an exception like (const AipsError& x) to avoid that a copy of the exception is made (which may fail if memory exhaustion is the culprit). But that is no problem here, so I'll merge the code.

PR #590, opened 2017-05-01
Handle FITS-IDI that uses the WEIGHT random parameter

kettenis:
FITS-IDI generated by the DiFX software correlator has weights encoded in
the WEIGHT random parameter instead of encoding it along the COMPLEX axis.
This replaces the no-working/unused code that looked for a WEIGHT keyword
which doesn't exist in FITS-IDI.

Tested with both GMVA data (DiFX data that uses the WEIGHT random parameter)
and EVN data (SFXC data that stores weights along the COMPLEX axis).

### DIFF
--- a/msfits/MSFits/FitsIDItoMS.cc
+++ b/msfits/MSFits/FitsIDItoMS.cc
@@ -851,9 +851,10 @@ void FITSIDItoMS1::describeColumns()
         kwl.first();
         uInt ctr=0;
 
-	weightKwPresent_p = False;
 	weightypKwPresent_p = False;
 	weightyp_p = "";
+	nStokes_p = 1;
+	nBand_p = 1;
 
         while((kw = kwl.next())){
 	    kwname = kw->name();
@@ -861,6 +862,14 @@ void FITSIDItoMS1::describeColumns()
 		maxis.resize(++ctr,True);
 		maxis(ctr-1)=kw->asInt();
 //		cout << "**maxis=" << maxis << endl;
+	    }
+	    else if(kwname.at(0,7)=="NO_STKD"){
+	        nStokes_p = kw->asInt();
+//		cout << "**nStokes=" << nStokes_p << endl;
+	    }
+	    else if(kwname.at(0,7)=="NO_BAND"){
+	        nBand_p = kw->asInt();
+//		cout << "**nBand=" << nBand_p << endl;
 	    }
 	    else if(kwname.at(0,8)=="WEIGHTYP"){
 	        weightypKwPresent_p = True;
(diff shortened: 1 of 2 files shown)


kettenis:
Not sure why this pull request says there are 15 commits.  I fear I somehow botched a merge somewhere...

tammojan:
If you're in tomorrow I'll drop by to fix the git stuff.

kettenis:
Thanks! I will be.

tammojan:
The following should work:
```
git reset 416d0ddc3 # The point where your unintentionally started diverging
git status # Check that only FitsIDItoMS.{cc.h} are modified
git diff # See what's changed
# Fix introduced whitespace errors (spaces at end of line, they are highlighted in the diff)
git add msfits/MSFits/FitsIDItoMS.cc msfits/MSFits/FitsIDItoMS.h
git commit -m "Handle FITS-IDI that uses the WEIGHT random parameter"
```
If you want, you can check again that you have only one commit on top of 416d0ddc3 by checking
```
git log --oneline -n3
```
If this looks ok, push the change to your github: 
```
git push https://github.com/kettenis/casacore.git master --force
```
(force because you changed history). The pull request will be updated automatically.
As a pro-tip for next time: use a branch on your fork (don't work on `master`). This will allow you to work on more than one feature at a time (if you want).

msfits/MSFits/FitsIDItoMS.cc
```diff
@@ -2062,8 +2059,10 @@ void FITSIDItoMS1::fillMSMainTable(const String& MSFileName, Int& nField, Int& n
 //	    cout<<"    visReal="<< visReal << "visImag="<< visImag<<endl;
 //	  }
 
-	  if(uv_data_hasWeights_p){
-	      memcpy(&visWeight, (static_cast<Float *>(data_addr[iFlux])) + count++, sizeof(Float));     
+	  if (uv_data_hasWeights_p) {
+	    memcpy(&visWeight, (static_cast<Float *>(data_addr[iFlux])) + count++, sizeof(Float));
+	  } else {
```
tammojan: The code in this `else` branch was not there before this change, i.e. if `uv_data_hasWeights_p==False` (data has no weights?). This affects the `count`. Does the code still work for data without weights?

kettenis:
Good catch. The FITS-IDI standard actually doesn't allow data without weights. Either the weights have to be provided in the data cube (`uv_data_hasWeights==True`) or the `WEIGHT` random parameter has to be present. But if `WEIGHT` is absent `getIndex()` will return -1, and using that as an array index isn't a good idea. Update version basically accepts data without weights and write 1 into the appropriate column of MS just like the old code did.

tammojan:
Pong. Thanks.

PR #592, opened 2017-05-03
CAS-8050 bugfix for MSRange::range "rows"

pford:
Restored line of code from comment for field MSS::ROWS


mpokorny:
Nice catch...that was a quite long-standing error.

PR #594, opened 2017-05-16
Cas 10143: Fix for inefficient access of FIELD sub-table columns

sanbw:
The bug this fixes adversely affected the time for field-selection with MSes with a large FIELD sub-table (100s of rows).

### DIFF
--- a/ms/MSSel/MSFieldIndex.cc
+++ b/ms/MSSel/MSFieldIndex.cc
@@ -93,17 +93,16 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
 	throw(MSSelectionFieldParseError(Mesg.str().c_str()));
       }
     //    cerr << "Pattern = " << strippedPattern << "  Regex = " << reg.regexp() << endl;
-    IPosition sh(msFieldCols_p.name().getColumn().shape());
+    Vector<String> names=msFieldCols_p.name().getColumn();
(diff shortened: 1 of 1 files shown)


ms/MSSel/MSFieldIndex.cc
```diff
@@ -132,14 +131,15 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
     else       reg=reg.fromPattern(pattern);
     
     //    cerr << "Pattern = " << pattern << "  Regex = " << reg.regexp() << endl;
-    IPosition sh(msFieldCols_p.name().getColumn().shape());
+    Vector<Bool> flagRow=msFieldCols_p.flagRow().getColumn();
+    Vector<String> codes=msFieldCols_p.code().getColumn();
+    IPosition sh(codes.shape());
```
mpokorny: Does codes.shape() here have the same value that msFieldCols_p.name().getColumn().shape() had in line 135 of the old code? Actually, your code is likely better given how sh is used later, but I thought I should ask.

sanbw: CODE and NAME are two columns of the FIELD sub-table and each NAME will have CODE cell associated with it.  So codes.shape() will be same as msFieldCols_p.name().getColumn().shape().

I made this change since, as you also note, it's the shape of the CODE column that really matters in the code later.  Also, in the earlier code the NAME column would be read unnecessarily to just get the shape.

PR #597, opened 2017-05-17
Cas 10151 disable caching of MS main table columns

dmehring:
because the cache can be swamped by columns for large MSes, meaning that smaller data structures, which are more computationally expensive to create, aren't cached. Also, the column data is usually only needed temporarily to compute smaller data structures, and the column data is not particularly expensive to recreate if necessary.

### DIFF
--- a/ms/MSOper/MSMetaData.cc
+++ b/ms/MSOper/MSMetaData.cc
@@ -67,9 +67,7 @@ MSMetaData::MSMetaData(const MeasurementSet *const &ms, const Float maxCacheSize
       _uniqueIntents(),
       _uniqueFieldIDs(), _uniqueStateIDs(),
       _avgSpw(), _tdmSpw(),
-      _fdmSpw(), _wvrSpw(), _sqldSpw(), _antenna1(), _antenna2(),
-      _scans(), _fieldIDs(), _stateIDs(), _dataDescIDs(),
-      _observationIDs(),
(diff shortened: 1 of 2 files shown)


mpokorny:
David, I assume that you have a test case that shows a performance improvement with your changes. I'm wondering, however, what effects these changes would have more generally. Your reasoning seems sound, but I don't want to try and predict the performance effects in general; do you have extensive performance test results? Does anyone else with more experience care to chime in?

dmehring:
Hi Martin,
This work is the result of issues uncovered in CAS-9098. Essentially, a smaller data structure could not be cached becasue column data was taking up most of the cache. The smaller structure was very expensive to construct, and the method called was being run in a loop so that the smaller data structure had to be computed many times. Retrieving of column data is comparably fast. And ultimately, I'd like to remove retrieving the entire column vector in place of using TableIterators which will signficantly reduce memory usage for large MSes. That is CAS-10150. I implemented the vast majority of MSMetaData, so I'm not sure if you will be able to find someone with more experience with it.

mpokorny:
I had a look at all the uses of MSMetaData in CASA aside from StatWt, and I don't see any cases where performance would clearly be harmed by your changes. I like your idea of replacing the retrieval of entire column vectors with TableIterators eventually. I'll proceed with the merge.

dmehring:
Thanks Martin!

PR #602, opened 2017-06-15
Implement digital corrections for DiFX/VLBA

kettenis:
Data correlated with the DiFX correlator needs to have quantization
corrections applied.  Since these corrections are different for auto
correlations and cross correlations it is easiest to apply them while
importing the data.  This implementation matches what the AIPS FITLD task
does and what is described in VLBA Correlator Memo 75 and VLBA Scientific
Memo 12.

### DIFF
--- a/msfits/MSFits/FitsIDItoMS.cc
+++ b/msfits/MSFits/FitsIDItoMS.cc
@@ -103,6 +103,8 @@
 
 #include <casacore/casa/iomanip.h>
 
+#include <casacore/scimath/Mathematics/FFTPack.h>
+
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
 //local debug switch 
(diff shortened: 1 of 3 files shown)


dpetry:
Hi Mark,
glancing over your changes, I couldn't immediately see a comment explaining exactly what are the valid values of the new parameter "correlat" in FITSIDItoMS1::FITSIDItoMS1. I think it would be good to enumerate them somewhere.

dpetry:
OK, I don't want to hold this up for cosmetic reasons. But in a second commit, you should
tidy up the valid values of correlat and ideally introduce an enumeration for them and clarify
what other valid values there are besides "DIFX" and whether they are case sensitive.

kettenis:
Thanks; I'll create a new pull request that adjusts the documentation.

PR #603, opened 2017-06-19
Cas 10150 reimplemenet _computeScanAndSubScanProperties()

dmehring:
Reimplemented _computeScanAndSubScanProperties() to use significantly less memory for large data sets. Associated refactoring, including moving the merge of properties to its own private method.

### DIFF
--- a/ms/MSOper/MSMetaData.cc
+++ b/ms/MSOper/MSMetaData.cc
@@ -35,12 +35,15 @@
 #include <casacore/measures/TableMeasures/ArrayQuantColumn.h>
 #include <casacore/ms/MSOper/MSKeys.h>
 #include <casacore/ms/MeasurementSets/MSFieldColumns.h>
+#include <casacore/ms/MeasurementSets/MSIter.h>
 #include <casacore/ms/MeasurementSets/MSSpWindowColumns.h>
 #include <casacore/scimath/Mathematics/ClassicalStatistics.h>
 #include <casacore/tables/Tables/ArrayColumn.h>
 #include <casacore/tables/Tables/ScalarColumn.h>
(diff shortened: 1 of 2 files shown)


mpokorny:
It looks like you've reduced the memory usage by accumulating the "properties" over subsets of rows at a time rather than over all the rows at once; is that about right? I see no issues with the changes, but I'd like to understand their purpose.

dmehring:
Yes. The issue being that it is becoming more common for MSes to have several of tens of millions or even hundreds of millions of rows. So, holding several entire columns in memory simultaneously can be very taxing. The somewhat arbitrary limit I've chosen of 10M row blocks seems to put a reasonable cap on memory usage with minimum impact on performance. I may modify it in the future, possibly even letting the client specify it (directly or indirectly) as problematic use cases are identified, but for now, it mitigates the most egregious issues with very large MSes.

mpokorny:
A couple of suggestions regarding your commits: you might want individual commit messages to better describe your changes, keeping in mind the 50-70 character limit on the first line. The line limit is a common git convention that I believe casacore would like to maintain. Depending on your level of confidence with git, you may want to consider performing an interactive rebase on your branch before making the pull request to clean up the commit history, as well.

dmehring:
Thanks. I don't know much about git rebase. Something to investigate. When Julian was reviewing my changes, he asked me to perform multiple commits on a single branch if the changes were extensive, so he could review each separately.

mpokorny:
Keeping individual commits relatively small and self-contained is generally preferred. An interactive rebase can be useful to tidy up your commits before making a pull request. However, you don't want to go too far and lump all commits into a single one when that defeats the goal stated in my first sentence.

PR #606, opened 2017-06-23
Modernised matrix syntax in FITS files now following standard 3.0:

dpetry:
Format is now XXi_j. No leading zeros in indices.
This is part 1 of 2 for a fix to CAS-9993 which reports general problems with the PC matrix
in cases where the matrix is non-diagonal.


fits/FITS/FITSKeywordUtil.cc
```diff
@@ -335,15 +335,20 @@ Bool FITSKeywordUtil::addKeywords(FitsKeywordList &out,
...
 			Int nrow = in.shape(i)(0);
  			Int ii = k % nrow + 1;
  			Int jj = k / nrow + 1;
- 			ostringstream ostr;
- 			ostr << setfill('0') << setw(2) << ii
-			     << "_"
-			     << setfill('0') << setw(2) << jj;
- 			name += String(ostr);
+			ostringstream ostr;
+			if(nrow>9){
```
tammojan: Does this magic number 9 mean something? If so, could you put a comment here?

dpetry: It's simply the transition from a one-digit to a two-digit decimal integer.

dpetry: I mean the integer for the indices of the matrix.

dpetry:
Any objections to me merging this pull request?

mpokorny:
I'm no FITS expert, but the changes look straightforward enough, and seem OK to me. As for magic number, I usually prefer not to have them; in this case, it's fairly clear what the meaning is, but  a comment surely wouldn't hurt for the benefit of the next person to look at the code.

dpetry:
OK, will do in the second instalment of my work on CAS-9993 . Promise.

PR #608, opened 2017-06-26
Properly initialize weights when importing FITS-IDI

kettenis:
Initialize the WEIGHT/WEIGHT_SPECTRUM and SIGMA/SIGMA_SPECTRUM columns
in a way that is compatible with the conventions used by CASA 4.4 an later.
This uses the correlator weights to calculate the effective integration time
that is used in the caclulation of the expected per-channel noise.
This should make is possible to combine data with different integration
times and channel widths.

### DIFF
--- a/msfits/MSFits/FitsIDItoMS.cc
+++ b/msfits/MSFits/FitsIDItoMS.cc
@@ -29,6 +29,7 @@
 #include <casacore/casa/Arrays/ArrayIO.h> 
 #include <casacore/casa/Arrays/ArrayLogical.h>
 #include <casacore/casa/Arrays/ArrayMath.h>
+#include <casacore/casa/Arrays/ArrayPartMath.h>
 #include <casacore/casa/Arrays/ArrayUtil.h>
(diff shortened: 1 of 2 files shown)


msfits/MSFits/FitsIDItoMS.cc
```diff
@@ -2208,6 +2210,40 @@ void FITSIDItoMS1::fillMSMainTable(const String& MSFileName, Int& nField, Int& n
       if (weightyp_p == "CORRELAT")
 	vis /= weightSpec;
 
+      // determine the spectralWindowId
+      Int spW = ifno;
+      if (iFreq>=0) {
+        memcpy(&spW, (static_cast<Int *>(data_addr[iFreq])), sizeof(Int));
+        spW *= (Int)tscal(iFreq);
+        spW += (Int)tzero(iFreq); 
+ 	spW--; // make 0-based
```
tammojan: Please avoid using both tab and space indentation within the same block.

kettenis:
Fixed the tab vs. spaces issue.

PR #609, opened 2017-06-30
Fix transpose problem of FITS PC matrix found in CAS-9993

dpetry:
This is part 2 of 2 of the fix for CAS-9993 / CAS-10342 (the latter ticket needed to be filed for technical reasons as a continuation of CAS-9993). The problem solved here is that the PC matrix written by FITSKeywordUtil.cc is transposed w.r.t. its proper value. However, FITSKeywordUtil.cc is self-consistent in its read/write behaviour. The problem is therefore solved at the transition from the image representation of the linear image coordinates transformation to the PC matrix, i.e. in FITSCoordinateUtil.cc where the pc matrix is retrieved from the CASA image.
Comprehensive tests were performed to verify that the change (a) solves the problem within the CASA framework and (b) does not damage anything else in casacore.
As promised in pull 606, I also added the comment explaining the magic number 9 in FITSCoordinateUtil.cc .

### DIFF
--- a/coordinates/Coordinates/FITSCoordinateUtil.cc
+++ b/coordinates/Coordinates/FITSCoordinateUtil.cc
@@ -485,7 +485,14 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
 	    while (cunit(i).length() < 8) cunit(i) += " ";
 	}
 //
-	pc = cSys.linearTransform();
+	Matrix<Double> imageLT = cSys.linearTransform();
+	pc = imageLT;
+	// need to transpose to conform with FITSKeywordUtil
+	for(uInt i=0; i<imageLT.nrow(); i++){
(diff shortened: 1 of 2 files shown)


tammojan:
Ideally this would be coupled with a test that fails with the old behavior and passes with the new behavior.

Also, since this changes the information in the FITS file, should it be documented in the FITS header which version of casacore was used?

dpetry:
Thanks for your review. This is a bug fix, not a format change or new feature. So nominally, all formats have stayed the same. When CASA writes FITS files, the version is placed into the header (HISTORY). We could think about also placing the casacore version there. I need to think about where to place the code for that. Better in a different pull request so this fix is not held up.

I agree that a new test will be good but since this feature is across two modules, I again need to think a bit about how and where to place the test.

PR #610, opened 2017-07-04
Fix tStatisticsUtilities, tLatticeStatistics and tLCEllipsoid tests for non-Intel architectures

olebole:
This is a patch I got for the [Debian bug #838424](https://bugs.debian.org/838424#15) from @egrimley. The following is copied from his bug comments:

The tests `tStatisticsUtilities` and `tLatticeStatistics` can be made to pass on arm64 with these adjustments to the expected accuracy.

The test `tLCEllipsoid` seems to be converting ellipses into bitmaps. I'm guessing it gives different output on amd64 and arm64 because the real curve passes through a critical point and numerical inaccuracy puts it on different sides of the point on the two architectures. One way to fix that might be to use non-round numbers in the parameters so that the curve does not pass through a critical point. That's what I tried to do, and with the following changes to the program and the expected output the test passes on both amd64 and arm64.

I'm not sure what to do about the other two tests that fail on arm64, `tLSQaips` and `tLSQFit`, because those two generate a load of output text containing floating-point numbers and compare that text with an expected output. It might not be so easy to make that work on all architectures. Perhaps one could simply disable those two tests. Preferably one would still run them, so that we can see what happens in the logs, but not have the build fail if only those tests fail.


tammojan:
> I'm not sure what to do about the other two tests that fail on arm64, tLSQaips and tLSQFit, because those two generate a load of output text containing floating-point numbers and compare that text with an expected output. It might not be so easy to make that work on all architectures. Perhaps one could simply disable those two tests. Preferably one would still run them, so that we can see what happens in the logs, but not have the build fail if only those tests fail.

Casacore's testing routines already ignore differences up to 10^-5, see http://casacore.github.io/casacore-notes/262.html  . I'm not sure why tLSQaips and tLSQFit still fail.

tammojan:
Thanks!

olebole:
> I'm not sure why tLSQaips and tLSQFit still fail.

I just double checked: On arm64, both still [fail](https://buildd.debian.org/status/fetch.php?pkg=casacore&arch=arm64&ver=2.3.0-1&stamp=1498248374&raw=0) on 2.3.0.

PR #615, opened 2017-07-24
Complete copy assignment for MSIter

mpokorny:
Complete implementation of MSIter copy assignment operator and new clone() method, with supporting changes to TableIterator and BaseTableIterator.

### DIFF
--- a/ms/MeasurementSets/MSIter.cc
+++ b/ms/MeasurementSets/MSIter.cc
@@ -323,6 +323,7 @@ MSIter::operator=(const MSIter& other)
   tabIter_p.resize(nMS_p);
   for (Int i = 0; i < nMS_p; ++i) {
     tabIter_p[i] = new TableIterator(*(other.tabIter_p[i]));
+    tabIter_p[i]->copyState(*other.tabIter_p[i]);
   }
   tabIterAtStart_p = other.tabIterAtStart_p;
   curMS_p = other.curMS_p;
--- a/ms/MeasurementSets/MSIter.cc
+++ b/ms/MeasurementSets/MSIter.cc
@@ -379,6 +379,10 @@ MSIter::operator=(const MSIter& other)
   return *this;
 }
 
+MSIter *
+MSIter::clone() const {
+	return new MSIter(*this);
+}
 
 const MS& MSIter::ms(const uInt id) const {
 
--- a/ms/MeasurementSets/MSIter.cc
+++ b/ms/MeasurementSets/MSIter.cc
@@ -83,19 +83,18 @@ int MSInterval::comp(const void * obj1, const void * obj2) const
 }
 
 
-MSIter::MSIter():nMS_p(0),msc_p(0),storeSorted_p(False),allBeamOffsetsZero_p(True),
-  timeComp_p(0) {}
+MSIter::MSIter():nMS_p(0),storeSorted_p(False),allBeamOffsetsZero_p(True)
+{}
 
 MSIter::MSIter(const MeasurementSet& ms,
 	       const Block<Int>& sortColumns,
 	       Double timeInterval,
 	       Bool addDefaultSortColumns,
 	       Bool storeSorted)
-: msc_p(0),curMS_p(0),lastMS_p(-1),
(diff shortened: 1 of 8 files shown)


gervandiepen:
Doesn't the setting of msc_p in setMSInfo without a previous delete result in memory leaks?
setMSInfo is called from various places, so msc_p can be overwritten several times without deleting its previous contents.
You could run it with valgrind to check it.

I don't know if I agree with the changes in TableIter and BaseTabIter. They change the behaviour. The clone was always meant as a way to get a new clean iterator, but now it is changed to making a copy with the same state. There might be code using it depending on the old behaviour.
Do you have a specific reason why you want the new behaviour?
NB. You do not copy the contents of lastVal_p and curVal_p in the BaseTabIter::clone method.

mpokorny:
CountedPtr<ROMSColumns> MSIter::msc_p; so an assignment results in an assignment to CountedPtr::pointerRep_p, which is a shared_ptr, and the current instance in that shared_ptr will be deleted before the new value is assigned. That said, I will run a test with valgrind to make sure.

As for TableIter and BaseTabIter, I agree that there is a change in behavior, and that is one of my concerns about this set of changes. I was hoping that the test suite, while not definitive, might uncover potential issues, but all tests pass. There is a specific reason for the new behavior, namely for some development I'm doing in CASA that requires a copyable MSIter that maintains state. I do think that a true copy/clone is generally better for copy assignments and such rather than something less complete, as that would agree with typical expectations. One alternative that might work would be to create new methods for the changed behavior and use them explicitly when copying an MSIter.

Regarding lastVal_p and curVal_p, they are copied in the copy constructor of BaseTableIterator called in BaseTableIterator::clone().

mpokorny:
I just completed a valgrind check of tMSIter, and it reports no memory leaks.

gervandiepen:
- I missed the change of msc_p to a CountedPtr. That's a nice change. Great that valgrind reported no errors.
- Are you sure curVal and lastVal are copied? It is not in the current implementation (it only allocates them) and I did not see it added in TableIter/BaseTabIter.cc.
- Maybe you can add a Bool reset parameter to clone with default value False to keep the old behaviour while supporting the new.

mpokorny:
A parameter to BaseTableIterator::clone() seems like it would work. That change would nevertheless require changing the call to that method in the TableIterator copy assignment operator, so the semantics of that operator would still change. Would that be OK?

As for BaseTableIterator::curVal_p and lastVal_p, making those members private would negate any requirement to copy their values. What are your thoughts on that approach?

gervandiepen:
I was thinking if a copyState function could do the job.
In that way operator= does not need to be changed, but copyState could be called in clone.

Copying the state requires copying the contents of curVal_p and lastVal_p, which is a bit tricky. They are declared as a Block of void* (one per column). Per column the function allocIterBuf in ScaColData<T> allocates a value with the correct data type, which is deallocated by freeIterBuf. ScaColData<T>::get gets a value into the value pointed to by the void*.
Copying the value of curVal and lastVal means doing a type dependent copy, so either a function is added to ScaColData to do so, or you need to have a switch statement in BaseTabIter.

mpokorny:
I would prefer that the copy assignment operator to copy the full observable state of TableIterator, but I guess that could break existing code. Are you suggesting instead that after a call to operator= the caller then use copyState to complete the copying of observable state? I guess that could work...I'll play with it a bit.

My point about curVal_p and lastVal_p is that the only usage of those values (aside from allocation and deallocation) in BaseTableIterator is by the next() method, where those arrays' elements are filled by calls to colPtr_p[i]->get every time before the array element values are used. Aside from efficiency concerns, those variables could be local to the next() method. In that sense, their values are of no significance externally, and they could be made private without losing any functionality.

gervandiepen:
You are right about curval and lastval. Basically they could indeed be local. I forgot about my own code.

A copyState function makes it in principle possible that the state is copied from a TableIterator using different columns. I guess it is not harmful, but I don't know if it is desirable.
Maybe it can be made private to avoid it.

mpokorny:
travis build failed -- should build again once #621 is merged.

PR #616, opened 2017-07-25
Cas 10358 add clone() method, allow user to set max number of threads

dmehring:
Implemented clone() method to all stats classes, added tests
Implemented ClassicalStatistics::setMaxThreads() to allow caller to set max threads used.
Substituted macros in some places.
Made more eclipse-friendly by fully qualifying certain classes with namesspace (eg vector -> std::vector)

### DIFF
--- a/scimath/Mathematics/ChauvenetCriterionStatistics.h
+++ b/scimath/Mathematics/ChauvenetCriterionStatistics.h
@@ -64,6 +64,9 @@ public:
         const ChauvenetCriterionStatistics<CASA_STATP>& other
     );
 
+    // Clone this instance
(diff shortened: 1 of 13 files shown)


dmehring:
@mpokorny perhaps you know a better way. When calling from an already omp parallel loop, it seems the number of threads in the lower level ClassicalStatistics omp parallel loop was one, and not all data were being included, so the wrong value was being returned. I'm uncertain as to how to configure the top level omp parallel to deal with that properly.

dmehring:
Actually, I think I found a better way. I'll create another branch for that.

PR #618, opened 2017-08-01
Use antenna name as station name when provided

kettenis:
The current approach of using ARRAYNAME:NUMBER as the station name breaks
concatenating imported FITS-IDI files that have different sets of antennas.
This results in differences in station names for the same antenna which
the concatenation code interprets as moving the antenna.

### DIFF
--- a/msfits/MSFits/FitsIDItoMS.cc
+++ b/msfits/MSFits/FitsIDItoMS.cc
@@ -2547,17 +2547,17 @@ void FITSIDItoMS1::fillAntennaTable()
      }
      if(name(i)==""){
        ant.name().put(row,String("ANT")+temps);
+       ant.station().put(row,arrnam+":"+temps);     
      }
      else{
        ant.name().put(row, name(i));
+       ant.station().put(row,name(i));
      }
      Vector<Float> tempf=offset(i);
      Vector<Double> tempd(3);
      for (Int j=0; j<3; j++) tempd[j]=tempf[j];
      ant.offset().put(row,tempd);
 
-     //ant.station().put(row,name(i));
-     ant.station().put(row,arrnam+":"+temps);     
      ant.type().put(row,"GROUND-BASED");
 
      // Do UVFITS-dependent position corrections:


dpetry:
OK, I wonder if one should precede the station name by some fixed string like "STATION_" in order to make it different from the antenna name. E.g. if the antenna name is "A1", then the station name would be "STATION_A1". What do you think?

kettenis:
I'm not sure what the benefit there would be. Is there a problem with having NAME and STATION the same?  There are VLBI stations that have multiple antennas, i.e. the Lovell and Mark2 at Jodrell and the 20- and 25-meter telescopes at Onsala. But that information isn't recorded in the FITS-IDI file.  Most VLBI users will probably prefer to refer to stations with their two-letter codes. I assume the same is true for e-MERLIN. Are there any other users of importfitsidi that you know of?

dpetry:
Sorry, was on vacation: My point is that by looking at the station and antenna name, the user will not be able to tell which is which is we don't make them different.
Users then might suspect a mistake somewhere.

dpetry:
OK, Mark is on vacation. I don't want to let this apparently not working code uncorrected.
Mark, let's discuss my suggestion when you are back. You can put it into a new pull request.

trhunter:
I agree with Dirk.  At present, for an SMA dataset loaded into CASA, you can tell both the antenna number and the pad number.  Having the pad number is useful for being able to simulate the data, though admittedly the pad location is also present.  It also convenient for being able to quickly know which antennas may be observing the mountainside of Pu'u Poliahu at low elevation.  Having the antenna number is important to identify which antennas may be misbehaving in various ways.

Antennas: 6:
  ID   Name  Station   Diam.    Long.         Lat.                Offset from array center (m)                ITRF Geocentric coordinates (m)
                                                                     East         North     Elevation               x               y               z
  0    SMA1  Pad 7     6.0  m   -155.28.39.8  +19.49.26.6        -20.0763      -15.7360       -0.1152 -5462462.038661 -2491953.821187  2164431.262884
  1    SMA2  Pad 1     6.0  m   -155.28.39.1  +19.49.27.1         -0.0000        0.0001        0.0000 -5462448.949395 -2491969.916598  2164446.105480
  3    SMA4  Pad 8     6.0  m   -155.28.41.3  +19.49.26.4        -63.8744      -22.0490       -3.2268 -5462479.501485 -2491913.647272  2164424.268770
  5    SMA6  Pad 4     6.0  m   -155.28.39.9  +19.49.27.2        -25.1524        1.3776       -0.0437 -5462458.926432 -2491946.822082  2164447.386556
  6    SMA7  Pad 5     6.0  m   -155.28.39.7  +19.49.27.7        -18.9854       16.6201       -0.0695 -5462451.641738 -2491950.277158  2164461.717016
  7    SMA8  Pad 23    6.0  m   -155.28.41.1  +19.49.28.3        -59.5573       34.3634       -6.6553 -5462457.369458 -2491908.295932  2164476.175219

kettenis:
Note that the FITS-IDI files only contain a single (antenna) name. So importfitsidi can't actually fill in the pad numbers unless we modify the FITS-IDI standard is modified. Judging from the smaImportFix.py script, the pad numbers are put in after importfitsidi runs.  They are simply overwritten, so I expect it doesn't actually matter what importfitsidi puts into the STATION column.

dpetry:
For SMA data processed with special scripts it might not matter but generally, it would be good for the user to be able to tell the difference between a station name and an antenna name.
That's why I wanted to propose that if importfitsidi has to make up station names,
it does not simply give them the same name as the antenna on it, but precedes the name
by the word "STATION" or similar. It's just a small thing but might make output of downstream
software more clear.

PR #620, opened 2017-08-02
Take care of all JSON escape chars

gervandiepen:
CAS-10459 reported an error about an unescaped quote in the ImageConcat Json file.

The Json code contained logic to escape quotes, etc., but accidently it was not used.
This has been fixed. Furthermore, the escape logic has been extended to handle control chars.
The tests have been extended accordingly.


casa/Json/JsonParser.cc
```diff
@@ -125,8 +125,48 @@ namespace casacore {
...
+            break;
+          case 'u':
+            {
+              // unicode repr of control character
+              Bool ok = False;
+              if (i < leng+4) {
+                String hex("0X" + in.substr(i+1,4));
+                char* endPtr;
+                Int64 val = strtoll(hex.chars(), &endPtr, 0);
+                if (endPtr == hex.chars()+hex.size()  ||  val < 128) {
```
mpokorny: If hex has the value "0X12hi", the first part of this conditional is false, but val has value 12, and so the second part of the conditional is true. Shouldn't the conditions be combined with && instead?

gervandiepen: You are entirely right. I'll change it. Thanks.

PR #621, opened 2017-08-07
Pin travis Linux OS at "precise"

mpokorny:
The default travis Linux OS has become Ubuntu "trusty". To continue using Ubuntu "precise" on travis, this configuration value is required. Note that a build of the branch for #615 was failing until I tried out this change (which has been reverted on that branch). Long term, I suggest we move to "trusty", in which case pinning the version at "precise" may be considered a workaround.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: cpp
 cache: ccache
+dist: precise
 
 os:
   - linux


gijzelaerr:
Thanks! yes, this is a temporary workaround until we get the build working with trusty.

mpokorny:
Created #622 to track the update to build under trusty.

PR #623, opened 2017-08-18
CAS-10543 Add workaround for SMA SYSTEM_TEMPERATURE table

kettenis:
FITS-IDI files from the SMA have scalar TSYS_1/2 columns instead of
array columns as required by the FITS-IDI standard.  Interpret these
as a single value that applies to all bands like we already do for
POLAA/B in the ANTENNA table.

### DIFF
--- a/msfits/MSFits/FitsIDItoMS.cc
+++ b/msfits/MSFits/FitsIDItoMS.cc
@@ -3138,16 +3138,32 @@ Bool FITSIDItoMS1::fillSysCalTable()
 
   Int nVal=nrows();
   Bool dualPol=False;
+  Bool TSYSisScalar=False;
 
   Table tyTab = oldfullTable("");
   ROScalarColumn<Double> time(tyTab, "TIME");
   ROScalarColumn<Float> timeint(tyTab, "TIME_INTERVAL");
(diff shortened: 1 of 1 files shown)


dpetry:
Hi Mark,
could you modify the WARNING text to say explicitly that the input FITSIDI file is not following the standard. E.g. TSYS_1 and TSYS_2 columns in input SYSTEM_TEMPERATURE table are scalar although FITSIDI standard requires array."
+	    << endl << "Will accommodate that by using same value for all bands." << LogIO::POST;

kettenis:
Hi Dirk,
Turns out that if there is only a single band the column ends up being scalar anyway, even if it is specified as 1E instead of E. So I probably should drop the warning change as it appears to be incorrect.

dpetry:
Just make sure, when you have a working fix, that it also works with "good" FITSIDI files.

kettenis:
Yes, verified that it still works for VLBA and EVN data. Also verified that the test_imporfitsidi.py test in CASA still passes.

dpetry:
So, should this be merged after all?

kettenis:
Yes, please. Without it loading of single-band FITS-IDI files that conform to the standard is broken too.

PR #625, opened 2017-08-23
Restore deleted line from TableIter, fixes #624

### LINKED ISSUE #624: TableIterator::operator= does not initialize tabIterPtr_p properly
The following change was made to TableIter.cc in commit #615 :

    @@ -101,8 +101,7 @@ TableIterator& TableIterator::operator= (const TableIterator& iter)
         tabIterPtr_p = 0;
         subTable_p   = Table();
         if (iter.tabIterPtr_p != 0) {
    -       tabIterPtr_p = iter.tabIterPtr_p->clone();
    -       next();                            // get first subtable
    +        tabIterPtr_p = iter.tabIterPtr_p->clone();
         }
     return *this;
     }

In particular, at TableIter.cc:105, `next()` was removed, I think by @mpokorny . This causes the `tabIterPtr_p` of the assigned-to object not to be properly initialized.

As I mentioned earlier today to @tammojan and @gervandiepen, this makes DPPP crash, since DPPP does this:

      itsIter = TableIterator (sortms, Block<String>(1, "TIME"), 
         TableIterator::Ascending, TableIterator::NoSort);
      Table sortab = itsIter.table().sort(sortCols, 
         Sort::Ascending, Sort::QuickSort + Sort::NoDuplicates);

Which I think is valid, and used to work before commit #106, but now results in:

    terminate called after throwing an instance of 'casa::AipsError'
      what():  (/home/anoko/Software/git-casacore/tables/Tables/BaseTable.cc : 603) Failed AlwaysAssert !isNull()

When I reinsert the "next()" statement after the assignment to tabIterPtr_p, it seems to work.

tammojan:


tammojan:
The discussion in the comments of #615 indeed mention a change to the behavior regarding keeping state or not. However, the discussion does not come to a conclusion on whether state should be preserved in the assignment operator.

The original behavior was to not copy the state. In the current master, the assignment operator creates a TableIterator with a reference to a NullTable, which makes things crash. So I suggest to revert to the original behavior by merging this PR.

mpokorny:
Sorry about the late response, but I was on vacation when this came up. FWIW, I agree with the fix. The TableIterator::copyState() method does what's required for the tabIterPtr_p member, and the copy assignment operator can therefore still do what it's done previously. We can discuss whether the copy assignment operator _should_ copy state some other time. Thanks for taking care of this, @tammojan.

PR #626, opened 2017-08-24
Reduce the unnecessary FIELD table access and phase center calculatio…

kgolap:
…n at every iteration. Made it on demand phasecenter and some field meta data is read only when requested rather than at every iteration. Also fixed and issue that phasecenter was wrong for time dependent phasecenters (i.e phasecenter for ephemerides or being a time polynomial). Added a phasecenter function so that users who define the iterator to have more than 1 field or time stamp in a given iteration can call the correct phasecenter function.

The field table access at every iteration causes a lot of small read and in multiprocess mode sometimes is part of the cause of memory bandwidth access (even though the whole table may be cached).

### DIFF
--- a/ms/MeasurementSets/MSIter.cc
+++ b/ms/MeasurementSets/MSIter.cc
@@ -83,7 +83,7 @@ int MSInterval::comp(const void * obj1, const void * obj2) const
 }
 
 
-MSIter::MSIter():nMS_p(0),storeSorted_p(False),allBeamOffsetsZero_p(True)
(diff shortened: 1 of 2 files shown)


dpetry:
So, do you take the default time value to give to phaseDirMeas from the TIME column of the FIELD table? Reading the code, I was not sure this is happening.

kgolap:
That used to be the old behavior, now i have provided 2 API's  with 2 different behaviors one...it takes the first time stamp in the MSIter iteration....this is okay if the iterator is set to have only one time integration per iteration...but for cases the caller set the iterator to have multiple time stamps and multiple fields per iteration s/he can then call the second function with explicit fieldid and time.

PR #627, opened 2017-08-25
Parallactic angle calculation for Nasmyth mounts

kettenis:
This introduces new mount types "NASMYTH-LEFT" and "NASMYTH-RIGHT"
for left-handed and right-handed Nasmyth mounts found on EVN, GMVA
and EHT antennas.  This change also makes the FITS-IDI import code
recognize these mount types, which are described in the update
FITS-IDI standard (AIPS Memo 114).

### DIFF
--- a/ms/MSOper/MSDerivedValues.cc
+++ b/ms/MSOper/MSDerivedValues.cc
@@ -194,8 +194,12 @@ MSDerivedValues& MSDerivedValues::setAntennaMount(const Vector<String>& mount)
         mount_p(i)=2;
       } else if (mount(i)=="orbiting" || mount(i)=="ORBITING") {
         mount_p(i)=3;
+      } else if (mount(i)=="nasmyth-right" || mount(i)=="NASMYTH-RIGHT") {
+	mount_p(i)=4;
+      } else if (mount(i)=="nasmyth-left" || mount(i)=="NASMYTH-LEFT") {
+	mount_p(i)=5;
       } else {
-        mount_p(i)=4;
+        mount_p(i)=6;
       }
     }
   }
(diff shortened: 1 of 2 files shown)


kettenis:
The names for the new mount types should perhaps be discussed before this gets merged. I chose "NASMYTH-LEFT" and "NASMYTH-RIGHT" because they are reasonably short strings. But perhaps "ALT-AZ+NASMYTH-LEFT" and "ALT-AZ+NASMYTH-RIGHT" are more accurate and in line with the "ALT-AZ+ROTATOR" that's already supported.  These longer names could probably be abbreviated to "ALT-AZ+NASMYTH-L" and "ALT-AZ+NASMYTH-R".

tammojan:
I'm not sure about the status of "ALT-AZ+ROTATOR", that was put in as "a small hack to allow simulation of a perfect feed rotator: ALT-AZ+ROTATOR mount is treated the same way as ALT-AZ" (bca90939616003a4b2433587a2f43d69e171ee05). It currently does exactly the same as "ALT-AZ".

I do agree that a consistent naming scheme is to be preferred; your last suggestion makes most sense to me.

dpetry:
The "+" in the names is maybe a problem in some contexts.
How about ALTAZ_NASMYTHL, ALTAZ_NASMYTHR, ALTAZ_ROTATOR ?
This is as short as possible and avoids special characters.

dpetry:
Sorry, I misunderstood at first.
The ARRAY_GEOMETRY table already contains a MNTSTA column which describes the mount type and codes 4 and 5 are for right and left handed Nasmyth.
You are talking about what to put into the MS ANTENNA table column "MOUNT".
Here "ALT-AZ" is already in the original spec. So, in this case, I agree that the "-" between ALT and AZ should stay.
How about ALT-AZ_NASMYTHL, ALT-AZ_NASMYTHR, ALT-AZ_ROTATOR ?

dpetry:
OK, thinking further about it, since the "+" was already introduced in ALT-AZ+ROTATOR  
and seems to work,
your "ALT-AZ+NASMYTH-LEFT" and "ALT-AZ+NASMYTH-RIGHT" are good.

tammojan:
@dpetry If you agree with the changes, could you review / merge this?

One comment I would have is that there is no test for any part of the FITS-IDI importer; @kettenis already suggested he'd make make one some time ago.

PR #628, opened 2017-09-06
Tabular SpectralCoordinate::near bug

kgolap:
Fixed an exception for the near test when it is a TabularCoordinate with 1 value only (which happens commonly when doing subimage on a Tabular spectral cube for e.g)

### DIFF
--- a/coordinates/Coordinates/SpectralCoordinate.cc
+++ b/coordinates/Coordinates/SpectralCoordinate.cc
@@ -1293,10 +1293,11 @@ Bool SpectralCoordinate::near(const Coordinate& other,
          }
       }
    }
-
 // LinearXForm components
-
-   {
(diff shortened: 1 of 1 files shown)


kgolap:
Does anybody know what failed ...the log is not too clear...
I ran make test on my branch before pushing and i got all ok.

100% tests passed, 0 tests failed out of 525

Total Test time (real) =  96.08 sec

sanbw:
PR #629 also failed for the same reason:

> The job exceeded the maximum time limit for jobs, and has been terminated.

tammojan:
One of the subjobs timed out. the one without C++11. The one with C++11 succeeded.

There is a time limit of around 45 minutes per build for using the Travis infrastructure. With C++11, the build time is quite a lot shorter thanks to explicit template instantiation.

While it is an important issue to keep Travis green at all times, I think fixing Travis should not be in the way of this PR. I've created #630 to track the Travis timeout.

PR #631, opened 2017-09-06
Correct the header paths.

vsuorant:
Grimmer noticed that some of the header paths were incorrect. This went undetected by the Casa build system but caused trouble if Carta.


dpetry:
Looks like the checks for this also failed because of time-outs in Travis.
Tammo is saying that this should not hold up the merge.

PR #632, opened 2017-09-06
Remove travis Linux OS version pin

mpokorny:
Allow build under Ubuntu trusty.
I don't expect the travis build for this PR to succeed. As I'm unsure why the travis build under Ubuntu trusty is failing, and I don't have an Ubuntu trusty system to do my own testing, I am using this pull request for build testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: cpp
 cache: ccache
-dist: precise
 
 os:
   - linux


mpokorny:
I'm wondering whether the numpy packages for the various Python versions should be installed by pip(3) or conda. I read on the Travis CI site that Python projects should install Python packages using pip, since Travis uses virtualenv to manage different versions of Python under which the project should build. Various web sites suggest conda might be a better alternative. Thoughts?

tammojan:
The important bit is to install a binary version of numpy; compiling it from source will take up too much travis time. If possible, I'd prefer not to use Conda, since that introduces another layer of virtualization (on top of the Ubuntu package manager and pip).
Given that the Travis documentation [suggests](https://docs.travis-ci.com/user/languages/python/#Travis-CI-Uses-Isolated-virtualenvs) to use pip, I think `pip install numpy` is the appropriate solution.

mpokorny:
After much flailing, my best guess as to the cause of the build problems is that Travis uses virtualenv for Python, installing Python 3.5 by default in the build environment, the Ubuntu trusty version of libboost-python is built for Python 3.4, and the cmake script gets greatly confused by these differences (as am I). Installing numpy using pip and pip3 works great, whereas libboost-python is another matter since there is no installable pip package for it AFAICT. I have not been able to find a way for the cmake script to use Python 3.4 instead of 3.5 in searching for libboost-python, which is unfortunate, since Python 3.5 isn't packaged for Ubuntu trusty. If anyone has some ideas, I'm willing to try them out, since I'm out of ideas of my own for today.

tammojan:
The two suggestions I'd have:
 * Install boost-python with apt and try to tell travis to use a virtualenv with the exact same version.
 * Use conda as a(nother) package manager; I've done this in https://github.com/casacore/python-casacore (obviously you don't want the casacore package).

mpokorny:
Fixed in #639

PR #635, opened 2017-09-12
 	CAS-7309 add support to ImageOpener for componentlist images

dmehring:
Added COMPLISTIMAGE to ImageOpener::ImageTypes enum
Added logic to ImageOpener::imageType() for case when to return the new enum.
Added IPositionComparator class as part of IPosition files to allow std::map to be created using IPositions as keys. Added tests for that to tIPostion.cc

### DIFF
--- a/casa/Arrays/IPosition.cc
+++ b/casa/Arrays/IPosition.cc
@@ -1162,5 +1162,24 @@ void IPosition::throwIndexError() const
     throw(AipsError("IPosition::operator() - index error"));
 }
 
+bool IPositionComparator::operator ()(const IPosition& lhs, const IPosition& rhs) const {
+    size_t lhsSize = lhs.size_p;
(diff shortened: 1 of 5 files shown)


dmehring:
It appears that this build failed because it exceeded the time limit? Would someone who understands how travis works please take a look?

dmehring:
@tammojan @mpokorny can one of you help with this? It appears the build was proceeding fine but the time out was reached.
Thanks

tammojan:
The clang build passes fine, the gcc build (or more relevantly the non-c++11 build) times out, as happened in most other PR's. This problem is known, I will try to look into it at the end of this week. I don't think the time-out should prevent this PR to be merged.

As comment on the actual PR: it looks fine to me. Although the merge commit should ideally be left out.

dmehring:
I will submit a new PR that doesn't have the merge. Thanks.

PR #637, opened 2017-09-18
CAS-7309 add support to ImageOpener for componentlist images

dmehring:
* Added COMPLISTIMAGE to ImageOpener::ImageTypes enum and added code to ImageOpener.cc for recognizing such images.
* Implemented IPositionComparator to allow IPositions to be used as keys in ordered lists.
* Fixed bug regarding progress meter in LatticeStatistics.

### DIFF
--- a/casa/Arrays/IPosition.cc
+++ b/casa/Arrays/IPosition.cc
@@ -1162,5 +1162,24 @@ void IPosition::throwIndexError() const
     throw(AipsError("IPosition::operator() - index error"));
 }
 
+bool IPositionComparator::operator ()(const IPosition& lhs, const IPosition& rhs) const {
+    size_t lhsSize = lhs.size_p;
+    size_t rhsSize = rhs.size_p;
+    if (lhsSize == rhsSize) {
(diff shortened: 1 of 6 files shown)


dmehring:
@mpokorny The gcc build for this failed for the same reason as in PR 635 (https://github.com/casacore/casacore/pull/635). @tammojan noted in the comments of that PR that this should not prevent the PR from being merged. So, please review this PR at your earliest convenience. Thanks.

mpokorny:
I'll merge this PR, but I have a couple of comments, which I'll leave up to you to consider. First, it might be nice to have a more direct test of the IPositionComparator in tIPosition.cc; it seems that the test only demonstrates that an IPosition can be used as a key in a map, not that it is sorting in the way you intend. Second, I can't tell from this PR where you're using a map with an IPosition key, but if you're able to replace those instances of maps with unordered_maps, you wouldn't need the IPositionComparator, and the unordered_map might be faster.

dmehring:
Thanks. I hadn't considered the unordered_map option; I'll look at that.

PR #641, opened 2017-09-27
CAS-10767 make ZScoreCalculator::getMaxZScore() thread safe

dmehring:
Also reformat files changing tabs to spaces

### DIFF
--- a/scimath/Mathematics/ZScoreCalculator.cc
+++ b/scimath/Mathematics/ZScoreCalculator.cc
@@ -39,85 +39,85 @@ Mutex ZScoreCalculator::_mutex;
 Double ZScoreCalculator::getMaxZScore(uInt64 npts) {
     ScopedMutexLock lock(_mutex);
     if (_nptsToMaxZScore.empty()) {
-		// initialize the map
-		_nptsToMaxZScore[0] = 0.5;
-		_nptsToMaxZScore[1] = 1;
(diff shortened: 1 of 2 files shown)


dmehring:
@mpokorny Hi Martin, I issued this PR about a week ago. Please review at your earliest convenience. Thanks

mpokorny:
I'm slightly concerned about the indefinite lifetime and upward ratcheting size of ZScoreCalculator::_nptsToMaxZScore. There seems to be no way to reclaim any of the memory it uses, and it can change size in one direction only, although in practice, I'm guessing that the size would remain small. An implementation that avoids this potential issue might be preferable, but the possible design considerations are unknown to me. Please give my suggestion some consideration. After you reply to this comment, I'll merge the PR one or another.

dmehring:
Hi Martin,
That's why I've set an arbitrary upper limit of 1M elements:
`       if (_nptsToMaxZScore.size() < 1000000) {
            _nptsToMaxZScore[nptsmin] = z;
        }`
So at maximum it can only use about 16 MB of memory. And I suspect in practice, there would be only very rare, at best, use cases that approach this limit.

mpokorny:
Hi Dave, I see...I missed that...seems reasonable. I'll go ahead and merge.

dmehring:
Thanks!

PR #643, opened 2017-10-05
Do not set default diameter for VLBA to 25m

kettenis:
While the VLBA dishes themselves have a diameter of 25m, VLBA
observations regularly include Green Bank or (phased) VLA.
On top of that, other VLBI arrays that use the DiFX correlator
often set the array name in their FITS-IDI file for compatibility
with (older versions of) AIPS.  Setting the dish diameter to 25m
for these arrays is simply wrong.  Instead, rely on the diameters
set in the FITS-IDI file for the VLBA.


kettenis:
Discussed with Walter Brisken and others at the CASA-VLBI workshop currently in progress here at JIVE.

PR #644, opened 2017-10-05
Avoid incrementing TableIterator in copy assignment

mpokorny:
Replace the call to TableIterator::next(), which initializes the subTable_p
member, in the TableIterator copy assignment operator with a copy assignment of
the subTable_p member. This is intended for the copy to have the same state as
the copied instance.

### DIFF
--- a/tables/Tables/TableIter.cc
+++ b/tables/Tables/TableIter.cc
@@ -99,10 +99,10 @@ TableIterator& TableIterator::operator= (const TableIterator& iter)
 {
     delete tabIterPtr_p;
     tabIterPtr_p = 0;
-    subTable_p   = Table();
+    subTable_p = Table();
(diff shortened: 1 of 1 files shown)


mpokorny:
This is an alternative fix to issue #624 than the solution in PR #625, which maintains the desired copying of state in the copy assignment operator by avoiding the call to TableIterator::next(). This branch should certainly be tested using DPPP to check that it's OK there.

gervandiepen:
I'd prefer to leave
    tabIterPtr_p = 0;
before the if, so the pointer is 0 if ever the clone throws an exception.
You might also leave the clearing of subTable_p there, so the else branch is not needed, but that is not crucial.

mpokorny:
Good point about setting tableIterPtr_p to 0, initially. I might as well remove the else branch after doing that, too.

PR #645, opened 2017-10-10
CAS-9007 allow data provider to control max number of threads

dmehring:
decreases overhead and increases performance when less than the number of maximum threads available to openmp can be used.
improves run time of one of my image stats test cases by almost a factor of 2

### DIFF
--- a/lattices/LatticeMath/LatticeStatsDataProvider.tcc
+++ b/lattices/LatticeMath/LatticeStatsDataProvider.tcc
@@ -28,6 +28,7 @@
 #define LATTICES_LATTICESTATSDATAPROVIDER_TCC
 
 #include <casacore/lattices/LatticeMath/LatticeStatsDataProvider.h>
+#include <casacore/scimath/Mathematics/ClassicalStatistics.h>
 
 namespace casacore {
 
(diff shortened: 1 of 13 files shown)


lattices/LatticeMath/LatticeStatsDataProvider.tcc
```diff
@@ -147,6 +157,12 @@ void LatticeStatsDataProvider<T>::setLattice(
 		_currentSlice.assign(lattice.get());
 		_atEnd = False;
 	}
+#ifdef _OPENMP
+	_nMaxThreads = min(
+	    omp_get_max_threads(),
+	    lattice.size()/ClassicalStatisticsData::BLOCK_SIZE + 1
```
mpokorny: A minor point, but assuming that the value should be ceil(lattice.size()/BLOCK_SIZE), wouldn't this value be marginally better if it were (lattice.size() + BLOCK_SIZE - 1) / BLOCK_SIZE?

lattices/LatticeMath/MaskedLatticeStatsDataProvider.tcc
```diff
@@ -157,6 +167,12 @@ void MaskedLatticeStatsDataProvider<T>::setLattice(
 		_currentMaskSlice.assign(lattice.getMask());
 		_atEnd = False;
 	}
+#ifdef _OPENMP
+    _nMaxThreads = min(
+        omp_get_max_threads(),
+        lattice.size()/ClassicalStatisticsData::BLOCK_SIZE + 1
```
mpokorny: Same comment about implementation of ceil() as for LatticeStatsDataProvider.

dmehring:
Yes, you are correct. Although, I think the use of ceil() would be more self-explanatory and slightly more compact, with the drawback that explicit type casting would need to be used. I'll submit another PR for that. Thanks for the suggestion.

PR #648, opened 2017-10-18
Issue #638 Fixed HDF-1.10 problem with 64-bit hid_t

### LINKED ISSUE #638: typedef of hid_t mismatches hdf5 headers
The typedef of hid_t in HDF5Object.h of casacore is int, while in the header file of HDF5 hid_t is typedefed as int64_t.

This mismatch can cause casarest build error, when linking lwimager. 
 In detail, in libcasa_casa.so, the first parameter of the function 
casa::HDF5Group::init(long, casa::String const&, casa::String const&, bool, bool)
is long (i.e., int64_t)

But in casarest, when cmake with default settting (cmake ..), the HAVE_HDF5 marco seems not be defined, so that in HDF5Object.h, hid_t will become int, which cause casarest try to link to
casa::HDF5Group::init(int, casa::String const&, casa::String const&, bool, bool)

This problem can be hacked by replacing 
typedef int hid_t;
with 
typedef int64_t hid_t; 
in
 HDF5Object.h, 
so that I can successfully compile casarest.

gervandiepen:
In HDF5-1.10 the hid_t has been increased from 32 to 64 bits, which caused problems in HDF5Object. This has been fixed in a way that both 32 and 64 bit hid_t is supported.
Furthermore, HDF5DataSet has been extended with a get and put function to handle an Array object. Finally, the error messages in HDF5DataSet have been improved.

Close #638

### DIFF
--- a/casa/HDF5/HDF5Object.h
+++ b/casa/HDF5/HDF5Object.h
@@ -38,24 +38,12 @@
 #ifdef HAVE_HDF5
 # include <hdf5.h>
 #else 
-  typedef int                hid_t;
-  typedef unsigned long long hsize_t;
+  typedef Int64 hid_t;
+  typedef uInt64 hsize_t;
 #endif
 
 
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
-
-  // Define 2 functions to check that hid_t and hsize_t are mapped correctly.
-  // They are called by the constructor, so the compiler will scream if
-  // incorrect.
-  // <group>
-  void throwInvHDF5();
-  inline void check_hid_t (int) {}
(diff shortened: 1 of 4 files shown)


casa/HDF5/HDF5Object.h
```diff
@@ -38,25 +38,13 @@
 #ifdef HAVE_HDF5
 # include <hdf5.h>
 #else 
-  typedef int                hid_t;
-  typedef unsigned long long hsize_t;
+  typedef Int64 hid_t;
```
tammojan: should be `casacore::Int64`

PR #657, opened 2017-11-06
CAS-10842 improve LatticeStatistics performance

dmehring:
in cases where there are many sets for which to compute statistics,
the sets are small enough to fit comfortably in memory,
and the algorithm is not ClassicalStatistics (which is already
covered by the older TiledApply method).

Implementation is to multithread at a higher level than the
older algorithm, thus stats on each set is computed using a single
thread, and computing multiple sets is multi-threaded


dmehring:
@mpokorny would you please review this PR which I submitted about a week ago? Thanks

scimath/Mathematics/ClassicalStatistics.tcc
```diff
@@ -908,25 +905,19 @@ void ClassicalStatistics<CASA_STATP>::_accumulate(
 
 CASA_STATD
 uInt ClassicalStatistics<CASA_STATP>::_nThreadsMax() const {
+    uInt nthr = OMP::nMaxThreads();
 #ifdef _OPENMP
```
mpokorny: Is this necessary any longer?

lattices/LatticeMath/LatticeStatistics.tcc
```diff
@@ -928,98 +1222,95 @@ void LatticeStatistics<T>::_doStatsLoop(
...
+    T& overallMin, T& overallMax, T currentMin, T currentMax,
+    const IPosition& minPos, const IPosition& maxPos,
+    Bool atStart, const SubLattice<T>& subLat
+) {
+    // CAUTION The way this has worked in the past apparently for
+    // lattices is that the max and min positions are representative
+    // of the *entire* lattice, and were not stored on a sublattice
+    // by sublattice basis. This is easy to fix now,
+    // but for backward compatibility, I'm leaving this functionality as
+    // it has been.
```
mpokorny: Do you see any way forward, to move away from the historical semantics? Your description implies that the historical version is not what it should be.

dmehring: It really depends on what users want. So far they've not complained about this, and I don't want to change it for backward compatibility purposes. In the end, it seems to be an issue of minor concern. If they want these per lattice, it will require removing the original code, which we still use, because in some cases, it performs better than the newer code.

scimath/Mathematics/ClassicalStatistics.tcc
dmehring: the ifdef probably isn't required. I'm working on another stats related issue atm, so if this isn't required, I'll remove it on that branch.
Thanks for the review.

PR #661, opened 2017-11-16
CAS-10849 setImageInfo() must be called for changes to persist

dmehring:


gervandiepen:
Wouldn't it be easier and better to do:

+    ImageInfo rwii;
+    ndone += rwii.setInfoSplitBeamSet (ndone, info, img.shape(),
                                                     img.coordinates(),
                                                     latticeConcat_p.axis());
+    img.setImageInfo(rwii);

dmehring:
Done.

images/Images/ImageConcat.tcc
```diff
@@ -259,9 +259,11 @@ Bool ImageConcat<T>::setImageInfo (const ImageInfo& info)
   for (uInt i=0; i<latticeConcat_p.nlattices(); ++i) {
     ImageInterface<T>& img =
       dynamic_cast<ImageInterface<T>&>(*(latticeConcat_p.lattice(i)));
-    ndone += img.rwImageInfo().setInfoSplitBeamSet (ndone, info, img.shape(), 
+    ImageInfo& rwii = img.rwImageInfo();
```
gervandiepen: I think it is better to do
    ImageInfo rwii;
which works fine. It is somewhat dirty to change the internal ImageInfo object and set it to itself.

gervandiepen:
See the 'Files changed' for a slight comment.

images/Images/ImageConcat.tcc
dmehring: ok, then there is no point in getting the rw version. one can just make a copy of the read-only version if it is going to be copied anyway.

dmehring:
Done.

gervandiepen:
In fact 'ImageInfo rwii;' could be used instead of getting it from the imge, b ut it makes no difference. So it's fine now.

PR #662, opened 2017-11-17
Cas 10906 circumvent binning accounting issue by changing the bin sizes up

dmehring:
Also vector -> std::vector to play nice with eclipse.

### DIFF
--- a/scimath/Mathematics/ChauvenetCriterionStatistics.h
+++ b/scimath/Mathematics/ChauvenetCriterionStatistics.h
@@ -32,7 +32,6 @@
 #include <casacore/scimath/Mathematics/ConstrainedRangeStatistics.h>
 
 #include <set>
-#include <vector>
 #include <utility>
 
 namespace casacore {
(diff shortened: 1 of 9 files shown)


dmehring:
@mpokorny please review at your earliest convenience; my perception is that this is critical for the VLASS team. Thanks

mpokorny:
I'm not quite sure what the problem is that you're fixing beyond what's described in this PR, but it strikes me as something of a brute force workaround to a deeper issue (although it's possible I've missed something). I read your note about the alternative, as well as the urgency of a fix for VLASS, but do you have any plans to address this issue by an alternative solution again later? I worry that it will be difficult for someone reading this code at a later date to understand why this code is necessary, and what issue it addresses. Please consider adding a few more comments to the affected code before I merge this PR.

dmehring:
Thanks for the quick review and suggestion. I've added a comment. Hopefully it sheds more light on the issue.

dmehring:
Thanks!

PR #667, opened 2017-11-23
changed the niceCursorShape value for ImageConcat

kgolap:
changed the behavior of niceCurshorShape to pick the smallest (along 3rd, 4th etc  axes) of the niceCursorShapes of the constituent images of a concatimage. This speeds up access for iterators which use the default cursorshapes. CAS-10428 ticket on the casa-jira system


gervandiepen:
One build failed complaining that INT64_MAX does not exist.
Probably its better to use the <limits>;  std::numeric_limits<Int64>::max()

PR #670, opened 2017-11-27
Cas 10428

kgolap:
This is a change in behaviour when calling niceCursorShape for the ImageIterators. Right now ConcatImage is returning a cursor shape which is as if it a real image on disk and it can be very slow when it is trying to stride across multiple images (image statistics for example can be 10 times slower than a real concatenated image).  This change make sure that the "nice" cursor shape is the minimum of each of the constituents images (along the non X-Y axes ...e.g spectral). In doing that imageStatistics for example take the same amount of time than a real concatenated image.

### DIFF
--- a/images/Images/ImageConcat.tcc
+++ b/images/Images/ImageConcat.tcc
@@ -487,7 +487,32 @@ Bool ImageConcat<T>::doGetMaskSlice (Array<Bool>& buffer,
 template <class T>
 IPosition ImageConcat<T>::doNiceCursorShape (uInt maxPixels) const
 {  
-   return latticeConcat_p.niceCursorShape(maxPixels);
+  //Return the smallest cursor shape along the non X-Y direction from the constituent images
+  if(nimages() > 0){
+    ///if image has 1 or  2 axes return cursor shape of first image
+    if(shape().nelements() <= 2)
+      return image(0).niceCursorShape(maxPixels);
+    Vector<Int> dirpixaxes, dirworldaxes;
+    Int coordaxis;
+    CoordinateUtil::findDirectionAxes(dirpixaxes, dirworldaxes,
+                                       coordaxis,
+				      coordinates());
+    Int64 minprod=INT64_MAX;
+    Int minimage=-1;
+    for (uInt k=0; k < nimages(); ++k){
+      IPosition curshape=image(k).niceCursorShape(maxPixels);
+      Int64 prod= curshape.product()/ Int64(curshape(dirpixaxes(0))*curshape(dirpixaxes(1)));
+      if(prod < minprod){
+	minprod=prod;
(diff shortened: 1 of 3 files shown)


mpokorny:
Kumar, would you please provide a short, one-line description of this pull request? CAS10428 isn't very descriptive outside of CASA itself.
FYI, I noticed that you submitted a few PRs that you then closed again, presumably after fixing bugs uncovered by the build process. You could instead just push more commits to your branch; your branch will get built and tested again, and the PR will be updated.

kgolap:
Thanks for the info about PR getting updated automatically with pushes to my branch.

images/Images/ImageConcat.tcc
```diff
@@ -487,7 +488,34 @@ Bool ImageConcat<T>::doGetMaskSlice (Array<Bool>& buffer,
...
+    Int64 minprod=std::numeric_limits<Int64>::max();
+    Int minimage=-1;
+    for (uInt k=0; k < nimages(); ++k){
+      IPosition curshape=image(k).niceCursorShape(maxPixels);
+      Int64 prod= curshape.product()/ Int64(curshape(dirpixaxes(0))*curshape(dirpixaxes(1)));
+      if(prod < minprod){
+	minprod=prod;
+	minimage=k;
+      }
+      return image(minimage).niceCursorShape(maxPixels);
```
gervandiepen: Shouldn't this line be outside the for loop?

gervandiepen:
There does not seem to be a test for this new piece of code.
It would be good to document in the .h file what the function is doing.
I also commented on the code.

images/Images/ImageConcat.tcc
```diff
@@ -487,7 +488,34 @@ Bool ImageConcat<T>::doGetMaskSlice (Array<Bool>& buffer,
 template <class T>
 IPosition ImageConcat<T>::doNiceCursorShape (uInt maxPixels) const
 {  
-   return latticeConcat_p.niceCursorShape(maxPixels);
+  
+  //Return the smallest cursor shape along the non X-Y direction from the constituent images
+  if(nimages() > 0){
+    ///if image has 1 or  2 axes return cursor shape of first image
+    if(shape().nelements() <= 2)
```
gervandiepen: A multi-line if or for without braces is considered bad practice.
A single line if/for is also bad, since it makes debugging and code coverage tests harder.

kgolap:
I fixed the order of closing bracket goof in for loop ..i had it right the first time.  And modified code + comment on Ger suggestions

gervandiepen:
There needs to be a test for this function. Otherwise it cannot be checked if the code is correct.

kgolap:
Ok have added niceCursorShape tests to tImageConcat.cc

PR #673, opened 2017-11-29
wcslib threadsafety

rtobar:
wcslib is thread-unsafe in two different ways, at least:

 * wcsset calls (indirectly) a non-reentrant flex scanner. Thus, calls to wcsset need to be serialized.
 * wcsini/wcssub read/write to a global variable. Thus, they need to be serialized.

More information is in the individual patches, which not only do this serialization but also remove code duplication, and add a unit test to ensure the fix is working.


dpetry:
No matter how this is reviewed, did you contact Mark Calabretta about the problems you found?

rtobar:
Yes, I did, just after submitting these patches.

rtobar:
With the latest commit all tests pass in the Linux Travis machines. The OSX builds are all failing even before building, with homebrew not being able to install gcc.

gervandiepen:
The changes look very good. Thank you very much for the fixes.
I'll leave it to Tammo Jan to do the merge (as he is the suggested reviewer).

coordinates/Coordinates/Coordinate.cc
```diff
@@ -1241,12 +1241,65 @@ void Coordinate::xFormToPC (::wcsprm& wcs, const Matrix<Double>& xform) const
...
+    static Mutex wcsset_mutex;
+    ScopedMutexLock lock(wcsset_mutex);
     if (int iret = wcsset(&wcs)) {
         String errmsg = "wcs wcsset_error: ";
         errmsg += wcsset_errmsg[iret];
         throw(AipsError(errmsg));
     }
 }
 
+#if WCSLIB_VERSION_MAJOR == 5 && WCSLIB_VERSION_MINOR >= 7
```
tammojan: Perhaps add an `|| WCSLIB_VERSION_MAJOR > 5` to this.

tammojan: (I added it myself, will merge once the tests pass.)

rtobar:
Thanks for merging the code in this branch, I really appreciate it. In the meanwhile, today Mark came back with some feedback regarding the underlying issues.

Bad news is that the extent of the two problems I identified (as stated in the first comment of the pull request) is actually bigger than I originally anticipated. Mark went into the code and had a closer look, and here are the results. He categorized thread-unsafe calls into three groups: flex-scanner-related, `NXXMAX`-related (three global variables used throughout the code), and the `wcserror_enable()` function. Next to each call under each category is shown its usage internally within wcslib (e.g., you can see the original `wcsset` -> `wcsulexe` dependency I pointed out for example).

```
Flex:
  fitshdr.
  wcsbth.
  wcspih.
  wcsulexe  wcsunitse   wcs_units  wcsset  wcsbchk.
                                           wcsp2s   cylfix  wcsfix.
                                                            wcsfixi.
                                           wcss2p   wcsmix.
                                           wcsmix.
                                           wcssptr.
                                           celfix   wcsfix.
                                                    wcsfixi.
                                           cylfix (see above)
                                           wcshdo.
                        wcsunits.
            wcsulex.
  wcsutrne  wcsutrn.

NXXMAX:
  NPVMAX    wcsini(ro)  wcsbth_init1  wcsbth.
                        wcspih_inits  wcspih.
			wcssub.
  wcsnpv    wcsbth(rw).
            wcspih(rw).
            wcssub(rw).
  NPSMAX    wcsini(ro)  (see above)
  wcsnps    wcsbth(rw).
            wcspih(rw).
            wcssub(rw).
  NDPMAX    disini(ro)  discpy  lincpy        linwarp.
                        linini  lincpy        (see above)
			        wcsini        (see above)
			lindis  wcspih_inits  (see above)
			        wcspih_final  wcspih.
				linwarp.
				wcssub.
				wcsset        (see above)
            discpy(rw)  (see above)
  disndp    wcspih(rw).
            wcssub(rw).
            wcsset(rw)  (see above)

wcserr:
  wcserr_enable.
```

As it can be seen, a big number of calls are affected by `wcsulexe`, including some that I didn't include in this original pull request by that are still used by casacore's Coordinte classes. I think I can easily create the necessary patches to serialize all remaining `wcsXXX` calls in the Coordinate classes. Should I go ahead with that as a separate pull request?

Good news is two fold: on the first hand, Mark says:

> After reading the manual again I realised that I could make them [the flex scanners] reentrant without changing the API  simply by writing a thin wrapper for each one.  It seems to work and should fix a swag of problems  created by `wcsulexe()`.

On the other hand, he is also reviewing the patch I sent to remove the need for having to keep global variable s in the library, which is not final but sets the basis for the work that is needed I think. Once these two improvements are implemented, we should see a wcslib version that is fully thread-safe.

rtobar:
Mark came back today with a preview of wcslib 5.18, which incorporates a number of fixes that make wcslib fully thread safe. I have the additional patches in casacore ready for this, as well as patches to provide better thread-safety for wcslib <=5.17. Please let me know how to proceed.

dpetry:
On what timescale is Mark planning to publish 5.18?

rtobar:
I'm not completely sure, but he sent me the pre-release today and it seems likely that he will publish this pretty soon. His words:

> I plan to produce a pre-release of 5.18 which I will give you privately for testing before releasing it  publically.  Probably tomorrow.

dpetry:
Then maybe we should wait until he publishes this and base further development on this new version.

rtobar:
wcslib 5.18 (which includes the thread-safety fixes I mentioned before)  was published yesterday. Would it now make sense to review my additional set of patches that take this into account?

gervandiepen:
On a system we are using wcslib-4.25 and it has the same problem. So I've commented out the #ifdef in wcslib version, whereafter tDirectionCoordinate runs fine.

I think we can uncomment those lines, but only not use the lock for wcslib 5.18 and higher after testing it carefully.

rtobar:
Hi Ger,

Thanks a lot for continuing looking into this. I did produce a set of extra patches in the past at the time when I came into contact with Mark, but after this pull request was accepted, so they may not have become visible. These patches better address the versions of wcslib, and also provide a unique lock around every wcslib-related call, since it became clear that the thread safety problems affected not only the functions identified in the first set of patches, but almost all wcslib functions eventually. In particular, please see https://github.com/rtobar/casacore/commit/ce8fb6a572fccae39d46f968666beabab1b8b453 and https://github.com/rtobar/casacore/commit/ef83d72a6210c83951a0191bfedbf1554e078172, which exist on my wcslib-threadsafety branch, on top of what has been already accepted. These changes should both fix the problem you mention in your comment above, and also provide more secure multithreading for the rest of the wcslib operations.

Please let me know if you want me to create a new pull request for these changes; otherwise you can also manually pull them. Tests pass on my CXX11=ON/USE_THREADS=ON build in Linux, and the original threading problems that triggered this work (when using casacore with ASKAPSoft) are effectively gone.

PR #674, opened 2017-12-04
Initialize preselected_p properly even if no DATA or FLOAT DATA columns exist

cquike:
This fixes a problem with commit 689fb9b487b2d2174faafdb19197e8fc9fd93f81
reported by valgrind. The preselected_p was left uninitialized when
these columns are not present.

This is part of fixing CAS-10458

### DIFF
--- a/ms/MeasurementSets/MSIter.cc
+++ b/ms/MeasurementSets/MSIter.cc
@@ -571,6 +571,8 @@ void MSIter::setMSInfo()
         }
       }
     }
+    else
+      preselected_p = false;
 
     // determine the reference frame position
     String observatory;


mpokorny:
@tammojan, the last build appears to have failed because of some issue in the build infrastructure. Would you please have a look? My opinion is that this branch can be merged with master as is; do you agree?

tammojan:
Fine to merge if just the OSX builds fail. Travis seems to have problems to keep the environment stable. 
The change looks good to me, we could consider enabling a CI-build with valgrind.

cquike:
I fully agree that a build with valgrind would be very useful.

mpokorny:
@cquike, I've created issue #675 for discussion of using valgrind/memcheck in the Travis CI build.

PR #676, opened 2017-12-04
Cas 10948 refactor to prevent multiple creation of stats algorithm objects

dmehring:
when computing accumulated and quantile statistics. This was a significant and unnecessary bottleneck for iterative algorithms like Chauvenet, especially on large images.


dmehring:
The failures here appear to be CI software related and not due to my changes. Would someone who understands the CI system please have a look? Thanks

mpokorny:
I believe you're right about the build system problems. I've seen similar issues lately. @tammojan tells me that if the issues are only with the OSX builds, I shouldn't let that prevent a merge. It won't be before tomorrow that I'll have a chance to look at the code changes in your PR, however.

dmehring:
Thanks Martin.

dmehring:
Thanks Martin.

PR #679, opened 2017-12-07
Return a shared_ptr rather than a naked pointer in NewMSSimulator

cquike:
So far, the MS created was stored in a pointer internal to the class but it
was also returned to the client. One of the clients is the MsFactory class in
casa/code, which also passes the pointer to the requester. This can be rather
confusing in terms of who owns actually the pointer and should deallocate it.
A shread_ptr will help to keep the lifecycle of the pointer better
controlled.

A very basic unit test on the NewMSSimulator class has also been added.

This has a side work of CAS-10458. The change implies changes small changes in two classes in casa itself. Once the pull request is merged I will make the changes in the casa branch and will update casacore in the casa branch, in order to keep everything consistent.


mpokorny:
I believe that the "casacore way" would be to use a casacore::CountedPtr. std::shared_ptr made its debut in C++11, but casacore should build under versions of C++ prior to 11, as well. The casacore::CountedPtr class is implemented using a std::shared_ptr under C++11, but its interface is not identical.

cquike:
Thank you Martin for your comment. I just saw that indeed travis failed for the CXX11=False job. Somehow I thought that C++11 was already mandatory for casacore. For casa/code that's the case, right?

mpokorny:
Yes, that's right.

ms/MSOper/NewMSSimulator.cc
```diff
@@ -288,7 +288,7 @@ NewMSSimulator::NewMSSimulator(const String& MSName) :
   }
 
   // Now we can create the MeasurementSet and add the (empty) subtables
-  ms_p=new MeasurementSet(newMS,0);
+  ms_p = std::make_shared<MeasurementSet>(newMS,0);
```
tammojan: Should also be a CountedPtr

ms/MSOper/NewMSSimulator.h
```diff
@@ -237,7 +238,7 @@ class NewMSSimulator
   Int lastSpWID_p;
   Int lastNchan_p;
 
-  MeasurementSet* ms_p;
+  std::shared_ptr<MeasurementSet> ms_p;
```
tammojan: Should also be a CountedPtr

cquike:
Hi,
I have changed to CountedPtr to allow compilation without C++11 support. Also one of the unit tests is disabled.
travis shows that the linux builds are fine, the macOS ones seem to have some problem relates to infrastructure.

tammojan:
Great, thanks for making the change. Could you squash your commits such that only one or a few commits remain? I'd like to avoid commits like `fix typo` to make bisecting in the future easier.

cquike:
Hi,
yes, I think it is a good idea to squash it. I have no much experience with git squashing, but I understood that a squashing after the commits have been pushed is not possible, right? I think that anyway you have the option when accepting the pull request to squash everything together, but then it doesn't appear as a merge commit...

cquike:
Should I use rebase, like here?
https://stackoverflow.com/questions/5667884/how-to-squash-commits-in-git-after-they-have-been-pushed#5668050

tammojan:
Yes, using rebase -i should do the trick, as mentioned on stackoverflow.

cquike:
Done!

PR #680, opened 2017-12-08
Fix memory leaks in the Measures module

cquike:
This pull request fixes some memory leaks in the Euler and MVPosition classes.
The part that leaks is the caching mechanism to avoid too many allocations put in place around two year ago. The leak is actually small but it clutters the output of valgrind in almost any unit test in casa/code, by outputting 150+ errors.
I have cleanup the output of valgrind for the tMSDerivedValues unit test but there are still 8 remaining memleaks, which I didn't dare to clean up since they look a bit more complex to me.
This patch uses smart pointers to ensure RAII, with the caveat of being a bit slower than the raw pointers used before. The slowdown is around 50% (only in the construction/destructor code) but it is still 5x faster than not using cache at all.

### DIFF
--- a/casa/Quanta/Euler.cc
+++ b/casa/Quanta/Euler.cc
@@ -41,7 +41,7 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
 // simplistic Vector(3) cache to reduce allocation overhead for temporaries
 #if defined(AIPS_CXX11) && !defined(__APPLE__)
 thread_local size_t Euler::available = 0;
-thread_local Euler::DataArrays Euler::arrays[50];
+thread_local std::vector<Euler::DataArrays> Euler::arrays(max_array_cache);
 #endif
 
 Euler::DataArrays Euler::get_arrays()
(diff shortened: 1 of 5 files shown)


cquike:
Hi,
I attach here the valgrind reports before and after the change.

The tests seem to pass, although the macOS builds still have issues...

[tMSDerivedValues.valgrind.after.txt](https://github.com/casacore/casacore/files/1548670/tMSDerivedValues.valgrind.after.txt)

[tMSDerivedValues.valgrind.txt](https://github.com/casacore/casacore/files/1548669/tMSDerivedValues.valgrind.txt)

cquike:
Hi,
did you have the chance to look into this pull request?

Regards,
Enrique

cquike:
Sorry, I inadvertently closed the pull request with the last comment

dpetry:
Hi Ger, could you have a look?

gervandiepen:
The change looks fine to me. However, has it been tested in a multi-threaded OpenMP run?

cquike:
Hi,
sorry for the late reply. Thank you for raising the point of OpenMP testing. I have expanded tMVPosition.cc to test both a C+11 thread test case (test_parallel()) as well as OpenMP (test_parallel_omp()). However I have disabled the OpenMP test by default. The reason is that valgrind tools (memcheck, drd and helgrind) report many false positives using thread_local storage and OpenMP. In comparison the C++11 threaded test case is clean under all three valgrind tools. I have also tried with AddressSanitizer and ThreadSanitizer enabled and the don't report any issue, both with C+11 threads and OpenMP.

PR #681, opened 2017-12-08
Initialize tableiterator in assignment like in constructor

tammojan:
It turns out the solution in #644 did break the DPPP build; I'm not sure how this slipped (I tested it then, but I don't know how it could have passed). Unfortunately this also means that the released 2.4.0 under some circumstances does not work well with DPPP.

In the following code, I expect `iter1` and `iter2` to be really the same:
```
TableIterator iter1(tab1, iv1, compObj, orders);
TableIterator iter2 = TableIterator(tab1, iv1, compObj, orders);
```

It turns out that although they have the same `nrows()`, they yield a different number of rows. This is demonstrated in the test in this PR.

I think it is valid to call `next()` in the copy constructor, because it is also called once in the normal [constructor](https://github.com/casacore/casacore/blob/master/tables/Tables/TableIter.cc#L52) of TableIterator. This first call of `next()` yields the initial state of a newly constructed TableIterator. So I do not think it copies the state of the copied constructor, as suggested in #644.

In short, I think the solution of #624 and #625 is valid and does not copy the state as suggested in #615. It merely initializes the newly created TableIterator in the same way as in the normal constructor.

### DIFF
--- a/tables/Tables/TableIter.cc
+++ b/tables/Tables/TableIter.cc
@@ -103,6 +103,7 @@ TableIterator& TableIterator::operator= (const TableIterator& iter)
     if (iter.tabIterPtr_p != 0) {
         tabIterPtr_p = iter.tabIterPtr_p->clone();
         subTable_p = iter.table();
+        next(); // Get first subtable, as in constructor
     }
     return *this;
 }
--- a/tables/Tables/test/tTableIter.cc
+++ b/tables/Tables/test/tTableIter.cc
@@ -228,6 +228,32 @@ void doiter3()
     iter2 = iter1;
     AlwaysAssertExit(iter2.table().nrow() == iter1.table().nrow());
 
+    // Test that copied new iterator gives the same number of rows
+    int iter1count = 0;
+    while(!iter1.pastEnd()) { iter1.next(); iter1count++; }
+    int iter2count = 0;
+    while(!iter2.pastEnd()) { iter2.next(); iter2count++; }
+    AlwaysAssertExit(iter1count == iter2count);
+    iter1.reset();
+    iter2.reset();
+
+    // Test that copied new iterator does not copy the state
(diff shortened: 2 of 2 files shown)


mpokorny:
@tammojan, I have a recent change to a local casacore branch that may affect this bug. I've checked out your branch locally, and I'm inserting my change into that branch, with the intention of running your new test on the merged code. However, the weekend is almost here, and I'm not sure how far I'll get today. I don't know I'll have time to look at this over the weekend -- I hope you can hang tight for a few days.

tammojan:
> I hope you can hang tight for a few days.

Sure; hope you enjoyed the weekend. Another solution to the problem is fine with me too, I hope you do agree that `TableIterator a = TableIterator(...)` and `TableIterator a(...)` should do the same. If the test in this PR passes in your code, DPPP should be fixed.

It would be nice if this could be fixed this week, so that we can roll out a 2.4.1 with the fix.

mpokorny:
I agree that `TableIterator a = TableIterator(...)` and `TableIterator a(...)` should do the same thing, but that involves the move assignment operator, not the copy assignment operator. The sequence
```
TableIterator a(...);
// some methods calls on a ...
TableIterator b = a;
```
should not necessarily have the same outcome, since the last statement invokes the copy assignment operator. I think it would be preferable to copy the state of `a` to `b` in the last statement. I see that you've updated the test code a bit since last week -- I'll rerun some tests here today, and see whether I can come up with a concrete suggestion.

mpokorny:
My tests show that this PR would not affect the CASA features that I have been developing. As for the semantic issues of move vs copy assignment that I raised above, I suppose that this PR is OK in light of the TableIterator::copyState() method, which is sufficient to obtain the outcome that my work needs. The modified tTableIter test at least provides an executable test of the current behavior.

PR #684, opened 2017-12-12
casacore configuration sans environment variables

schiebel:
As work begins on CASA's (https://casa.nrao.edu) transition to Python 3, we would like to fix structural issues at the same time because the Python 3 conversion will necessarily involve changes and fixes on the part of CASA's users. Two related things we would like fix is (1) removing CASA's dependency on environment variables (CASAPATH in this case) without requiring that static paths be compiled into the application, and (2) support checking multiple directories when attempting to open image or measurement sets (i.e. a data path in the $PATH sense of the word).

This change introduces a hook for applications which use casacore to specify a data path, and it includes changes so that measures is able to find the IERS data when it is included in the optionally specified data path. Other casacore state could be added the AppState class as time goes by (and more environment variable dependencies are uncovered), but this change is sufficient for our first cut at python 3 support and its accompanying reorganization.

The use of this hook is obviously optional. Not specifying a data path (the default) leaves the behavior of casacore unchanged (at least that is the objective).

### DIFF
--- a/casa/CMakeLists.txt
+++ b/casa/CMakeLists.txt
@@ -199,6 +199,7 @@ System/AipsrcValue2.cc
 System/AipsrcVBool.cc
 System/AipsrcVString.cc
 System/AppInfo.cc
+System/AppState.cc
 System/Casarc.cc
 System/Choice.cc
 System/ObjectID.cc
@@ -579,6 +580,7 @@ System/AipsrcValue.tcc
 System/AipsrcVector.h
 System/AipsrcVector.tcc
 System/AppInfo.h
+System/AppState.h
 System/Casarc.h
 System/Choice.h
 System/ObjectID.h
--- a/casa/System/AppState.cc
+++ b/casa/System/AppState.cc
@@ -1,0 +1,60 @@
+//# AppState.cc: casacore library configuration without environment variabes
+//# Copyright (C) 2017
+//# Associated Universities, Inc. Washington DC, USA.
+//#
+//# This library is free software; you can redistribute it and/or modify it
+//# under the terms of the GNU Library General Public License as published by
+//# the Free Software Foundation; either version 2 of the License, or (at your
+//# option) any later version.
+//#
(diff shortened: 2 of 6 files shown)


gervandiepen:
Casacore does not allow C++11 constructs yet (you use a lambda), although we hope to do so soon. Therefore one build fails.
Other remarks:
- In the documentation you mention AppStateStore, but I don't see one. Do you mean AppStateSource?
- Setting and getting the static *user_state is not thread-safe.
- The include guard should be CASA_APPSTATE_H.
- There is no test program.
- I don't see how the datapath can be set.

schiebel:
Thanks for your quick and succinct comments. The failure was not due to the use of a lambda function (I knew to expect this problem and included a hack to avoid the lambda function; the problem was a gcc 4.8.* defect which prevents local types from being used to expand templates; which I didn't anticipate). Your comments were helpful and the update I just pushed is designed to address them.

This class is designed as a hook to provide applications using casacore with the greatest flexibility in providing settings. I wanted the application to get control when state is fetched so that e.g. it could be generated dynamically, fetched from a DB, or retrieved from some shared-state process. For that reason, setting, changing, etc. of the state is delegated to the application class(es) derived from AppState while the AppState base class just provides the interface which casacore uses to retrieve the state from the wider application. The new test program provides an example.

gervandiepen:
I see. 
There is still a possible race condition (with initialize) in the fetch function. It will return default_state when user_state is set between the test and return. However, I assume it is acceptable; it certainly does not cause havoc.

You say you want to use unique_ptr once we support C++11. You could use (S)PtrHolder for the time being, but as I expect we will support C++11 very soon, we can leave it as is.

schiebel:
I think this is a good point. While I agree that the initialize/fetch race condition is benign, I think the race condition this forces the application (using casacore) into with dataPath returning a reference is not. The last fix forgoes the efficiency of a reference for the safety of a value.

schiebel:
From my perspective, this is ready to be merged...

PR #688, opened 2017-12-15
Cas 11028 Create StatisticsDataset to encapsulate data specific code

dmehring:
Also various refactors including moving static methods from StatisticsAlgorithm et al to StatisticsUtilities.


mpokorny:
David, do the current unit tests cover the changes in this pull request? I'm guessing that they do, but it's a large set of changes, and I'd like to verify.

dmehring:
Hi Martin,
Yes. The setData(), addData(), and setDataProvider() calls all call methods in the new class. There are many, many  setData(), addData(), and setDataProvider() in the associated statistics algorithm related unit tests, and other related unit tests (such as in the Lattice and Image statistics unit tests). During development, I used the casacore tests to find issues in the implementation. Outside of casacore, I've also verified that the statwt2 tests and the ia.statistics() tests all pass.

dmehring:
Thanks.

PR #689, opened 2018-01-03
CAS-11050 parallelize ClassicalStatistics::_doNpts()

dmehring:
using similar pattern to what has been used in other methods that
have been parallelized in this class. In addition to obvious performance
boost, is also necessary for further refactoring so similar patterns
can be used.

* Add unit tests.
* vector -> std::vector in some places in tClassicalStatistics to play more nicely with eclipse.

### DIFF
--- a/scimath/StatsFramework/ClassicalStatistics.h
+++ b/scimath/StatsFramework/ClassicalStatistics.h
@@ -712,6 +712,11 @@ private:
         WeightsIterator weightsIter, uInt64 dataCount
     );
 
+    void _computeNpts(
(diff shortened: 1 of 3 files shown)


dmehring:
Note that the CI failure was due to a server side networking issue apparently, and is not due to my modifications. So, my mods can be reviewed.

dmehring:
@mpokorny Hi Martin, Please review at your earliest convenience; I have another branch that depends on this one for which I also need to submit a PR. I think the review for the current branch should be pretty straight forward.
Thanks
Dave

dmehring:
Thanks Martin.

PR #690, opened 2018-01-10
CAS-11065 fix progress meter estimated number of steps

dmehring:
Minimum number of full lattice scans when _doRobust = True is 3, not 2.


dmehring:
@mpokorny Hi Martin,
Would you please review this one line code change at your earliest convenience?
Thanks

Issue #675, opened 2017-12-04
Enable valgrind in Travis CI


mpokorny:
Should valgrind be employed in the Travis CI build? If so, what components of valgrind -- only memcheck? Does anyone do this on a regular basis with their own builds, and if so, how "clean" is casacore vis a vis memcheck currently? Would we want or need to add some custom suppressions?

gervandiepen:
I don't think we should add it to the Travis build as the runtime will be
too long.
Regularly I run my tests with valgrind (memcheck only), but must confess I
haven't done so lately. The code is pretty clean; in the past I've fixed
leaks and so I found.

Note that the Casacore build system has support for running tests with
valgrind. For myself I've defined the following shell function
    ctestcheck() { (export CASACORE_CHECK=1; ctest "$@") }
showing how to use the Casacore valgrind support. It creates a .valgrind
file for each test run through valgrind, but only keeps the files showing
leaks, etc. and shows the names of these files, so you know which tests
have problems.
Note that if a test is run using a .run file, the env.var. $casacore_check
should be used to be able to memcheck the test program runs in that script.
Also note that a .run file can be used to not run a test through valgrind
if it would take too long (by not using that env.var.).

As a start you could give it a try in e.g. build/opt/casa to memcheck all
casacore/casa test programs.
Note that running a dbg version gives better tracebacks.

gervandiepen:
Another remark on using valgrind during testing.
In principle
    ctest -T memcheck
can be used to run tests through valgrind. However, casacore has the extra assay script running the test and checking the output. ctest does not know about it, so using -T would only memcheck the (ba)sh command starting the script.

I did ctestcheck on the casa module.
In opt mode valgrind reported a few issues due to openmp threading leaking some memory. In dbg mode everything was fine.

I noticed tHDF5File failed due to a difference in output which I fully understand and will fix. The strange thing is that it does not fail for a normal ctest, which I'll look into.

cquike:
Apart from a few memory leaks (see pull request #680 ), I also find memory issues reported by valgrind of this kind:
==4121== Syscall param write(buf) points to uninitialised byte(s)
and less often like this:
==4121== Conditional jump or move depends on uninitialised value(s)
This usually indicates undefined behaviour, but it might be a false positive by valgrind. Did anybody investigate into that?

gervandiepen:
The first one is known and is suppressed (with a few other valgrind errors)
when running the tests using casacore's valgrind support. Alignment might
[...]

PR #691, opened 2018-01-10
CAS-11049 accounting-specific methods/fields 

dmehring:
…taset

from ClassicalStatistics to StatisticsDataset
Also, bundle data chunk-specific fields into a single struct and reduce number of necessary fields by using PtrHolders to indicate if mask, weights, and ranges are set to make it easier to pass them to and return them from methods.

### DIFF
--- a/scimath/StatsFramework/ClassicalStatistics.h
+++ b/scimath/StatsFramework/ClassicalStatistics.h
@@ -179,10 +179,6 @@ public:
     // see base class description
     virtual std::pair<Int64, Int64> getStatisticIndex(StatisticsData::STATS stat);
 
-    // Has any data been added to this object? Will return False if the object has
-    // been reset and no data have been added afterward.
-    // Bool hasData() const  { return _hasData; }
-
     // reset object to initial state. Clears all private fields including data,
     // accumulators, etc.
     virtual void reset();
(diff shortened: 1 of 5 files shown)


dmehring:
@mpokorny Hi Martin, Along with looking at the one line change I mentioned a few days ago in PR 690, I'd greatly appreciate it if you could also review this one which I submitted about a week and a half ago. I have other work in the queue which depends upon it.
Thanks

mpokorny:
David, have you tried running your test code under valgrind (or similar tool) to check for memory leaks? I don't see any obvious issues, but it might be worth your time to try something like that (if you haven't already done so.)

dmehring:
Hi Martin,
I always use smart pointers precisely to avoid memory leaks. Do you see any bare pointers allocated with new in my code? Because there shouldn't be.

mpokorny:
No, I don't see any problems, my question was not specifically in reference to this PR. I agree that you should use smart pointers, but it's nice to do a test for leaks occasionally -- even when using smart pointer there could still be leaks (e.g, circular references).

dmehring:
Hi Martin
I don't routinely run valgrind. I agree it is not a bad idea to do so, although my preference would be that it be done periodically as part of automated CI, if possible. It seems to me that would be a much better approach than to rely/expect developers to do it periodically, since it is a non-trivial thing to do and I suspect it could be automated and all casacore/casa C++ could be run using it in such an environment.
Thanks for the reviews today.

mpokorny:
FYI, there's some discussion about valgrind running in the CI build in #675.

PR #695, opened 2018-01-27
check flag while traversing data directories

schiebel:
This bug from my last commit was missed because the tables were found from other locations.


mpokorny:
Looks OK, but would you please change the PR title to be a bit more descriptive of the change? casacore is trying to stick to the recommendations something along the lines of what's described [here](https://chris.beams.io/posts/git-commit/), although not strictly that set of recommendations in particular.

schiebel:
OK, thanks @mpokorny . How does this title look?

mpokorny:
I reworked the title to use the imperative mood

PR #698, opened 2018-01-31
Removed semicolon after extern "C" { ... }

jjdmol:
Pedantic compilers complain about the ; after extern "C" { ... }, which indeed is not required.


tammojan:
Thanks!

PR #699, opened 2018-01-31
Fixed deprecated check for Linux.

jjdmol:
The `__linux` define is deprecated (see https://sourceforge.net/p/predef/wiki/OperatingSystems/), we should use `__linux__` instead.


tammojan:
As a background: we're trying to make nicer commit messages. See e.g. https://chris.beams.io/posts/git-commit/ . The proper commit message for this would be `Fix deprecated check for Linux`, i.e. imperative, no period at end. For next time :)

PR #700, opened 2018-01-31
CAS-11077 refactor quantile-related methods

dmehring:
into separate classes to decrease maintenance costs.

There are no (purposeful) changes in functionality, just a lot of refactoring and reformatting.


dmehring:
@mpokorny Hi Martin. I know this looks like a big one, but really all it mostly is is moving code from old classes to new classes. There is some other minor refactoring and reformatting that took place, but there are no (purposeful) changes in functionality. All relevant casacore and CASA tests pass in my sandbox. Unfortunately, git seems to have gotten confused in a couple of places in the diffs; or at least some of the changes of removed code to new code don't match up with how I would have presented it.
Anyway, I didn't want the size of the changes to alarm you. The bottom line here is this is to lower maintenance costs, which were becoming fairly high for some of the larger classes especially, just because of all of the code in individual files. I did check for any performance changes, and for the cases I used there was no degradation in performance, and in a couple of cases, there is a significant increase in performance (of order 25%).

mpokorny:
David, it looks like there was a failure in the non C++-11 build of your branch, and it's not build system related. Would you please have a look?

dmehring:
@mpokorny Thanks for the heads up. Looks like something C++11 lets through but older versions do not. Looks like it was a mindless copy/paste error on my part anyway. Fixing now.

dmehring:
Thanks Martin.

PR #701, opened 2018-02-09
Concatenate SYSCAL tables

kettenis:
When concatenating MeasurementSets the associated SYSCAL tables should be concatenated as well. Otherwise only Tsys values from the first MeasurementSet survive and amplitude calibration will fail for the appended data. With this change the importfitsidi task in CASA can be used to read in multiple FITS-IDI files that contain a SYSTEM_TEMPERATURE table with Tsys measurements.

For "virtual" concatenation, additional values are appended to the SYSCAL table of the first MeasurementSet when creating the MMS.

### DIFF
--- a/ms/MSOper/MSConcat.cc
+++ b/ms/MSOper/MSConcat.cc
@@ -345,7 +345,10 @@ IPosition MSConcat::isFixedShape(const TableDesc& td) {
   if(!antIndexTrivial){
     copyPointingB(otherMS.pointing(), newAntIndices);
   }
-  
+
+  // SYSCAL
+  copySysCal(otherMS.sysCal(), newAntIndices);
+
   /////////////////////////////////////////////////////
 
   // copying all subtables over to otherMS
@@ -1050,6 +1053,11 @@ IPosition MSConcat::isFixedShape(const TableDesc& td) {
     itsMS.pointing().removeRow(delrows); 
   }
 
+  // SYSCAL
+  if(!copySysCal(otherMS.sysCal(), newAntIndices)){
+    log << LogIO::WARN << "Could not merge SysCal subtables " << LogIO::POST ;
+  }
+
 
   //////////////////////////////////////////////////////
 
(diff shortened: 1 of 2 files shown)


dpetry:
Looks good in principle. What kind of tests have you done so far?
The syscal concat was not needed on the ALMA side so far because each execution block is Tsys calibrated individually before any concat takes place.

gervandiepen:
Note that the Table constructor doing virtual concatenation has an argument making virtual concatenation of any subtable also possible (this was done with SYSCAL in mind).

kettenis:
The code has been tested on EHT data using the CASA importfitsidi, concat and virtualconcat tasks using two somewhat randomly chosen FITS-IDI files that have a different set of antenna's. I verified that importing those files together into a single MS and importing them separately and concatenating afterwards result in SYSCAL tables with the same number of entries and that plotcal shows entries that cover the expected time ranges for each antenna after running gencal to create a calibration table.
I had to cheat a bit with virtualconcat as the top-level Python code in CASA assumes the MSs that are concatenated contain a SOURCE table.

kettenis:
For VLBI data, copying the SYSCAL data is probably good enough since the data is in the TSYS column. For ALMA doing a virtual concatenation of the SYSCAL tables might make more sense as ALMA uses the TSYS_SPECTRUM column which means there may be significantly more data to copy. I can look into that a bit further. Assuming I get that to work, should that be handled in the same pull request?

dpetry:
concat is used a lot in ALMA data analysis and probably also VLA.
We need to verify that this still works. Did you run the concat unit test (NRAO CASA)?

kettenis:
The test_concat, test_virtualconcat, test_testconcat and test_importfitsidi unit tests all pass with this change.

gervandiepen:
I don't know what the CASA ms concatenate function is doing exactly.
The Tables concat has the functionality I described, but assumes that
non-concatenated subtables are the same (it takes them of the first ms).
Does CASA handle the different ANTENNA subtables correctly? If so, that's
nice.

dpetry:
The CASA concat task does a detailed merge of all subtables and the main 
table.
It makes no assumptions and reindexes all tables where needed.

PR #703, opened 2018-02-16
Allocator::INIT/NO_INIT and MCDoppler static state initialization prob

schiebel:
I encountered static initialization errors when trying to compile casacore object modules directly into a stand-alone executable (i.e. a stand-alone distribution of the casa viewer). The problems arise with the Array initialization constants and the initialization of MCDoppler state. These static are initialized properly when casacore is built into the various standard libraries (libcasa_casa, etc.). However, the initialization-before-use only happens because of this particular breakdown of the code into libraries. It falls apart when the object files are compiled into a single library or a single executable. It’s easy to see the issue, but there are a couple of solutions to the static issue in general:

1. convert the static to a static function which returns the static value

1. convert to constexpr static

1. add static initialization object (one per header file inclusion)

For the problems with MCDoppler, I opted for (3) because the initialization involves something complicated (filling an array). For the Allocator static initialization issues, I opted for (2) because it was really just initializing a static boolean. However, with the ArrayInitPolicy class it cannot have static ArrayInitPolicy constexpr members because they must be defined inline and ArrayInitPolicy is an incomplete class until the class declaration is completed. So I created an ArrayInitPolicies struct (I would prefer to create an ArrayInitPolicies namespace, but the designer of ArrayInitPolicies seems want to prevent others from creating policies (why I don’t really know exactly), but a namespace cannot be a friend of a class). This has the bad affect of changing the symbol space which provides INIT and NO_INIT (i.e. from ArrayInitPolicy to ArrayInitPolicies). However, I think the correct naming is more important than preventing downstream changes in this particular case.

The failure mode for the MCDoppler case was non-termination due to an infinite loop when doing the first conversion. The failure case for Allocator was more difficult, but the problem was discovered by printing "Allocator::INIT == Allocator::NO_INIT" which yielded "true" (with proper initialization this became false).

### DIFF
--- a/casa/Arrays/Array.h
+++ b/casa/Arrays/Array.h
@@ -196,7 +196,7 @@ public:
     // initialize elements before they are referred, especially when <src>T</src> is such type like <src>std::string</src>.
     // <srcblock>
     //   IPosition shape(1, 10);
-    //   Array<Int> ai(shape, ArrayInitPolicy::NO_INIT);
+    //   Array<Int> ai(shape, ArrayInitPolicies::NO_INIT);
     //   size_t nread = fread(ai.data(), sizeof(Int), ai.nelements(), fp);
     // </srcblock>
     Array(const IPosition &shape, ArrayInitPolicy initPolicy);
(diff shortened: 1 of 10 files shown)


mpokorny:
casacore is required to build under both C++11 and pre-C++11.

schiebel:
@mpokorny I added std=c++0x #ifdefs. However, the test times out for the ancient compiler... I guess this means this is ready to be merged.

mpokorny:
Looks like the build timeouts under pre-C++11 have reappeared.
You mentioned the failure modes for the two problems -- are there any tests in the test suite to verify the problems are fixed (and to avoid their reappearance)? Or would that require doing the build differently in the first place?

schiebel:
The problems arise if casacore is compiled into a single executable (i.e. no libcasa_casa, libcasa_tables, etc.). It is not practical to test this use case.

mpokorny:
Thanks Darrel, that's what I was thinking might be the case. I'll proceed with the merge.

schiebel:
Thanks for the quick response, Martin!

PR #704, opened 2018-02-21
Cas 11100 implement biweight stats algorithm and add support to LattStats for it

dmehring:
Also, make various stats framework methods virtual that should be virtual
change npts type from Int64 to uInt64 in a bunch of places
Add comprehensive unit tests for new BiweightStatistics class
Add unit tests to tLatticeStatistics to verify biweight algorithm works as expected there

### DIFF
--- a/casa/BasicSL/Constants.cc
+++ b/casa/BasicSL/Constants.cc
@@ -68,6 +68,8 @@ const Double C::gamma 		=  0.577215664901532860606512;
 const Double C::lngamma 	= -0.549539312981644822337662;
 const Double C::etogamma	=  1.7810724179901979852;
 
+// statistics related
+const Double C::probit_3_4 = 1.482602218505602;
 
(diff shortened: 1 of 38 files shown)


dmehring:
The CI failure appears to be due to a timeout; ie something beyond my control:

"The job exceeded the maximum time limit for jobs, and has been terminated."

dmehring:
Just a note that this is a high priority item for the CASA project. It is related to n-sigma clean.

mpokorny:
It looks like the failures in the non-c++11 builds are back. I can't _see_ any problems in your PR that would be responsible, so I think blaming the failure on the build system is reasonable.
That's a lot of changes! But it looks as if you have sufficient unit tests -- enough for me to proceed with the merge.

dmehring:
Thanks for the quick review Martin.

PR #710, opened 2018-03-13
Keep numpy array printing consistent for doctest with numpy v1.14 or above

zhengmeyer:
keep numpy 1.13 printoption for numpy 1.14 or above, otherwise doctest for tConvert will fail


tammojan:
Thanks!

PR #713, opened 2018-03-20
CAS-11195. Make LatticeApply::lineMultiApply more effecient at reading

dmehring:
add static method to FITSImage to be able to test for name of
image type without having to create an object to do it


mpokorny:
Is it OK to assume that these changes, functionally, are covered by the current test suite?

dmehring:
Hi Martin,
It is covered by tLatticeApply.cc and also by test_immoments in gcwrap.

dmehring:
Thanks Martin.

PR #714, opened 2018-03-22
CAS-11230 LCPolygon should include edge pixels where boundary is very close to zero

dmehring:


dmehring:
@gervandiepen Are you available to review and comment since I believe this may be your code?
Thanks

dmehring:
Thanks Ger.

PR #715, opened 2018-03-23
Cas 11195 need to initialize vectors with copies of arrays 

dmehring:
if done in constructors, all Arrays will reference the same array, which will screw things up for multiple output lattices.

### DIFF
--- a/lattices/LatticeMath/LatticeApply.tcc
+++ b/lattices/LatticeMath/LatticeApply.tcc
@@ -48,10 +48,6 @@
 #include <casacore/casa/Exceptions/Error.h>
 #include <casacore/casa/iostream.h>
 
-// debug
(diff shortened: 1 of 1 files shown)


dmehring:
@mpokorny Hi Martin,
This was exposed in a regression test when building the packages for the CASA PR. Unfortunately, it is a bit involved to write a casacore unit test to verify the fix, so I'm writing a CASA immoments unit test to at least expose it earlier in the development process should it become an issue again.
It is necessary to close out CAS-11195, which will improve immoments performance on the NRAO clusters, so its a high priority for folks using the pipeline, so if you could review at youir earliest convenience, that would be great.
Thanks

mpokorny:
Looks fine to me.

dmehring:
Thanks Martin.

Issue #580, opened 2017-03-16
Make C++11 required in casacore 3.0


tammojan:
I think it is time to make C++ a requirement for casacore. This will allow us to clean up the code base quite a lot, and make things easier, as discussed in #117. Also, it will facilitate making casacore more thread safe, by merging parts of #431.

Building with C++11 has been the default in casacore version 2.3. When we make it a requirement, I think we could also aim at releasing a 3.0 version, where the move to C++11 is the major change.

At ASTRON, the last cluster without support for c++11 will be upgraded in May. CASA already builds only with C++11. Are any of the other partners relying on support for older compilers? @mmarquar ?

cquike:
I would really be in favour of that! casacore could be a lot leaner if C++11 is used to its full power.

gervandiepen:
We are working on it.
First we have to ask the casacore users if they use older systems not supporting C++11 yet. I don't expect so, but we need to be sure. I'll do that coming week.
I want to use the call_once stuff once C++11 can be used.

tammojan:
@gervandiepen: have you heard back from other casacore users?

tammojan:
We've switched to C++11, closing this issue.

PR #716, opened 2018-03-27
Remove altogether the Vector cache for MVPosition and Euler classes.

cquike:
This PR removes altogether the Vector cache for MVPosition and Euler classes that was introduced
in #258 and modified in #680.
The cache was implemented using a thread_local storage, which means that the cache shouldn't be shared among different threads. This was unfortunately the case, since the MeasTable class creates some global objects that include instances of MVPosition derived classes. This causes
undefined behaviours, since the constructors for thread_local variables are called before the destructors for the global variables, according to the C++11 standard. Running the tTableMeasures test triggered in fact a segmentation fault due to this and valgrind reports those errors.

The original cache implementation didn't clean the memory, so the problem was simply hidden. The original idea of the cache was to improve the performance, achieving around 5x boost when objects are constantly being allocated and deleted. However, in a quick test I replaced Vector<Double> with std::array<double,3> (C++11) and I got a ~70x performance boost without any extra cache, so that might be a better route to explore, rather than a cache that is difficult to get it right.

### DIFF
--- a/casa/Quanta/Euler.cc
+++ b/casa/Quanta/Euler.cc
@@ -38,43 +38,14 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
 // Euler class
 
-// simplistic Vector(3) cache to reduce allocation overhead for temporaries
-#if defined(AIPS_CXX11) && !defined(__APPLE__)
-thread_local size_t Euler::available = 0;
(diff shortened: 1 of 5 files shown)


tammojan:
Thanks. Looks fine with me (and apparently no one objects). Note that by now we can assume having C++11 available (see #580), which should pave the path for better optimizations.

PR #717, opened 2018-03-28
Measures conversion of ephemerides table direction is ignoring the frame

kgolap:
The code in MCDirection setting up the conversion for ephemerides is hardwired hardwired to MDirection::APP. This is a fix for it. 
[meascomet_bug.py.txt](https://github.com/casacore/casacore/files/1857876/meascomet_bug.py.txt)

You can try the attached file (edit it to point to your data repository ephemerides) in any  casa and you'll see that it gets nearly 0.13 deg difference between the value given by DE405  and the ephemerides

*here is the output of the script in non-fixed casa*

Direct value for JPL table  18:44:52.203168 -023.02.14.394480

Using Measures me.direction("SUN") converted 18:44:52.202443 -023.02.14.390270

Using Measures me.direction("SUN") converted correcting for parallax 18:44:51.692250 -023.02.19.888463

Using measures  me.direction("COMET") converted  18:44:16.891719 -023.02.44.318668

Using measures me.direction("COMET") converted  correcting for parallax 18:44:16.381059 -023.02.49.809811

=
Once fixed ...the difference is of the order of arcsecs.

### DIFF
--- a/measures/Measures/MCDirection.cc
+++ b/measures/Measures/MCDirection.cc
@@ -33,6 +33,7 @@
 #include <casacore/casa/Arrays/ArrayMath.h>
 #include <casacore/measures/Measures/MCDirection.h>
 #include <casacore/measures/Measures/MeasFrame.h>
+#include <casacore/measures/Measures/MeasComet.h>
 #include <casacore/casa/Quanta/MVPosition.h>
 #include <casacore/measures/Measures/Nutation.h>
 #include <casacore/measures/Measures/MeasTable.h>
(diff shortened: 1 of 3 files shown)


kgolap:
Yes,  but the problem is a comet table may have posrefsys keyword but the value is APP ..and the default value when there is no posrefsys value set. So just in case the caller really needs to know if there was a posrefsys keyword defined or not that is the clear way.

mpokorny:
That is the case I was wondering about. Although I would prefer to leave out features that aren't used, I respect your judgment about what could be useful. I'll go ahead and merge this PR.

PR #733, opened 2018-06-13
DirectionCoordinate::convert() mods

dmehring:
* improve precision of angle returned and update tests accordingly
* avoid over the pole offsets
* improve API documentation


dpetry:
Looks good!

dmehring:
Thanks Dirk.

PR #734, opened 2018-06-18
TAQL leak fixing and Measures support

gervandiepen:
In case of exceptions TaQL could have some leaks due to its own reference counting. This has been fixed by using CountedPtr (shared_ptr) with its standard (atomic) reference counting. All code using it has been changed accordingly.

The TaQL Measures UDFs have been extended with support for MEarthMagnetic, MFrequency, MDoppler and MRadialVelocity. TableExprNode has been extended with attributes (a Record), so a UDF can recognise a nested UDF and its type.

The doxygen documentation of all code has been built. A few errors have been fixed.


tammojan:
You added several measures modules to TaQL in this PR, is this intended?

mpokorny:
Is there any chance of reverting the commits in this PR? These commits changed the casacore API, which resulted in breaking the CASA build. We should try to coordinate better on such changes.

mpokorny:
Thanks, @tammojan.

mpokorny:
@gervandiepen, the issue regarding the CASA build was reported to me. I have not yet had the time to investigate, but I will do so today. Thanks for the pointer about what has and hasn't changed; since the commits affect many files, it'll make my job easier. I'll send you the build log if I'm unable to determine what the problem is, or don't see the way to handle it in CASA.

mpokorny:
`ExprNodeRep` is indeed used by a single class in a method that traverses a `TableExprNode` in order generate a text description of the expression. It's done in a class used for plotting table data, and off-hand it seems a reasonable usage. Perhaps there's an alternative, but no method of `TableExprNode` jumped out at me.

I was able to convert the single method in CASA to the new interface, but I should probably have someone else on the CASA team test out my changes, and then we can decide on a schedule to re-apply these commits to the master branch.

gervandiepen:
Thanks. It's probably best to add that method to the TableExprNode
interface. Can you tell me which function is used?

PR #735, opened 2018-06-19
Align the unit for atmospheric pressure with documentation

tnakazato:
In [MSv2 design document](http://casacore.github.io/casacore-notes/264), the unit for atmospheric pressure is defined as
"hPa". However, implementation was different (Pa). This commit fix the
unit to be consistent with the documentation.

cf. https://open-jira.nrao.edu/browse/CAS-11528

### DIFF
--- a/ms/MeasurementSets/MSWeather.cc
+++ b/ms/MeasurementSets/MSWeather.cc
@@ -157,7 +157,7 @@ void MSWeather::init()
 		  "Flag for average column density of electrons","","");
 	// PRESSURE
 	colMapDef(PRESSURE, "PRESSURE", TpFloat,
-		  "Ambient atmospheric pressure","Pa","");
+		  "Ambient atmospheric pressure","hPa","");
 	// PRESSURE_FLAG
 	colMapDef(PRESSURE_FLAG, "PRESSURE_FLAG", TpBool,
 		  "Flag for ambient atmospheric pressure","","");
--- a/msfits/MSFits/MSFitsOutput.cc
+++ b/msfits/MSFits/MSFitsOutput.cc
@@ -2538,8 +2538,8 @@ Bool MSFitsOutput::writeWX(FitsOutput *output, const MeasurementSet &ms) {
             *temperature = 0.0;
         }
         if (hasPressure) {
-            //covert from Pa to mbar
-            *pressure = weatherColumns.pressure()(i) / 100;
+            //covert whatever units to mbar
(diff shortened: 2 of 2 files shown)


tnakazato:
I think the code filling the WEATHER table is outside the casacore code (e.g. asdm2MS). I intend to check those codes in the context of CASA development (CAS-11528). Actually, asdm2MS overwrites pressure unit to hPa so it should be no harm. I plan to check other filler code. 

If there is a filler code inside casacore, that should be checked. But I don't know if it exists.

tnakazato:
It seems casacore/msfits converts FITS into MS (and vice versa). I need to check the code.

tammojan:
There are some more references to the pressure unit in casacore, particularly when converting to/from UVFits (found by `grep -ri pressure casacore`):
https://github.com/casacore/casacore/blob/e10fff37817ce2281bb076881f15c6e7ebf74fcd/msfits/MSFits/MSFitsOutput.cc#L2541-L2542
https://github.com/casacore/casacore/blob/e10fff37817ce2281bb076881f15c6e7ebf74fcd/msfits/MSFits/FitsIDItoMS.cc#L3246
https://github.com/casacore/casacore/blob/e10fff37817ce2281bb076881f15c6e7ebf74fcd/scimath_f/atmroutines.f#L240-L242

tnakazato:
Thanks Tammo Jan. I looked at the files you mentioned.

### msfits/MSFits/MSFITSOutput.cc
I believe I've rewritten the conversion of pressure value into the value in mbar more correctly. That should work for whatever pressure units while the previous one only worked for Pa. 

### msfits/MSFits/SDWeatherHandler.cc
I don't think I need any update on this file. Furthermore, the class is not referred from anywhere in casacore nor in casacode. 

### scimath_f/atmroutines.f
I don't think I need any update on this file as long as the subroutine is correctly used (i.e. input pressure value is in mbar). Like SDWeatherHandler, however, the subroutine is not referred from anywhere in in casacore nor in casacode.

I tested my change on CASA using existing ALMA data. I exported the data to UVFITS format using exportuvfits task. I found that the old code doens't handle pressure conversion correctly since it blindly divide the input value by 100 regardless of what the pressure unit is. On the other hand, new code I committed handles the conversion right. So, I believe the change is OK. 

I also tried re-importing UVFITS data with weather data to MS using importuvfits task. However, it seems that the task just ignores UVFITS WX table.

PR #739, opened 2018-06-22
CAS-11554 add support for Double and DComplex lattices to LatticeAddNoise

dmehring:


dmehring:
@mpokorny Hi Martin, would you please review this one? I think it should be pretty straight forward. Thanks.

lattices/LatticeMath/LatticeAddNoise2.tcc
```diff
@@ -0,0 +1,56 @@
...
+#include <casacore/casa/BasicSL/Complex.h> 
+#include <casacore/casa/BasicMath/Random.h> 
+#include <casacore/casa/BasicSL/String.h>
+
+namespace casacore { //# NAMESPACE CASACORE - BEGIN
+ 
+template <class T> void LatticeAddNoise::add (MaskedLattice<T>& lattice) {
+    ThrowIf(! itsNoise, "You have not yet called function 'set'");
+    LatticeIterator<T> it(lattice);
+    for (it.reset(); !it.atEnd(); it++) {
```
mpokorny: A minor issue, but it might be a bit better to pre-increment rather than post-increment `it` here to avoid a potential copy of LatticeIterator.

dmehring: Thanks Martin,
According to the docs, the pre- and post-fix operators do exactly the same thing:

  // Increment operator - increment the cursor to the next position.  These
  // functions are forwarded to the current LatticeNavigator and both
  // postfix and prefix versions will do the same thing.
  // <br>They return True if the cursor moved (which should always be the
  // case if the iterator is not at the end).
  // <group>
  Bool operator++();
  Bool operator++(int);
  // </group>

dmehring: And in LatticeIterInterface, which that LatticeIterator methods call:

  // Increment operator - increment the cursor to the next position. The
  // implementation of the prefix operator calls the postfix one.
  // <group>
  Bool operator++();
  Bool operator++(int);
  // </group>

PR #741, opened 2018-06-26
Catch bad_alloc and exit instead of throwing an exception

vsuorant:
Currently Casa crashes when trying to load very large data sets. I think it makes more sense to report the memory allocation exception and exit.

### DIFF
--- a/casa/Containers/Allocator.h
+++ b/casa/Containers/Allocator.h
@@ -147,14 +147,21 @@ struct casacore_allocator: public std11_allocator<T> {
   }
 
   pointer allocate(size_type elements, const void* = 0) {
-    if (elements > this->max_size()) {
-      throw std::bad_alloc();
-    }
     void *memptr = 0;
-    int result = posix_memalign(&memptr, ALIGNMENT, sizeof(T) * elements);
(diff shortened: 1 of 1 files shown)


vsuorant:
One point I failed to mention in the original message is that the bad_alloc is caused by insufficient physical memory. So the very large data sets work fine as long as they have very large physical memory to work with. 

Casa crashes with sigabrt when there is no more memory available. I believe this is because nothing catches the bad_alloc. I was trying to figure out a right place to handle this and it seems that it would be good to handle this at the source of the problem rather than pass the exception downstream.

mpokorny:
I tend to agree with Ger on this. The end result of an uncaught exception or a call to exit() is more or less the same, namely the program stops running immediately. Whether it's a sigabrt or not is sort of beside the point (in fact, for debugging sigabrt may be better as one could get a core file). At least an exception _could_ be handled, which would allow other outcomes. I understand that it is difficult to determine where in casa the exception(s) should be handled, but I don't find that convincing enough to use this alternative approach, which cuts off all other alternatives.

vsuorant:
Ok. I will file a Casa ticket on this then.

Issue #729, opened 2018-06-08
Enhancing parallel IO in Casacore Table Data System


JasonRuonanWang:
I don't know if this is the best place but I was looking to some discussions on enhancing parallel IO for the table system as well as Measurementsets. 

A few years ago I wrote a parallel storage manager for the casacore table system based on ADIOS (https://github.com/SKA-ScienceDataProcessor/AdiosStMan.git) when I was at ICRAR, which was put into the SKA-SDP as an early investigation into SKA scale measurementset experiments. When @gervandiepen was visiting us at ICRAR we talked about merging this AdiosStMan into the master but I was worried about the code quality because it was only a prototype rather than a production software. @gervandiepen was also worried about introducing MPI dependency to casacore. So eventually we dropped the plan. 

Recently I have been working on updating the Adios Storage Manager using our new software ADIOS 2.0 (https://github.com/ornladios/ADIOS2) that we are developing at Oak Ridge National Lab under the US Department of Energy's Exa-scale Computing program. The motivation behind this is that we are planning on running a full-scale SKA 1 data pipeline simulation, which may require writing data into a single measurementset table from multiple processes. If this is successful then we can re-think about improving the code quality and merge our efforts into the master.

However, the current parallel storage manager design is more like a hack than a generic solution because we couldn't find parallel support in casacore's table layer API. As a result, we need to hack user applications and make them believe that each MPI rank is writing to its own independent table, while the Adios Storage Manger does dirty tricks underneath and actually writes data into a single column file. This has worked so far but it makes user applications very ugly and less maintainable. 

I have roughly read through the table layer code trying to think of a solution. If we can add some MPI-aware code in the procedures of creating a table, making only Rank 0 actually write table metadata file, while other ranks hold the table metadata in memory without writing to files, this would make the table system much more friendly to parallel storage managers as well as MPI incorporated applications. 

I am wondering if this is possible or if there is any other suggestions or plans on enhancing parallel IO for casacore. Looking forward to any discussions or suggestions. Thank you.

JasonRuonanWang:
There is another approach that I can think of, which is adding another mode in TableLock::LockOption. This mode does the follow: Any process comes first owns the table lock, while if the table is already locked, other processes skip writing the table metadata file. This will also work for any MPI based parallel IO storage managers or applications. And the advantage of doing so over the previous approach is that this would not require introducing MPI dependency into casacore. 

Any suggestions?

tammojan:
[...]

PR #742, opened 2018-06-26
Added MPI support for parallel CTDS I/O

JasonRuonanWang:
See discussions in #729 

This PR added MPI support for the casacore table data system. Previously we can partially achieve parallel I/O by introducing a parallel storage manager (e.g. https://github.com/SKA-ScienceDataProcessor/AdiosStMan). However, the table layer API was strictly serial for writing. In this PR, I added MPI-awareness to the table layer API. The logic is to write the table files only on Rank 0, while other ranks do not write any data. This fixed the issue that slave ranks of user applications have to pass in garbage tables with the table API to avoid conflicts when writing table files. 

To use this feature, one must configure casacore using the following flags

-DUSE_MPI=ON 
-DENABLE_TABLELOCKING=OFF

Also, one must use a parallel CTDS storage manager. Some of the existing parallel storage managers are:
https://github.com/SKA-ScienceDataProcessor/AdiosStMan
https://github.com/JasonRuonanWang/Adios2StMan
https://github.com/JasonRuonanWang/Hdf5StMan

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,6 +73,7 @@ option (USE_HDF5 "Build HDF5 " NO)
 option (USE_FFTW3 "Use FFTW instead of FFTPack" NO)
 option (USE_THREADS "Use Mutex thread synchronization" NO)
 option (USE_OPENMP "Use OpenMP threading" NO)
+option (USE_MPI "Use MPI for parallel IO" NO)
 option (USE_STACKTRACE "Show stacktrace in case of exception" NO)
 option (CASA_BUILD "Building in the CASA (http://casa.nrao.edu) environment" NO)
 
@@ -408,6 +409,10 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel")
    set(CASACORE_ARCH_LIBS ${CASACORE_ARCH_LIBS} -L${INTEL_LIB_PATH} -limf -lsvml -lirng -lintlc -lifport -lifcore -liomp5)
 endif()
 
+if(USE_MPI)
+    find_package(MPI COMPONENTS C REQUIRED)
+endif(USE_MPI)
+
 if(USE_OPENMP)
     set (USE_THREADS YES)
     find_package (OpenMP)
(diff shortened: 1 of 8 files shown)


gervandiepen:
Hi Jason,
It looks nice. It's smart that you use openedForWrite() to not write the meta info for MPI ranks > 0.
I assume you have tested it with the Adios2StMan. Did you get good results?
The only thing you possibly could improve on is that the code in the BaseTable and PlainTable constructors you cloned is put in a single shared function.
Cheers, Ger

JasonRuonanWang:
@gervandiepen Yes sir! I have put the common code for the constructor in a BaseTableCommon() and PlainTableCommon() and call them from the MPI and non-MPI constructors respectively. The reason why I didn't do this is because I wasn't sure if I need to change the procedures in those. But since I found the openedForWrite() trick I believe I probably won't need to touch the constructors any more.

I am still working on Adios2StMan, but I have tested it using AdiosStMan and Hdf5StMan. Everything is working as expected, which is not surprising because this table layer change does not affecting how parallel storage managers work. It just provides a nicer way to write parallel applications and makes them fully compatible with serial applications. I will let you know once I finish Adios2StMan and the performance testing.

gervandiepen:
Great work sir! I'll take a look at your new StMans.
I have still one question left. How did you solve the addRow problem which
(in principle) has to be done before you can write a new row? Is it ignored
in the parallel StMans?

Do you have any idea how much overhead the table system causes compared to
raw Adios or HDF5?

JasonRuonanWang:
@gervandiepen I haven't pushed Adios2StMan to the repo yet but it wont't take too long time. 

For addRow what I did previously in AdiosStMan is to disable that. But if it's a requirement then it shouldn't be hard to enable it in Adios2StMan. All this kind of table operations, I will only do it on Rank 0, since that's the rank which actually holds the table metadata. So from the user application's perspective, it can either call addRow on Rank 0 or call it on all ranks, which won't make difference. 

But from what I understand most of the use cases including MeasurementSets, it's not difficult to determine a total number of rows beforehand. So I wouldn't encourage applications to change table properties frequently which for sure will harm the performance. 

For the overhead issue, I haven't compared between the CTDS table system and ADIOS/HDF5, because for everything I have done, table metadata is still stored in CTDS/AipsIO. But if you mean the column level, then if I remember ADIOS1 takes a little more overhead than the standard storage manager. I think it's because it needs to manage the meta data information for every rank.

mpokorny:
I confess to not thinking too much about the details of whether this code would be sufficient to support the work I've been doing, but it does address the first major design issue in the table system code that I encountered. These changes are certainly necessary for what I'd like to do, and the implementation looks fine.

JasonRuonanWang:
@mpokorny Thanks for the follow-up. I believe this is the base functionality that any parallel CTDS IO is likely to need. In the meantime we can certainly keep discussing what are the common things we will need for the next step.

tammojan:
Merging this; future extensions can go into a new PR.

tammojan:
Thanks for the great work so far, @JasonRuonanWang !

PR #746, opened 2018-06-27
CAS-11565 templated methods to support Double/DComplex in LatticeFFT

dmehring:


lattices/LatticeMath/LatticeFFT.h
```diff
@@ -64,48 +66,66 @@ template <class T> class Lattice;
...
   // Non-folded version
-  static void cfft0(Lattice<Complex> & cLattice,
+    template <class ComplexType> static void cfft0(Lattice<ComplexType> & cLattice,
   		  const Vector<Bool> & whichAxes, const Bool toFrequency=True);
 
   // N-D in-place complex->complex FFT. Transform over all axes.
-  static void cfft(Lattice<DComplex> & cLattice, const Bool toFrequency=True);
-  static void cfft(Lattice<Complex> & cLattice, const Bool toFrequency=True);
+    template <class ComplexType> static void cfft(
+        Lattice<DComplex> & cLattice, const Bool toFrequency=True
```
mpokorny: Does the cLattice argument have the correct type here? Or do you even need both this method and the following one?

dmehring: Good find. Looks like I missed that one and also a couple more. I'll fix that up. Thanks.

PR #751, opened 2018-07-06
CAS-11613 templatize method to support stats for Double-valued lattices

dmehring:


dmehring:
Thanks for the quick review Martin.

PR #754, opened 2018-07-10
Cas 11632 enable support for Double valued lattices in LatticeHistograms

dmehring:


dmehring:
Note that I merged the branch from PR 752 into this branch, so PR 752 should be merged before this one is reviewed. Thanks.

PR #755, opened 2018-07-11
Fix addColumn in ConcatTable

### LINKED ISSUE #753: Table::addColumn doesn't work with Multi-MS (ConcatTable, ConcatColumn)
This is for CASA issue: https://open-jira.nrao.edu/browse/CAS-11243 ("Fix failure in Table::addColumn with Multi-MS").

This affects at least the calibrater tool, tclean, and sdcal, when processing Multi-MSs.

Addition of scratch columns is not working well for Multi-MSs. In casa/code/msvis/MSVis/VisSetUtil.cc:VisSetUtil::addScrCols() two calls to TableCopy::cloneColumnTyped<>() and TableCopy::fillColumnData() produce an exception for ConcatTables (in the second method, because the first one doesn't work well).

I think I have a fix for this which seems to work and makes sense.
ConcatTable is not updating correctly the colMap_p when columns are added. It needs to update it, similarly as RefTable does.
Then, after fixing this, tests with the calibrater tool also uncovered that ConcatColumn needs tileShape(), which is used by TableCopy::fillColumnData().

FedeMPouzols:
Fixes #753.

See issue for a description of the problem. I've extended tConcatTable to catch this kind of issue (but I'm not totally sure I'm following the casacore test conventions properly). 
In CASA, this fixes the issue for the calibrater tool and uvcontsub (for which we had an ugly workaround) and tclean (for which we don't have a workaround).

### DIFF
--- a/tables/Tables/ConcatColumn.cc
+++ b/tables/Tables/ConcatColumn.cc
@@ -119,6 +119,13 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
     return refColPtr_p[tableNr]->shape (tabRownr);
   }
 
+  IPosition ConcatColumn::tileShape(uInt rownr) const
+  {
+    uInt tableNr, tabRownr;
+    refTabPtr_p->rows().mapRownr (tableNr, tabRownr, rownr);
+    return refColPtr_p[tableNr]->tileShape (tabRownr);
+  }
+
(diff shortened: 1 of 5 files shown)


gervandiepen:
Very nice. Thanks.
Ger

PR #757, opened 2018-07-12
Cas 11643 Fit2D method to enable casa imfit for Double images

dmehring:


dmehring:
Note that I merged the branch in PR 754 into this branch.

mpokorny:
Creating branches that include other unmerged branches makes the review more difficult, as previously reviewed changes are mixed in with new changes when reviewing the PRs in order. It might have been better in this case to create just one PR branch and push changes to it a few times. I guess that you might have chosen to do it this way since I was away for a week, although I'm not sure of the benefit. Anyway, the changes look good.

dmehring:
Thanks Martin. It was the result of how I was working the corresponding branches in CASA, each branch building on the previous, although I suppose I could have created a single composite branch locally and pushed the individual branches separately, although each separate commit amounted to a different branch anyway. But I will keep in mind for the future.

PR #759, opened 2018-07-20
MSHistoryHandler: don't flush the HISTORY subtable for every addMessage()

### LINKED ISSUE #758: MSHistoryHandler flushes the HISTORY subtable for every addMessage() - bad performance
The addMessage methods of MSHistoryHandler do `histTable_p.flush()` at the end. This has very bad performance implications for CASA tasks when they write the history.

When trying to address unexpected delays at the end of CASA tasks when they update the HISTORY subtable, we noticed that even if one tries to write several messages in sequence, there will be a flush for every row.
This can cause a delay of 10-30 seconds at the end of CASA tasks like flagdata. CASA tasks write one message/row per input parameter, which for tasks like flagdata implies 70+ rows are added. Every flush seems to take at least of the order of a 10th of a second.

The addMessage static method flushes at the end, and that looks reasonable as it seems designed for single message write operations. But I think that the non-static methods (at least the one with signature `MSHistoryHandler::addMessage(String message, String cliComm,String origin)`
should not flush on every message added, and defer that to the destructor. I'm preparing a patch for this.

CASA issue: https://open-jira.nrao.edu/browse/CAS-11432.

FedeMPouzols:
Fixes #758.

This PR modifies the non-static addMessage() methods to not do a flush(). This is deferred to the destructor.
The static addMessage() still does flush(), and that makes sense to me as it is a one-off message add operation.

### DIFF
--- a/ms/MeasurementSets/MSHistoryHandler.cc
+++ b/ms/MeasurementSets/MSHistoryHandler.cc
@@ -35,14 +35,14 @@
 
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
-MSHistoryHandler::MSHistoryHandler(MeasurementSet& ms, String app){
+MSHistoryHandler::MSHistoryHandler(MeasurementSet& ms, const String& app){
   histTable_p = ms.history();
   msHistCol_p= new MSHistoryColumns(histTable_p);
   application_p=app;
 }
 
 
-MSHistoryHandler &MSHistoryHandler::operator=(MSHistoryHandler& other){
+MSHistoryHandler &MSHistoryHandler::operator=(const MSHistoryHandler& other){
 
   if(this != &other){
     histTable_p=other.histTable_p;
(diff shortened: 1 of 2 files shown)


mpokorny:
This seems OK to me on its own, but I don't have a good understanding of the context -- is the history table a special case, or does this PR change some assumptions about history table updates made elsewhere in casacore? I would prefer @gervandiepen to comment, but I think that he's on vacation at the moment. Perhaps @tammojan should review this PR as well.
That said, it does appear that the commit contains two sorts of changes: one being the changes to the flushing behavior, and the other being the addition of const modifiers in several places. I think that we'd prefer the changes to be split into two commits, but I'd like to get @tammojan 's opinion on that as well.

FedeMPouzols:
I'd also prefer to split the const related changes, but the notes in CONTRIBUTING.md seem to suggest that a smaller number of commits is preferred. So I put it all in one commit considering that the const changes don't imply any change in functionality and don't break anything, as only more constness is added, not removed.

FedeMPouzols:
I've created an alternative PR: #760, with two commits, so that both alternatives would be ready to be merged assuming everything else is fine.
I also noticed that the MSHistoryHandler class doesn't have any test and no other tests seem to be testing it.

PR #763, opened 2018-07-30
CAS-11694 refactor/reformat/C++11 changes only

dmehring:


mpokorny:
Looks good. Since I'm a fan of replacing loops with functions from \<algorithm\>, I was happy to see that you've done that a bit. I think that using explicit rather than explicit lambda captures is generally considered good practice, as it helps avoid errors, but I know that it can be inconvenient...something to consider.

PR #764, opened 2018-08-01
CAS-11711 reformat/refactor/C++11 upgrades, no functionality changes

dmehring:


dmehring:
@mpokorny Martin, I have more changes along these same lines on another branch. Would you like to review the current branch as is and I'd push the newer branch for review separately, or would you like me to merge my more recent changes into this branch and then review everything (or just discard this branch and I'd push the new one that incorporates these changes which would still amount to you reviewing essentially the changes on both branches)?

mpokorny:
Although these sorts of changes are not difficult to review, they are voluminous. If your second branch depends on this one, you should probably just merge the two since they aren't independent. If they are independent, then I'd prefer that you keep them separate just to reduce the code I have to review at one time.

dmehring:
I probably should just merge the newer branch back into this one then, since the newer one was created from this branch so I could build upon these changes. I'm hoping this will be the last batch of changes of this sort for awhile.

dmehring:
I've created PR 766 which supersedes the current PR which I'll remove in a minute. I just created a new branch and committed all the changed files there so there wouldn't be multiple commits for you to wade through. Hopefully that will make things a little easier than if there are a bunch of commits for which some changes from previous commits might be redone in later commits.

Issue #767, opened 2018-08-17
ScalarRecordColumnDesc increasing table size on rewrite of idenitcal data


dmehring:
I reported this in the CASA JIRA system, but it appears to have been orphaned.

https://open-jira.nrao.edu/browse/CAS-10737

When overwriting data in a ScalarColumn<TableRecord> with the exact same data, the size of the associated table file increases. I would expect the file size to remain the same. Here is code that demonstrates the issue, followed by the output. The contents of the table cell and the number of rows remain unchanged as expected, but the file size grows after each update. Can you tell me if this is this expected behavior or a bug?

```
void update(const String tabName, const String& colName, const TableRecord& rec) {
    Table tab(tabName, TableLock(TableLock::AutoLocking), Table::Update);
    ScalarColumn<TableRecord> recCol;
    recCol.attach(tab, colName);
    recCol.put(0, rec);
    TableRecord rec1;
    recCol.get(0, rec1);
    cout << "nrow " << tab.nrow() << endl;
    cout << "rec1 " << rec1;
}


String tabName = "x.tab";
String colName = "Test";
TableRecord rec;
rec.define("j","x");
{
    // set up a new table
    TableDesc td;
    td.addColumn(ScalarRecordColumnDesc(colName));
    SetupNewTable newtab(tabName, td, Table::New);
    Table tab(newtab, TableLock::AutoLocking, 1, false);
    ScalarColumn<TableRecord> recCol;
    recCol.attach(tab, colName);
    recCol.put(0, rec);
    TableRecord rec1;
    recCol.get(0, rec1);
    cout << "nrow " << tab.nrow() << endl;
    cout << "rec1 " << rec1;
}
File file(tabName + "/table.f0i");
cout << "size " << file.size() << endl << endl;
for (uInt i=0; i<10; ++i) {
    // overwrite the record with the identical record multiple times,
    // and note how the file size grows
    update(tabName, colName, rec);
    cout << "size " << file.size() << endl << endl;
}
```

```
 nrow 1
rec1 j: String "x"
size 99

nrow 1
rec1 j: String "x"
size 187

nrow 1
rec1 j: String "x"
size 275

nrow 1
rec1 j: String "x"
size 363

nrow 1
rec1 j: String "x"
size 451

nrow 1
rec1 j: String "x"
size 539

nrow 1
rec1 j: String "x"
size 627

nrow 1
rec1 j: String "x"
size 715

nrow 1
rec1 j: String "x"
size 803

nrow 1
[...]

PR #768, opened 2018-08-21
Take care similary shaped data overwrites existing data

### LINKED ISSUE #721: MS growth hormone with StandardStMan
I have been (mistakenly) inserting visibility data columns using a StandardStMan rather than a TiledStMan, using the following pattern:

```
        if col_name not in self.ms.colnames():
            desc = self.ms.getcoldesc(like_col)
            desc['name'] = col_name
            self.ms.addcols(desc)
```

I now realize ``dminfo`` also needs to be specified to make a proper tiled column.

However, there is a mysterious side effect: if I now repeatedly write visibility data to such a (StandardStMan-managed) column -- overwriting existing rows, not adding new rows -- the size of the MS on disk continuously grows. It seems every new overwrite of the column causes one of table.fxx files inside the MS to grow in size. 

The problem goes away with a tiled column. Still, surely this is not normal behaviour for StandardStMan?

gervandiepen:
With the StandardStMan the indirect array file (e.g., table.f0i) grew if an array (or record) with the same shape was written. It has been fixed by taking care the existing shape is known.\

close #721 #767

### DIFF
--- a/build-tools/casacore_memcheck.supp
+++ b/build-tools/casacore_memcheck.supp
@@ -1,4 +1,5 @@
 # This file contains the valgrind errors to be suppressed for casacore.
+# Where needed, it has lines for namespace casa and casacore.
 
 # Very often libc gives an invalid free error at exit.
 {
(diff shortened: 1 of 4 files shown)


gervandiepen:
When running the test through valgrind, it appeared that some innocent errors were not suppressed if namespace casacore was used. Lines have been added to suppress them.

PR #772, opened 2018-08-29
Change HDF5Object hid type to match hdf5 headers.

### LINKED ISSUE #771: Undefined reference in HDF5Group caused by use of casacore Int64
I am getting a linker error when trying to compile some code using the HDF5 classes from casacore, though a simple program is enough to demonstrate the problem on my RHEL7 system.

```c++
#include <casacore/casa/HDF5/HDF5Group.h>

int main() {
  casacore::HDF5Object object;
  casacore::HDF5Group group(object, "0");
  (void)group;

  return 0;
}
```
Casacore built with cmake arguments `-DUSE_FFTW3=ON -DUSE_HDF5=ON -DUSE_THREADS=ON`, program compiled as:

```
$ g++ -o test test.cc -lcasa_casa
/tmp/ccKzWEgs.o: In function `casacore::HDF5Group::HDF5Group(casacore::HDF5Object const&, casacore::String const&, bool, bool)':
test.cc:(.text._ZN8casacore9HDF5GroupC2ERKNS_10HDF5ObjectERKNS_6StringEbb[_ZN8casacore9HDF5GroupC5ERKNS_10HDF5ObjectERKNS_6StringEbb]+0x7f): undefined reference to `casacore::HDF5Group::init(long long, casacore::String const&, casacore::String const&, bool, bool)'
collect2: error: ld returned 1 exit status
```
The issue boils down to the use of `Int64` in HDF5Object's hid union. On my system, `int64_t` is defined as simply `long`, whereas Int64 is `long long`. This causes a mismatch in the types when hdf5's `hid_t` definition is used, since it is also using `int64_t`. (`hsize_t` should not be affected since it is explicitly defined as `unsigned long long` in H5public.h.) With the changes below, everything works as expected.
```diff
diff --git a/casa/HDF5/HDF5Object.h b/casa/HDF5/HDF5Object.h
index 4313bad0f..08e2d04d2 100644
--- a/casa/HDF5/HDF5Object.h
+++ b/casa/HDF5/HDF5Object.h
@@ -38,7 +38,7 @@
 #ifdef HAVE_HDF5
 # include <hdf5.h>
 #else 
-  typedef casacore::Int64 hid_t;
+  typedef int64_t hid_t;
   typedef casacore::uInt64 hsize_t;
 #endif
 
@@ -122,7 +122,7 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
     //# for older HDF5 versions where hid_t is a 32 bit integer.
     union {
       hid_t  itsHid;
-      Int64  itsDummyHid;
+      int64_t  itsDummyHid;
     };
     String itsName;
```

I can issue a pull request with these changes if desired.

patbrandt:
Fixes #771.
Casacore defines it's own 64-bit integer type, Int64, in casacore/casa/aips.h as 'long long'. The HDF5 library defines it's hid_t as int64_t from <stdint.h>. On different systems int64_t can be 'long long' or simply 'long', and in the latter case the compiler cannot match HDF5Group::init's arguments for the constructor that uses implicit conversion from HDF5Object to hid_t.

### DIFF
--- a/casa/HDF5/HDF5Object.h
+++ b/casa/HDF5/HDF5Object.h
@@ -38,7 +38,7 @@
 #ifdef HAVE_HDF5
 # include <hdf5.h>
 #else 
-  typedef casacore::Int64 hid_t;
+  typedef int64_t hid_t;
   typedef casacore::uInt64 hsize_t;
 #endif
 
@@ -122,7 +122,7 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
     //# for older HDF5 versions where hid_t is a 32 bit integer.
     union {
       hid_t  itsHid;
-      Int64  itsDummyHid;
+      int64_t  itsDummyHid;
     };
     String itsName;
 


gervandiepen:
Thanks for the fix.

PR #776, opened 2018-09-06
CAS-11855 more robust bin index search

dmehring:


tammojan:
Have you successfully viewed travis logs before? I.e. is it just broken now, or do you need to set up the authentication etc.?

I can see the log: only one of the 4 builds fails, namely the MacOS python3 variant. There's a missing include `boost/numarray.h` or so, so likely unrelated to your change.

dmehring:
As of a couple of months ago, I was able to see logs, yes. I'm not certain what has changed in the interim or what I need to do so I can see them again. My changes do not depend on that include, but I'm curious as to why CI does not fail on the master branch (from which I created my branch) in that case.
Thanks

dmehring:
Looks like a possible browser issue. I restarted my browser, and now I can see the logs. Weird.

tammojan:
The missing include was `numpy/arrayobject.h` in python/Converters/PycArrayNP (and probably other files too). I have no idea why this has stopped working, perhaps Travis has updated their MacOS environment. In that case, the next build of the master will fail as well...

Since the failing build is not related to this PR, I suggest to not let it block it.

dmehring:
@schiebel Hi Darrell, when you have a chance, would you please review and merge this one? The CI failure is due to an unrelated issue. @tammojan has noted that the current PR should not be blocked because of that failure.
Thanks

schiebel:
@dmehring is std::set actually used?

dmehring:
@schiebel Good catch. It was temporarily (and may be again in the future) but I forgot to remove the include. Change pushed.

schiebel:
OK, thanks @dmehring after travis-ci finishes, I'll merge this change...

PR #782, opened 2018-10-02
Added Adios2StMan and test with writems and readms

JasonRuonanWang:


gervandiepen:
Hi Jason,
Thanks for the PR.
I noticed there is no test program. Would it be hard to add one?

Looking at Adios2StMan.cc I get the impression that MPI is not mandatory when using Adios2StMan. Yet, it is only registered if HAVE_MPI is defined. Is that really needed? In other words, could I use Adios2StMan on a single machine without MPI?

JasonRuonanWang:
@gervandiepen @tammojan 

This is pretty much finished. I have added Adios2StMan and a test for it, and also changed ms/apps/writems and readms to add an option for using Adios2StMan. All of these are verified to work correctly. Could you have a look and let me know if there is anything else I need to do? Thanks.

PR #784, opened 2018-10-04
CAS-11929 added leading casacore to include path

dmehring:


dmehring:
It looks to me like there was an issue unrelated to my change during CI. My change is pretty simple, and as long as things compile (which they do), all tests should pass, assuming they all passed before my change. I think this can be merged.

PR #785, opened 2018-10-04
CAS-11932 bug fix of histogram index finding code

dmehring:


dmehring:
Again, this looks like a false failure. It would be good if the CI environment or whatever is the root cause here could be fixed. Or this particular CI run should just be removed. It is not useful at this point.

tammojan:
I disabled the failing CI build in #786 (it was failing because of #778).

PR #786, opened 2018-10-07
Travis: allow python3 on OSX to fail

tammojan:


tammojan:
See #778 for the status of the Travis OSX python3 build.

PR #787, opened 2018-10-23
Add support for extra optional columns in the spectral window subtable.

cquike:
Some correlators, including ALMA and VLA in WIDAR mode, include a functionality
to apply a windowing function with a number of bins. This information,so far,
hasn't been readly available to CASA.
This patch adds support for two extra optional columns in the spectral window
subtable, namely SDM_WINDOW_FUNCTION and SDM_NUM_BIN. See CASA ticket CAS_11547
(and tickets mentioned therein) for a background.

Fixes partially CAS-11547

This has high priority since it is blocking the release of CASA 5.4.1

### DIFF
--- a/ms/MeasurementSets/MSSpWindowColumns.cc
+++ b/ms/MeasurementSets/MSSpWindowColumns.cc
@@ -93,6 +93,8 @@ ROMSSpWindowColumns(const MSSpectralWindow& msSpWindow):
   bbcSideband_p(),
   dopplerId_p(),
   receiverId_p(),
+  sdmWindowFunction_p(),
+  sdmNumBin_p(),
   chanFreqMeas_p(msSpWindow, MSSpectralWindow::
 		 columnName(MSSpectralWindow::CHAN_FREQ)),
   refFrequencyMeas_p(msSpWindow, MSSpectralWindow::
@@ -320,6 +322,8 @@ ROMSSpWindowColumns::ROMSSpWindowColumns():
   bbcSideband_p(),
   dopplerId_p(),
   receiverId_p(),
+  sdmWindowFunction_p(),
+  sdmNumBin_p(),
   chanFreqMeas_p(),
   refFrequencyMeas_p(),
   chanFreqQuant_p(),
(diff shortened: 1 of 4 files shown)


gervandiepen:
The idea of the MS definition is that observatory or instrument specific columns (or subtables) get a prefix with the name of the observatory or instrument and that such code does not get part of Casacore.
ASTRON has done that for Lofar and Apertif. I think that NRAO should do this in a similar way by using prefix CASA_ or whatever.
Note that these columns have not been mentioned in the MSv3 discussion, so I assume they are quite specific.

cquike:
Hi,
these columns are shared by ALMA and VLA, so in that sense they are not fully instrument specific. I am not an expert on this but it looks like using pre-averaging in the correlator can be a fairly common thing.
I mentioned the needed changes to Kumar and George in during the CASA F2F this year and I believe they have added provisions for that in MSv3, but I don't know the details about how this has been conveyed to the MSv3 project.

bgarwood:
These columns are found in the science data model (SDM) shared with ALMA and VLA so I chose to use "SDM" as the prefix in what seemed to me at the time to be a reasonable adaptation of that MS definition comment on telescope-specific column name prefixes.

gervandiepen:
I hadn't realized SDM was used as the prefix. That sounds very good. I have
no problems with this change.

schiebel:
I have not kept up with the MS design and implementation. Is it not possible to create a derived class that has the telescope specific content upon which CASA depends on instead of putting it in the base class?

gervandiepen:
That's indeed what I did for LOFAR specific subtables and columns (and put
it in the LOFAR repository and documented it in a LOFAR note).
When doing that for CASA it will require several changes in the CASA
application code as it uses the MS classes. Since CASA is used by other
instruments as well, I very much hesitate to make such changes since CASA
could make such columns mandatory.

Note that when adding these columns to the standard MS classes, it should
be documented in note 229.

cquike:
Hi,
the class in question is MSSpWindowColumnsas returned by MS class itself. Unless we also create a derived class for MeasurementSet itself I don't see a clear fit of a derived class here. Needless to say, to add a derived MeasurementSet class thought CASA would be a major effort.
If I understand this paper (https://arxiv.org/pdf/1801.04834.pdf) correctly, also LOFAR applies some average on the correlator side, so it could even apply to that instrument. The original request to add this to MSv3 goes in the same line: it is an observational mode commonly used by correlators.

cquike:
Dropping PR since a solution was implemented on the casa level

PR #791, opened 2018-10-29
CAS-11985 squash warnings in casacore/ms

dmehring:


gijzelaerr:
It looks like the code was commented out in a sync from the NRAO repository into the casacore tree:

https://github.com/casacore/casacore/commit/f1ff7dbd7254e038349af821fde00689c4117125#diff-1f373144767b32acd515840d9fa4603bR61

dmehring:
Thanks for tracking this down. I'm not sure we at NRAO have this repository around from six years ago, so it may not be possible to follow the bread crumb trail farther. My feeling is that, unless there is a good reason to leave this commented out, my current change should be merged as 1. it reverts the code to follow the long adopted pattern of including .tcc files, 2. there are no compile issues in doing so, and 3. it eliminates numerous clang compiler warnings.

dmehring:
Thanks for the merge. I did some digging and was able to find our old casacore svn repos, and the revision in which this was commented out. The person who made that change is no longer with NRAO, and the commit message does not indicate why that particular change was made. It appears to have been made as part of a much larger commit, and from what I can tell, that particular change was not relevant to that commit intent. So, it remains a mystery. For now, I think we are good on reverting that old change, as now has been done. Thanks

PR #802, opened 2018-11-06
Harmonized get/setProperties

### LINKED ISSUE #801: Make set/getProperties consistent
It is possible to get or set the properties of a data manager, in particular the cache size of a storage manager. The Record fields to do so are named in an inconsistent way and should be harmonized.
Furthermore, the argument order of TableProxy::setProperties does not match its Python counterpart.

gervandiepen:
Close #801


gervandiepen:
The builds and tests are fine, but some builds timed out. Why does it take > 50 min while the others take < 10 min?

gervandiepen:
It would be nice if it could be part of the release.

gijzelaerr:
i think those are issues with the (free of charge) travis build farm.

Issue #799, opened 2018-11-01
Measures solar-system-objects directions not working


sabourke:
_From @sabourke on October 25, 2018 13:53_

Hi, the following code used to work for me:
```
import casacore.measures

me = casacore.measures.measures()
oso = me.observatory('OSO')
time = me.epoch('UTC', 'today')
sun = me.direction('SUN')

me.do_frame(oso)
me.do_frame(time)

sun_azel = me.measure(sun, 'AZEL')
```
but now it fails with:
```
  File "t1.py", line 10, in <module>
    sun_azel = me.measure(sun, 'AZEL')
  File "/usr/lib/python2.7/dist-packages/casacore/measures/__init__.py", line 119, in measure
    return _measures.measure(self, v, rf, off)
RuntimeError: Cannot convert due to missing frame information

```
It will work with a J2000 direction. Any ideas? This is with ubuntu 18.04 stock packages.

(Edited to add do_frame calls which I had omitted in this example.)

_Copied from original issue: casacore/python-casacore#141_

sabourke:
_From @gijzelaerr on October 30, 2018 14:58_

do you have the casacore data installed?

sabourke:
Yeah. I have all the casacore packages installed.

sabourke:
I did a bit more digging and it the necessary data is not in the casacore-data packages. If I use NRAO's data it works. Specifically the needed information seems to be in `ephemerides/DE200`

Additionally, the debian casacore data packages cause a lot of accuracy warnings. See sample below. When the `geodetic/IERSeop97`, `geodetic/IERSpredict` and `geodetic/TAI_UTC` tables (included in the NRAO data) are present all warnings disappear.

```
2018-11-01 11:11:21	SEVERE	MeasTable::dUTC(Double) (file /build/casacore-IDmUo1/casacore-2.4.1/measures/Measures/MeasTable.cc, line 4396)	Leap second table TAI_UTC seems out-of-date. 
2018-11-01 11:11:21	SEVERE	MeasTable::dUTC(Double) (file /build/casacore-IDmUo1/casacore-2.4.1/measures/Measures/MeasTable.cc, line 4396)+	Until the table is updated (see the CASA documentation or your system admin),
2018-11-01 11:11:21	SEVERE	MeasTable::dUTC(Double) (file /build/casacore-IDmUo1/casacore-2.4.1/measures/Measures/MeasTable.cc, line 4396)+	times and coordinates derived from UTC could be wrong by 1s or more.
[...]

PR #803, opened 2018-11-07
Fix JPL ephemerides download (DE200, DE405)

sabourke:
Fix for casacore/casacore/issues/537

### DIFF
--- a/measures/apps/measuresdata/measuresdata.cc
+++ b/measures/apps/measuresdata/measuresdata.cc
@@ -391,7 +391,7 @@ struct tableProperties {
 
 // Types that can be specified
 const String intypes[][2] = {
-  { "ALL", 		"IERS IGRF"}, //-- JPL IGRF"},
+  { "ALL", 		"IERS IGRF JPL"},
   { "IERS", 		"TAI_UTC EOP Predict"},
   { "EOP",		"IERSeop97 IERSeop2000"},
   { "Predict",		"IERSpredict IERSpredict2000"},
@@ -756,6 +756,13 @@ String blockIntToString(Block<Int> yn) {
   return out;
 }
 
+uInt StringToUInt(String s) {
+  uInt out;
+  istringstream ss(s);
+  ss >> out;
+  return out;
+}
+
 // Get today's MJD
 Double today_mjd() { return (floor(today_now())); }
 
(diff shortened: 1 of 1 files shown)


dpetry:
It is great that you took the time to work on this. Thanks!
Since the main site to run this is ASTRON, I leave it to somebody from there to approve this.
What tests did you perform?

sabourke:
I tested them both from python with code similar to the sample in #799, compared the keywords and desc of the generated tables with the ones distributed in CASA<sup>1</sup> and examined them in the table browser. 

<sup>1</sup> some keywords in the CASA tables have a newline at the end which doesn't look intentional. It doesn't cause problems though. The tables generated by measuresdata in casacore don't have this issue.

dpetry:
Which keywords have the additional newline?
Could this still be fixed?
(Might cause problems in the automated interpretation of the keywords.)

gijzelaerr:
Awesome, I think this is important enough to get into the imminent 3.0 release, but I don't have the knowledge to verify this results in the correct data and this PR is correct. @tammojan @gervandiepen ?

dpetry:
@tammojan , could this be tested in the ASTRON context ?

sabourke:
@dpetry The issue is present in the tables distributed by NRAO, **not** the tables generated by casacore + this PR and it's not actually causing any problem. JDEPOC, MA0001 and QQQQQQ

dpetry:
Yes, the NRAO tables are updated from the ASTRON master copy since some time.
So, from my point of view this could be merged but I would like @tammojan to comment or better test  this because he knows best (AFAIK) how this code is used to create the ASTRON distribution of CASA data.

tammojan:
@gijzelaerr Could you run `measuresdata-update`  on this branch?

gijzelaerr:
I can successfully run `measuresdata-update` on the casacore-data data, and it looks like `IERSeop2000`, `IERSeop97`, `IERSpredict`, `IERSpredict2000` and `TAI_UTC` get updated.

sabourke:
@gijzelaerr which `measuresdata` is getting called by `measuresdata-update`? The one from this branch or is the standard old version in you PATH?

gijzelaerr:
good point. that was the old `measuresdata`. If I remove my old measuredata and use the new one I get:

```
gijs@gijs-artica:/opt/casacore/data  
λ  PATH=/opt/casacore/bin:$PATH measuresdata-update
Calling measuresdata
 
Create data tables for Measures
-----------------------------------------------
terminate called after throwing an instance of 'std::system_error'
  what():  Unknown error -1
Aborted (core dumped)
Severe error calling measuresdata with
```

sabourke:
Aw man... ok, I'll look into this either today to Monday

sabourke:
This looks like a regression in casacore. If you build the master branch it also crashes, right? @gijzelaerr can you confirm?

gijzelaerr:
I can confirm this also happens with the current master. https://github.com/casacore/casacore/issues/814

tammojan:
The astron server that updates the measures data is still on casacore 2.7 and will not be updated in 2018. While I think it is great to fix the measures update programs on master, this should not block the release of casacore 3.0.

gijzelaerr:
ok, then we are good to go for 3.0. will do on Monday.

gijzelaerr:
https://github.com/casacore/casacore/issues/815 has been resolved.

PR #809, opened 2018-11-07
Fixing warnings due to misleading indentations

rtobar:
In the spirit of clearing warnings before 3.0, please have a look at these. I found them with gcc 6.3 + -Wall (in particular with -Wmisleading-indent) some time ago. Most of the changes are statements that should have been within brackets in the first place. The commit message contains more information.

### DIFF
--- a/measures/Measures/test/tMeasJPL.cc
+++ b/measures/Measures/test/tMeasJPL.cc
@@ -75,8 +75,8 @@ int main()
       os << "Venus:      " << val << endl;
       for (uInt i=0; i<3; i++) {
         mvd1(i) = val(i);
-        mvd1.adjust();
       }
+      mvd1.adjust();
       os << "Venus:      " << mvd1 << endl;
       MeasJPL::get(val, MeasJPL::DE200, MeasJPL::EARTH, dat);
       os << "Earth:      " << val << endl;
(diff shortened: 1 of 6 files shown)


measures/Measures/test/tMeasJPL.cc
```diff
@@ -73,7 +73,10 @@ int main()
       os << "Mercury:    " << val << endl;
       MeasJPL::get(val, MeasJPL::DE200, MeasJPL::VENUS, dat);
       os << "Venus:      " << val << endl;
-      for (uInt i=0; i<3; i++) mvd1(i) = val(i); mvd1.adjust();
+      for (uInt i=0; i<3; i++) {
+        mvd1(i) = val(i);
+        mvd1.adjust();
```
gervandiepen: Shouldn't this be outside the loop?

rtobar: Yes, that's right, and actually that one test failed. I'll fix accordingly

gervandiepen:
I had one inline comment in tMeasJPL. This test program failed, probably due to this issue.

gervandiepen:
Thanks for making these changes. I also very much dislike such constructs.

tammojan:
Thanks a lot! The tests succeed now, so I merged it.

tammojan:
Hm, I merged it into casacore:2.5prep, should probably go to master instead. Am looking into it.

PR #813, opened 2018-11-08
Fixed TaQL segfault and a valgrind error

### LINKED ISSUE #812: TaQl segfaults when referring to calculated column
The TaQL command
    select mscal.ha() as rt from ms orderby rt
gives a segfault.

gervandiepen:
valgrind found an array bound error in ExprFuncNode.cc.
It also improves NullTable how to avoid compiler warnings.

Close #812


tammojan:
Thanks Ger!

PR #821, opened 2018-11-19
CAS-12048 add parens to squash compiler warning

dmehring:


schiebel:
This changes the behavior, but without this change the left element of the disjunction is always false (which doesn't seem like the expected behavior).

PR #832, opened 2018-11-26
Improve calculation of TSM cache size; specify max in MiB

### LINKED ISSUE #822: Table system caching limit ignored, memory use grows without bound
We've been getting some Out of Memory errors on our cluster when splitting an MS. This means it is using more than 64GB of memory. I've done some tests on a smaller dataset on my Mac to track this down.

When splitting of the first 216 channels of a 3880 channel spectrum, tiled with [4,2578,78] and setting a 2GB max cache size, the tiling stats show the limit is ignored and ~40GB is allocated.
Below are the input and ouput cube stats for the DATA column. 
For the input MS the bucketsize is 6.5MB and it tries to cache an entire column of tiles (6277), using 
about 40GB (testing on a 16 GB Mac). 
I tracked this down to two problems in TSMCube.cc:
Firstly the test in validateCacheSize uses cacheSize*bucketSize. This overflows the range of uInt many times, failing the test and it then continues with the full 6277 tile cache. Suggested fix:
```
     // An overdraft of 10% is allowed.
-    if (maxSize > 0  &&  cacheSize * bucketSize > maxSize) {
+    if (maxSize > 0  &&  cacheSize > maxSize / bucketSize) {

```
Secondly, there is a test on total memory usage in setCacheSize that would be used if we don't specify a maximum cache size.  However, it is never triggered because we're 'only' reading and caching half the tiles in the cube not the entire cube. Not sure how to best fix this one.
```
    // If not userset and if the entire cube needs to be cached,
    // do not cache if more than 20% of the memory is needed.
    if (!userSet  &&  cacheSize >= nrTiles_p) {
      uInt maxSize = uInt(HostInfo::memoryTotal() * 1024.*0.2 / bucketSize_p);
      if (cacheSize > maxSize) {
	cacheSize = 1;
      }

```

Finally, we might want to specify a maximum cache size > 4GB, but this is currently not possible, because the size is specified in bytes using a uInt. A change to uLong or a unit change to MB would fix this.

Input MS:
>>> TSMCube cache statistics:
cubeShape: [4, 3888, 489600]
tileShape: [4, 2592, 78]  (tile shape set wrong for cube shape - this will be fixed)
maxCacheSz:2044723200
cacheSize: 6277 (*6469632)
#buckets:  12554
#reads:    6278
#accesses: 6298        hit-rate:  0.317561%
<<<

Output MS:
>>> TSMCube cache statistics:
cubeShape: [4, 216, 489600]
tileShape: [4, 1, 2048]
maxCacheSz:0
cacheSize: 216 (*65536)
#buckets:  51840
#inits:    51840
#writes:   51624
#accesses: 55944        hit-rate:  7.33591%

gervandiepen:
The calculation of the TSM cache size has been improved by avoiding a 32-bit overflow problem.
The cache size is now specified in MiB instead of bytes.
Also the maximum memory limit is always checked to avoid using too much memory.

Close #822

### DIFF
--- a/tables/DataMan/ISMBase.h
+++ b/tables/DataMan/ISMBase.h
@@ -128,7 +128,7 @@ public:
     virtual Record getProperties() const;
 
     // Modify data manager properties.
-    // Only ActualCacheSize can be used. It is similar to function setCacheSize
+    // Only MaxCacheSize can be used. It is similar to function setCacheSize
     // with <src>canExceedNrBuckets=False</src>.
     virtual void setProperties (const Record& spec);
 
--- a/tables/DataMan/SSMBase.h
+++ b/tables/DataMan/SSMBase.h
@@ -194,7 +194,7 @@ public:
   virtual Record getProperties() const;
 
   // Modify data manager properties.
-  // Only ActualCacheSize can be used. It is similar to function setCacheSize
+  // Only MaxCacheSize can be used. It is similar to function setCacheSize
   // with <src>canExceedNrBuckets=False</src>.
   virtual void setProperties (const Record& spec);
 
--- a/tables/DataMan/TSMCube.cc
+++ b/tables/DataMan/TSMCube.cc
@@ -161,7 +161,7 @@ void TSMCube::showCacheStatistics (ostream& os) const
         os << ">>> TSMCube cache statistics:" << endl;
         os << "cubeShape: " << cubeShape_p << endl;
         os << "tileShape: " << tileShape_p << endl;
-        os << "maxCacheSz:" << stmanPtr_p->maximumCacheSize() << endl;
+        os << "maxCacheSz:" << stmanPtr_p->maximumCacheSize() << " MiB" << endl;
         cache_p->showStatistics (os);
         os << "<<<" << endl;
     }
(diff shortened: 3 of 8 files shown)


gervandiepen:
Thanks for finding this issue.
I had run all tests, but clearly I did not pay enough attention to
PagedArray. I'll change that code. I think that one time the nr of pixels
should be changed from uInt to Int64 (or size_t), but that causes an ABI
change which I do not like. So I'll leave that for the time being.

Could you change the CASA code where needed?

Cheers,
Ger

cquike:
Hi,
the problem is that if a class holds a pointer to DataManager and calls virtual method setMaximumCacheSize(), the semantics change depending on the DataManager subclass used at runtime. For instance, in PlainColumn::setMaximumCacheSize() this calls dataManPtr_p->setMaximumCacheSize(nbytes). Depending on which data manager is being used the nbytes will be interpreted as bytes for most DataManagers or as megabytes (in TiledStMan). So there is no reliable way to "fix" the client code. All the derived classes and all the calls to setMaximumCacheSize() in casacore have to be changed first.

gervandiepen:
I'll create an issue and streamline all of it.

PR #837, opened 2018-11-29
fixed MPI and Adios2StMan bugs

JasonRuonanWang:
Somehow a MPI_Comm_rank call is missing in BaseTable.cc which causes all ranks not knowing what itself is. 

@rtobar This should fix the error you have seen from writing multiple mpi processes

### DIFF
--- a/tables/DataMan/Adios2StMan.cc
+++ b/tables/DataMan/Adios2StMan.cc
@@ -31,71 +31,31 @@
 namespace casacore
 {
 
-bool Adios2StMan::itsUsingMpi;
 std::string Adios2StMan::itsAdiosEngineType;
 adios2::Params Adios2StMan::itsAdiosEngineParams;
 std::vector<adios2::Params> Adios2StMan::itsAdiosTransportParamsVec;
 
-#ifdef HAVE_MPI
 MPI_Comm Adios2StMan::itsMpiComm = MPI_COMM_WORLD;
 
 Adios2StMan::Adios2StMan(MPI_Comm mpiComm)
     : DataManager(),
     itsAdios(mpiComm, true)
 {
-    itsUsingMpi = true;
     itsMpiComm = mpiComm;
     std::string engineType;
     adios2::Params engineParams;
     std::vector<adios2::Params> transportParams;
-    Adios2StManCommon(engineType, engineParams, transportParams);
+    Adios2StMan(mpiComm, engineType, engineParams, transportParams);
 }
 
 Adios2StMan::Adios2StMan(
-    MPI_Comm mpiComm, std::string engineType,
-    std::map<std::string, std::string> engineParams,
-    std::vector<std::map<std::string, std::string>> transportParams)
-: DataManager()
+        MPI_Comm mpiComm, std::string engineType,
+        std::map<std::string, std::string> engineParams,
+        std::vector<std::map<std::string, std::string>> transportParams)
+    : DataManager(),
+    itsAdios(mpiComm, true)
 {
-    itsUsingMpi = true;
     itsMpiComm = mpiComm;
-    Adios2StManCommon(engineType, engineParams, transportParams);
-}
-#endif
-
-Adios2StMan::Adios2StMan() : DataManager()
-{
-    itsUsingMpi = false;
-    std::string engineType;
-    adios2::Params engineParams;
-    std::vector<adios2::Params> transportParams;
-    Adios2StManCommon(engineType, engineParams, transportParams);
-}
-
-Adios2StMan::Adios2StMan(
-    std::string engineType, std::map<std::string, std::string> engineParams,
-    std::vector<std::map<std::string, std::string>> transportParams)
-: DataManager()
-{
-    itsUsingMpi = false;
-    Adios2StManCommon(engineType, engineParams, transportParams);
-}
-
-Adios2StMan::~Adios2StMan()
-{
-    if (itsAdiosEngine)
-    {
-        itsAdiosEngine->EndStep();
-        itsAdiosEngine->Close();
-    }
-}
-
-void Adios2StMan::Adios2StManCommon(
-    const std::string &engineType,
-    const std::map<std::string, std::string> &engineParams,
-    const std::vector<std::map<std::string, std::string>> &transportParams)
-{
-
     itsAdiosEngineType = engineType;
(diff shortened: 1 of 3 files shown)


gervandiepen:
Jason,
I assume MPI_Initialized is thread-safe, so the new BaseTable code works fine if multiple threads call this part of the code at the same time. I did not see thread-safety mentioned in the OpenMPI man pages I found on the internet.
Ger

JasonRuonanWang:
@gervandiepen 

All MPI functions are thread-safe if MPI is built in the threading mode. But if you have a non-thread-safe MPI installation, any of the MPI function calls can have problem, not only this MPI_Initialized, but everything else I added for this whole MPI-aware casacore table layer and Adios2StMan. So I think the good practice of ensuring this is to check the thread mode of MPI in cmake, and enforce a dependency saying if casacore is built with thread enabled and MPI enabled at the same time, you have to have a thread enabled MPI installation, otherwise cmake should give an error.

JasonRuonanWang:
That being said, I think for writers there isn't any issue, because no CTDS writers were supposed to operate on the same physical table in parallel until the appearance of this MPI add-on and Adios2StMan. 

So I suppose the problem you mentioned is the reader side. Then the question is: is the original non-MPI table interface and all of the other storage managers thread-safe to be called from multiple threads at the same time for reading? If this is the case then I probably need to refine a lot of things in Adios2StMan because what has been in my mind is only MPI level parallelism. I haven't even thought about multi-threading.

gervandiepen:
A table can be neither read nor written from multiple threads. But
different tables can be accessed in different threads. I'm not sure if
MPI_Init is done once per table or once per process. If it's the latter, it
should be thread-safe.

JasonRuonanWang:
Oh okay. I think I understand your concern now. Then indeed this is not thread-safe even if it is linked to a thread-safe MPI. Because the block that checks the MPI_initialized and does the MPI_Init is not atomic. So even if the MPI is built with multi-threading, multiple casacore application threads can still reach the point where they pass the MPI_Initialized and call MPI_Init at the same time. But it's pretty easy to fix. Let me add a mutex around all these pieces that check the initialization of MPI. 

The real problem after that is Adios2StMan. It has a lot of ADIOS2 function calls that potentially use MPI communicators underneath. So even if multiple threads are dealing with different tables, they can still possibly use the same MPI communicator at the same time. To make this absolutely safe I have to add a mutex for every ADIOS2 function call if MPI is not built thread-safe. I don't need to change anything if we can assume MPI is built with thread-safe though.

JasonRuonanWang:
@gervandiepen 

 I have done a bunch of tests on different platforms using a large number of threads (1000s), and it seems that MPI_Initialized and MPI_Finalized are always thread safe. So I think we can assume so for the time being. 

I was thinking about adding a mutex but this won't work after I thought it over, because if each of the threads has its own Table or Adios2StMan object, it won't share the same mutex with other threads anyway. You can only use mutex to lock these operations at the user application layer. Usually this kind of MPI initialization operations should be taken care of at the user application layer, because users know exactly what combination of thread + MPI environments they are in, so they can initialize everything correctly. Leaving this to the middleware layer can cause a bit of uncertainties. @rtobar also let me know if you have any suggestions on this?

rtobar:
@JasonRuonanWang I just tried out your changes. Although failing for the case of writing applications in principle sounds fine, unmodified reading applications (i.e., those that haven't started MPI on their own) are also failing due to the unmet requirement on the threading support of the MPI library.

For example, I tried running the `taql` tool to read an adios2 column, and it failed with:
```
Error in TaQL command: 'using style python SELECT gmin(MY_FLOAT_COLUMN) FROM my_table'
  Casacore is built with thread and MPI enabled, but the MPI installation does not support threads
```

This happens with any other tool. Moreover, this happens even when reading a table that has non-adios2 columns:

```
$> showtableinfo in=my_non_adios2_table
terminate called after throwing an instance of 'std::runtime_error'
  what():  Casacore is built with thread and MPI enabled,                             but the MPI installation does not support threads
```

I'm not sure what the best course of action is really. One possibility would be to change this to a warning instead of a `std::runtime_error` so applications can still run, but even then it would still "pollute" the output of those applications that have nothing to do with adios2.

> I have done a bunch of tests on different platforms using a large number of threads (1000s), and it seems that MPI_Initialized and MPI_Finalized are always thread safe. So I think we can assume so for the time being.

I read the standard and the only mentioning about thread safety is that it is guaranteed that all MPI calls are thread-safe *for a thread-compliant implementation* (so not exactly always). Reading around it seems that both OpenMPI and MPICH implement MPI_Init and friends on a thread-safe manner though, and maybe most vendors do.

Irrespective of the above, even though individual calls might be thread-safe, a collection of calls is not, unless properly guarded. Threads t1 and t2 can query at the same time whether MPI is initialized or not, both get a "false" result due to the particular serialization of their calls, and both try to initialize MPI.

> I was thinking about adding a mutex but this won't work after I thought it over, because if each of the threads has its own Table or Adios2StMan object, it won't share the same mutex with other threads anyway.

This would work fine I think if you use `std::once` or a static `std::mutex` for the MPI calls you perform.

JasonRuonanWang:
@rtobar The MPI support in casacore is a completely separate thing from Adios2StMan. In principle you can compile casacore with MPI enabled but Adios2StMan disabled. The MPI stuff was put into the table layer to ensure that when any parallel storage manager wants to create a table collectively, only rank 0 will actually create the table files, while other ranks stay idle.

You have seen this error because your MPI installation does not support threads, while you compiled casacore with threads enabled. If this is the case then those multi-thread tables readers will probably have problems regardless you use Adios2StMan or not, because this table layer MPI code will possibly crash anyway. Currently I don't have any better solution for this. 

The idea isn't really about any specific storage manager, but the fact that if you want to compile casacore with MPI and threads enabled at the same time, then your MPI installation must also have thread support.

std::call_once may work for this but I need to think carefully through. std::mutex is not likely to work because multiple threads may have multiple independent table objects, which will have multiple independent std::mutex 'es underneath. Do you know if I can make the mutex a static class member to ensure every thread locks the same mutex object? I have never done this before...

JasonRuonanWang:
So I think a better practice is to put this dependency into cmake, enforcing cmake to check the MPI installation's thread support when casacore is built with threads and MPI enabled at the same time.

PR #835, opened 2018-11-27
ADIOS2/MPI compilation fixes

rtobar:
This is a small set of changes around the compilation of casacore using `USE_ADIOS2=ON` (and therefore `USE_MPI=ON`). @JasonRuonanWang might be interested on these changes too.

After these changes I was able to compile the software. I still get a failure in {{tMSSelection}}, but I'd venture it's unrelated to these compilation fixes, and probably already present in the {{master}} branch with `USE_ADIOS2=ON` (but definitely not seen with `USE_ADIOS2=OFF`).

PR #840, opened 2018-12-03
Make sure lock file is read if NoLocking is used

gervandiepen:
PR #835 mentioned that tMSSelection failed. This issue has been fixed.
Furthermore, a memory leak in Adios2StMan was detected which has been fixed.


JasonRuonanWang:
@gervandiepen Could you go ahead and merge this so it will solve the Adios2StMan memory leaking problem too? Let me know if you need my effort on resolving the conflicts. Thanks.

Issue #844, opened 2018-12-13
Boost 1.67 Python components require a Python version suffix


gijzelaerr:
> Boost Python components require a Python version suffix (Boost 1.67 and later), e.g. python36 or python27 for the versions built against Python 3.6 and 2.7, respectively. This also applies to additional components using Python including mpi_python and numpy. Earlier Boost releases may use distribution-specific suffixes such as 2, 3 or 2.7. These may also be used as suffixes, but note that they are not portable.

This means the cmake find boost script requires us to be a bit more specific about what python version to expect.

https://cmake.org/cmake/help/v3.13/module/FindBoost.html

first reported here

https://github.com/ska-sa/homebrew-tap/pull/19

rtobar:
I just want to give my +1 here, as this bite me a few days ago. Locally I changed:

```
-        find_package(Boost REQUIRED COMPONENTS python-py${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR})
+        find_package(Boost REQUIRED COMPONENTS python${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR})
```

but this obviously is not generic enough. I also added:

```
-set(Python_ADDITIONAL_VERSIONS 3.5 3.4)
+set(Python_ADDITIONAL_VERSIONS 3.6 3.5 3.4)
```

because I'm using 3.6, but then again I'm not sure this is actually required.

gijzelaerr:
PR here https://github.com/casacore/casacore/pull/846

PR #846, opened 2018-12-13
Fix support for newer cmake boost issue #844

gijzelaerr:


gervandiepen:
All builds failed.

gijzelaerr:
see https://github.com/casacore/casacore/issues/847

tammojan:
If there's an easy way to make this work also in ubuntu 16 (perhaps involving a command-line override for the boost-python library name), that would be helpful. We cannot assume everyone has switched to ubuntu 18.

gijzelaerr:
there you go

PR #849, opened 2018-12-13
CAS-12122 fix overflow bug for datasets with >= 2**32 points

dmehring:


dmehring:
This appears to have failed because of a timeout on the server and does not appear to be due to my one line code change.

PR #851, opened 2018-12-15
Extended variance support and other improvements

gervandiepen:
The Casacore variance/stddev was somewhat limited and only supported the sample variance. It has been extended with a ddof parameter (a la numpy) to support the population variance as well. Furthermore, support has been added for the variance, stddev and avdev of complex values.
Array, MaskedArray and TaQL have been changed for this purpose.

reverse functions have been added to TaQL to reverse a string or array elements (per axis).
Finally the taql program has been changed to be more recursive.


tammojan:
There's a bit much in this commit to review easily. But please don't spend time splitting the commits into more.

I think the change that `taql 'select * from bla.MS`' does not output results anymore is not desirable. You told that this is to overcome the problem that taql often spits out too much data. An alternative way to overcome this same problem would be to set a maximum number of rows.

gervandiepen:
I'll change taql such that by default it lists the selected columns.
The (to be added) option -m gives the maximum nr of rows to show which
defaults to 50. Note that -m is only used for the implicit print. Explicit
printing shows all rows.

tammojan:
What do you mean by implicit and explicit print?

tammojan:
Some of your changes add whitespace at the end of a line. Could you `sed` those out?

tammojan:
Some of your changes introduce tabs (e.g. tables/TaQL/ExprFuncNode.h). Could you replace those with the appropriate amount of whitespace?

gervandiepen:
Explicit is printing using the -p option.
Implicit is a count or select where columns are specified.

PR #853, opened 2018-12-18
improve unicode support

gijzelaerr:
This improves the unicode support for python2, which was previously not really existing.

https://github.com/casacore/python-casacore/pull/159


gijzelaerr:
I tried to make the tabs spaces situation uniform, but I see now the indentation is not fully correct everywhere.

gijzelaerr:
I think this is good to go.

casa/Utilities/Regex.cc
```diff
@@ -357,6 +357,7 @@ String Regex::fromSQLPattern(const String &pattern)
 	case ')':
 	case '\\':
             result.push_back ('\\');
+            break;
```
bmerry: That looks wrong, although I haven't looked at the entire surrounding context. It's pushed the escaping backslash, but the character itself should still be pushed I think. It should rather have a comment to document that it is falling through.

python/Converters/PycArrayComCC.h
```diff
@@ -246,7 +248,7 @@
   void ArrayCopy<String>::toPy (void* to, const String* from, size_t nr)
   {
     PyObject** dst = static_cast<PyObject**>(to);
-    for (size_t i=0; i<nr; i++) {
+    for (uInt i=0; i<nr; i++) {
```
bmerry: I'm not sure what type `uInt` is, but if it's `unsigned int` then this change will break conversions where `nr` is more than 2**32 on 64-bit systems.

python/Converters/PycBasicData.h
```diff
@@ -99,19 +99,24 @@ namespace casacore { namespace python {
...
       boost::python::converter::rvalue_from_python_stage1_data* data)
     {
-#ifdef IS_PY3K
-    PyObject * temp_bytes = PyUnicode_AsEncodedString(obj_ptr, "ASCII", "strict"); // Owned reference
-    char* value;
-    if (temp_bytes != NULL) {
-        value = PyBytes_AS_STRING(temp_bytes); // Borrowed pointer
-        value = strdup(value);
-        Py_DECREF(temp_bytes);
+    char* value = 0;
```
bmerry: Personally I dislike `0` as a null pointer constant, and prefer `NULL` (or `nullptr` in C++11).

python/Converters/PycBasicData.h
```diff
@@ -99,19 +99,24 @@ namespace casacore { namespace python {
...
+    if (PyUnicode_Check(obj_ptr)) {
+        PyObject * temp_bytes = PyUnicode_AsEncodedString(obj_ptr, "UTF-8", "strict"); // Owned reference
+
+        if (temp_bytes != NULL) {
+            value = PyBytes_AS_STRING(temp_bytes); // Borrowed pointer
+            value = strdup(value);
+            Py_DECREF(temp_bytes);
+        } else {
+          boost::python::throw_error_already_set();
```
bmerry: Mis-indented.

python/Converters/PycBasicData.h
```diff
@@ -99,19 +99,24 @@ namespace casacore { namespace python {
...
-        value = PyBytes_AS_STRING(temp_bytes); // Borrowed pointer
-        value = strdup(value);
-        Py_DECREF(temp_bytes);
+    char* value = 0;
+    if (PyUnicode_Check(obj_ptr)) {
+        PyObject * temp_bytes = PyUnicode_AsEncodedString(obj_ptr, "UTF-8", "strict"); // Owned reference
+
+        if (temp_bytes != NULL) {
+            value = PyBytes_AS_STRING(temp_bytes); // Borrowed pointer
+            value = strdup(value);
```
bmerry: I think this copy is going to leak (unless the String constructor takes ownership, but in that case it's going to steal ownership from PyString_AsString in Python 2).

python/Converters/PycValueHolder.cc
```diff
@@ -238,58 +233,54 @@ namespace casacore { namespace python {
...
-	}
-        // Use the 'highest' type.
-	if (result != TpDComplex) {
-	  if (dt == TpDComplex) {
-	    result = dt;
-	  } else if (result != TpDouble) {
+        if (result == TpBool  ||  result == TpString || dt == TpBool  ||  dt == TpString) {
+          throw AipsError ("PycValueHolder: incompatible types in sequence");
+        }
+            // Use the 'highest' type.
```
bmerry: Comment now seems mis-indented.

gijzelaerr:
Thanks to Bruce the python-casacore test suite runs without errors now!

gijzelaerr:
ok I'm just gonna merge this then!

python/Converters/test/tConvert.cc
```diff
@@ -101,6 +103,7 @@ namespace casacore { namespace python {
       .def ("testcomplex",    &TConvert::testcomplex)
       .def ("testdcomplex",   &TConvert::testdcomplex)
       .def ("teststring",     &TConvert::teststring)
+      .def ("testunicode",     &TConvert::testunicode)
```
bmerry: Second parameter is mis-aligned.

gijzelaerr:
o no! i only see your feedback now! going add this

casa/Utilities/Regex.cc
gijzelaerr: Ok i'll just revert this change, don't fix it if it ain't broken (for me).

python/Converters/PycArrayComCC.h
gijzelaerr: again not sure what happened there, it has been a long time ago.

python/Converters/PycBasicData.h
gijzelaerr: yeah indentation was a mess, tried to fix it but it is still messy.

gijzelaerr: i think the project needs a reformatting PR, going to leave this now.

python/Converters/PycBasicData.h
gijzelaerr: well spotted. I'm not sure what to do here, it is a bug in the original code. What do you propose?

bmerry: I would do the following
- create a Boost.Python wrapper around `temp_bytes` to ensure that it gets decref'ed once it goes out of scope (you'd need to check the syntax to ensure that the wrapper borrows the reference rather than adding its own).
- Put that wrapper at a sufficient outer scope so that it lives past creation of the `String`.
- Get rid of the `strdup`.

That should not only fix the leak, but eliminate a copy so that the `String` is constructed directly from the data in `temp_bytes`.

gijzelaerr: opened a new issue for this. Thanks Bruce! https://github.com/casacore/casacore/issues/908

python/Converters/PycArrayComCC.h
gijzelaerr: yes, this actually made the python-casacore tests fail. Fixed it! thanks Bruce!

python/Converters/PycBasicData.h
gijzelaerr: changed

python/CMakeLists.txt
```diff
@@ -80,6 +80,15 @@ Converters/PycExcp.cc
 Converters/PycImport.cc
 Converters/PycRecord.cc
 Converters/PycValueHolder.cc
+Converters/PycArray.h
```
ludwigschwardt: @gijzelaerr, why did you have to add these headers? Do we do the same for Python 3 as well?

PR #859, opened 2019-01-04
Added support for Int64 data columns

### LINKED ISSUE #839: Overflow in taql
I'm not sure where the overflow happens, but this output surprises me:
```
taql 'select gntrue(FLAG), gnfalse(FLAG) from NGC807.MS'
    select result of 1 rows
2 selected columns:  Col_1 Col_2
-1995702272     0
```
The data set has 100620 rows, 16384 channels, 4 polarizations. So the answer should be 6,594,232,320. Indeed, a simpler overflow:
```
taql '6594232320'
-1995702272
```

gervandiepen:
TaQL uses table columns for the results of select expressions. To fully support 64-bit integers in TaQL, it was needed to add such columns to the table system. Furthermore, support of 64-bit columns as such is a nice feature.
An extensive test program tStManAll has been written to test all read and write access functions for all types of columns with all types of data managers and tables. It revealed a small error in VirtualTaQLColumn which has been fixed by implementing default access functions in DataManagerColumn.

Close #839


tables/DataMan/test/tStManAll.cc
```diff
@@ -0,0 +1,841 @@
...
+  // Update data column slices.
+  cout << "  writeColumnSlice .." << endl;
+  updateTable (writeColumnSlice);
+  // Update data slice ranges.
+  cout << "  writeRangeSlice .." << endl;
+  updateTable (writeRangeSlice);
+}
+
+void testTaQLColumns()
+{
```
tammojan: This is just a placeholder?

tammojan:
For future PRs, please consider making slightly smaller commits. In this case, you made two major changes: add the test for all data managers and all data types (and fix minor issues that this test revealed), and add the Int64 data type. No need to split the commit after the fact though.

PR #862, opened 2019-01-18
Correct method name in error messages

rtobar:
Today while playing with the Adios2StMan I was surprised to see a
reference to putArray in my error messages, as I was actually juggling
with the putSlice* methods. I dug a bit and realized that the error
message was misleading. This patch fixes that, giving the correct method
name in the error message.


rtobar:
The Mac build on Travis failed because it ran out of time *just* before finishing, it looks like one or two more minutes and it would have finished running the unit tests correctly.

PR #863, opened 2019-01-23
Further Adios2StMan fixes

rtobar:
After playing with the Adios2StMan for a while now, including using it within the [OSKAR](https://github.com/OxfordSKA/OSKAR) simulator to write measurement sets, we have come up with the following set of fixes and improvements for it. They include:

 * Adding `getSlice`/`setSlice` support
 * Avoiding code duplication in the `Adios2StMan` and `Adios2StManColumn` classes
 * Implemented `Adios2StMan` using a pimpl idiom to avoid passing on ADIOS-specific dependencies to users
 * Improved the MPI checks on the `BaseTable` class, including error checking, code centralization and serialization of the initial MPI checks.

I've run these changes through @JasonRuonanWang already and he thinks they seem fine.

### DIFF
--- a/tables/DataMan/Adios2StMan.cc
+++ b/tables/DataMan/Adios2StMan.cc
@@ -25,6 +25,8 @@
 //#
 //# $Id$
 
+
+#include "Adios2StManImpl.h"
 #include "Adios2StManColumn.h"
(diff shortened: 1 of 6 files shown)


JasonRuonanWang:
Thanks very much. This looks great!

PR #864, opened 2019-02-01
Added missing feature in copyWeather in MSConcat (don't reindex ant ids==-1)

dpetry:
See also CAS-12226 .


schiebel:
thanks Dirk...

PR #869, opened 2019-02-14
Made static initialisation of Unit and MS more robust

### LINKED ISSUE #865: MS::requiredTableDesc seems to have a race condition
Occasionally the Apertif DataWriter stalls when creating its MeasurementSets (in parallel). It looks as if the (quite complex) static requiredTableDesc function can have  a deadlock due to a spurious race condition.
If possible, the code should be simplified using the CallOnce semantics.

gervandiepen:
The Apertif software creates many MSs in parallel. Sometimes it stalled at the beginning of creating the MSs, which is most likely due to a race condition in the static initialization of UnitMap or MS::requiredTableDesc resulting in a deadlock. A potential problem in these initializations is the recursion that can take place for which special code constructs had to develoiped.
The implementation has been simplified considerably by removing the MS recursion and by solving the UnitMap recursion in a different way. Furthermore, the UnitMap statics are now contained in a function static variable to solve the problem of the static initialization order.
Note that C++11 guarantees that a function static is initialized only once, also in a multi-threaded program.

Close #865

### DIFF
--- a/casa/Quanta/UnitMap.cc
+++ b/casa/Quanta/UnitMap.cc
@@ -34,92 +34,21 @@
 
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
-// static initialization
-unsigned long unit_map_initialize_::count = 0;
-
 // Initialize statics.
 Mutex UnitMap::fitsMutex;
 
 
-void UnitMap::initUM() {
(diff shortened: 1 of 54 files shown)


schiebel:
It looks like the test failure is with fetching test data.

PR #871, opened 2019-02-15
Fix adios2 cmake

JasonRuonanWang:
Asked our cmake expert to help refine the ADIOS2 related cmake scripts.


rtobar:
Just to chip in, I had an extremely similar set of local changes to the cmake scripts regarding the definition of the MPI-related macros to avoid linking against the C++ MPI libraries.

This global macro definitions are needed to add some missing bits of functionality in the `ms/` directory that allow passing down an explicit MPI communicator when creating a MeasurementSet and its subtables. Once this pull request is accepted I will create mine on top of it.

tammojan:
Thanks!

PR #875, opened 2019-02-24
fix osx w/g++ single library segv

schiebel:
Using statics is problematic. This change modifies MSTable statics to protect object initialization within a function context. The function scope ensures proper initialization. In one case, a static shared pointer was introduced to fix an earlier SEGV (i.e. requiredTD_p). Unfortunately, this does not resolve the general problem but rather a particular SEGV instance. Fixing this case was also required because my use case also encountered an exception with this.

My use case is casacore compiled into a single library on OSX with libraries built with g++ 5.5.

### DIFF
--- a/ms/MeasurementSets/MSAntenna.cc
+++ b/ms/MeasurementSets/MSAntenna.cc
@@ -126,7 +126,7 @@ MSAntenna& MSAntenna::operator=(const MSAntenna &other)
 
 void MSAntenna::init()
 {
-    if (! columnMap_p.ndefined()) {
+    if (! columnMap_p().ndefined()) {
 	// the PredefinedColumns
 	// DISH_DIAMETER
 	colMapDef(DISH_DIAMETER, "DISH_DIAMETER", TpDouble,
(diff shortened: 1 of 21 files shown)


gervandiepen:
I've submitted PR 869 about 10 days ago (which has not been merged yet). That PR addresses the same problem and gets rid of the recursion in the MSTableImpl::init function by separating the initialisation of the maps and the tabledesc. It also addresses the initialisation of UnitMap (using function statics).

I fear that PR 869 and 875 result in conflicts, so we have to choose what to do. I like your function statics, but I do not like the recursion in MSTableImpl::init. Could you take a look at PR 869 and say what you think?

Cheers,
Ger

gervandiepen:
Hi Darrell,

Another question.
Do you have any idea of the cost of having a static member in a function? In a multi-threaded environment the compiler has to ensure that the static is initialized only once, so for each call it has to do some mutex locking I presume. I wonder if it would be cheaper to have one struct containing all these statics and have a function getMaps() containing that static struct. columnMap_p() will look like {return getMaps().columnMap;}
I fear that having a static in a static inline function gives problems because the compiler will make a static per compilation unit. See https://stackoverflow.com/questions/185624/static-variables-in-an-inlined-function. In this case it does not give real problems because those statics are readonly. But having many, many copies of them is a waste.

gervandiepen:
Also see 
https://stackoverflow.com/questions/5567529/what-makes-a-static-variable-initialize-only-once?noredirect=1&lq=1

schiebel:
Hi Ger,

I think this initialization is very messy. Before your comment, I hadn't really looked at the full scope of initialization for MSTable et al. In fact, I had to use a debugger to understand. For my particular problem, your changes with respect to these maps do not resolve my SEGV (i.e. replacing SimpleOrdMap with std::map). In attempting to resolve your concerns as well, I did exactly what you just suggested. This can be seen in my GitHub branch [https://github.com/schiebel/casacore/tree/mstable-init-mess](url). However, this deadlocks with tMeasurementSet. I think the current initialization process may be too complicated to rely on compiler protections. I did remove the thread protections in MSTableImpl::init because I thought the compiler protections should be sufficient (so restoring these protections may help resolve the problem).

On my GitHub branch, I collected the static maps into a struct that I then attempted to construct within a static function. The best solution seemed like factoring out all of this a priori initialization for the maps and do the initialization directly in the struct which contains the collection of maps. I think this would make the initialization more clear (and allow initialization be ensured by the compiler). Unfortunately, I didn't know how much of the state in derived classes was needed to do the initialization so I threaded a pointer to the shared map collection struct down through the initialization functions. This initialization is very messy so it would be nice to clean it up, but that doesn't seem possible without a more significant refactoring.

[This](https://stackoverflow.com/questions/38391014/cost-of-thread-safe-local-static-variable-initialization-in-c11) seems like a good summary of static variable cost with C++11.

gervandiepen:
Hi Darrell,

I was not fully satisfied with the solution. It still used 2 static
CallOnce objects to do the initialization.
The whole MSTable stuff is awful.
I'm now transforming it to using a single function static MSTableMaps (per
table type) which gets initialized on creation. Hope to finish it today
(have a long meeting).

Cheers,
Ger

schiebel:
Thanks Ger, for my issues it is important but not urgent...

gervandiepen:
Hi Darrell,

Attached is a ms.tgz file containing the ms subdirectory with my new
sources. All tests on my Mac succeed. Tomorrow I want to test it on a Linux
machine with multi-threading.
Could you try it out for CASA and see it works fine?

As you'll see I've made quite some changes.
- MSTable::getMaps() contains a static with the maps and requiredTD; it is
initialized by an overloaded initMaps function in MSTableImpl.
- MSTable is now templated on a single parameter instead of 2. I needed
that for the first bullet (and is cleaner I think).
- To avoid the recursive call to getMaps() I've made functions such as
columnDataType taking a MSTableMaps object. They are used during
initialization.

Cheers,
Ger

PR #877, opened 2019-03-01
Fix thread safety of bulk allocator

aroffringa:
I'm having trouble getting casacore to open different measurement sets from multiple (~20) threads, resulting in crashes and messages that say that tables cannot be removed from the table cache. Helgrind identifies several race conditions, one of them being in the bulk allocator. This commit solves those race conditions. On a simple test application that opens two different measurement sets in two different threads, this fix decreases the number of helgrind errors from 124 to 71 in this test application.

When I run `make check` (somewhat to my surprise) I see several failures because of this edit. I'm quite confused why this is, but the old code generates a race condition so needs some kind of change, so I'm posting this for now as a pull request -- maybe someone else can comment on why apparently other code is depending on the non-thread-safe initialization of this static.

### DIFF
--- a/casa/Containers/Allocator.h
+++ b/casa/Containers/Allocator.h
@@ -239,14 +239,14 @@ class Allocator_private {
     typedef typename Allocator::pointer pointer;
     typedef typename Allocator::const_pointer const_pointer;
     typedef typename Allocator::value_type value_type;
-    virtual pointer allocate(size_type elements, const void *ptr = 0) {
+    virtual pointer allocate(size_type elements, const void *ptr = 0) override {
       return allocator.allocate(elements, ptr);
     }
-    virtual void deallocate(pointer ptr, size_type size) {
+    virtual void deallocate(pointer ptr, size_type size) override {
       allocator.deallocate(ptr, size);
     }
 
-    virtual void construct(pointer ptr, size_type n, const_pointer src) {
+    virtual void construct(pointer ptr, size_type n, const_pointer src) override {
       size_type i = 0;
       try {
         for (i = 0; i < n; ++i) {
@@ -258,7 +258,7 @@ class Allocator_private {
       }
     }
     virtual void construct(pointer ptr, size_type n,
-        value_type const &initial_value) {
+        value_type const &initial_value) override {
       size_type i = 0;
       try {
         for (i = 0; i < n; ++i) {
@@ -269,7 +269,7 @@ class Allocator_private {
         throw;
       }
     }
-    virtual void construct(pointer ptr, size_type n) {
+    virtual void construct(pointer ptr, size_type n) override {
       size_type i = 0;
       try {
         for (i = 0; i < n; ++i) {
@@ -280,7 +280,7 @@ class Allocator_private {
         throw;
       }
     }
-    virtual void destroy(pointer ptr, size_type n) {
+    virtual void destroy(pointer ptr, size_type n) override {
       for (size_type i = n; i > 0;) {
         --i;
         try {
@@ -290,10 +290,10 @@ class Allocator_private {
         }
       }
     }
-    virtual std::type_info const &allocator_typeid() const {
+    virtual std::type_info const &allocator_typeid() const override {
       return typeid(Allocator);
     }
-    virtual ~BulkAllocatorImpl() {}
+    virtual ~BulkAllocatorImpl() override {}
 
   private:
     static Allocator allocator;
@@ -306,8 +306,11 @@ class Allocator_private {
 
   template<typename Allocator>
   static BulkAllocatorImpl<Allocator> *get_allocator_raw() {
+    // Because this function gets called from destructors of statically allocated objects that get destructed
+    // after the program finishes, the allocator is constructed in a static storage space and is never
+    // destructed.
     static typename std::aligned_storage<sizeof(BulkAllocatorImpl<Allocator>), alignof(BulkAllocatorImpl<Allocator>)>::type storage;
-    static BulkAllocatorImpl<Allocator> *ptr =
+    static BulkAllocatorImpl<Allocator>* ptr =
       new (reinterpret_cast<BulkAllocatorImpl<Allocator>*>(&storage)) BulkAllocatorImpl<Allocator>();
     return ptr;
   }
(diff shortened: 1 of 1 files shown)


aroffringa:
I did some analysis of the allocator classes and I understand why my fix is giving problems, but it is pretty involved. In casacore there is a static instantiation of `BulkAllocatorImpl`, which is currently a static local variable in `BulkAllocatorImpl::get_allocator_raw()` (and since it is a template, this exists for every allocator type).

The `get_allocator_raw()` method is indirectly called from several places, including from the destructor of some of the `Block` class. Some of the tests have a `Block` class as static (possibly local) variable. Since that class calls `get_allocator_raw()`, it (might) instantiates the `BulkAllocatorImpl` at that point, which implies that the `BulkAllocatorImpl` will be destructed before the destructor of `Block` (since the compiler will destruct static objects in reverse), and causes UB when `Block` tries to use a reference to the destructed `BulkAllocatorImpl`.

The old thread-unsafe code sort-of prevented this by instantiating the `BulkAllocatorImpl` in local trivial storage that current compilers don't reclaim before destruction. I suspect it is undefined behaviour, because of this (from https://en.cppreference.com/w/cpp/utility/program/exit ):

> If a function-local (block-scope) static object was destroyed and then that function is called from the destructor of another static object and the control flow passes through the definition of that object (or if it is used indirectly, via pointer or reference), the behavior is undefined. 

This implies that if the static store were an object calling `get_allocator` after exit is undefined behaviour (unless destruction order can be guaranteed, which is not the case here), but I think object here refers to class instantiations and not trivial types, so it might be that the behaviour is defined. However, this rule does imply that wrapping a static mutex around it _is_ undefined behaviour, since a mutex is of course an object. So there is no way to make the allocator thread safe with a mutex.

I think the following is tried to guarantee that the bulk allocator is always instantiated before `main()`:

```
template<typename Allocator>
  class BulkAllocatorInitializer {
    BulkAllocatorInitializer() {
      get_allocator_raw<Allocator>();
    }
    static BulkAllocatorInitializer<Allocator> instance;
  };
```
However, this doesn't do anything; this would only instantiate a `BulkAllocatorInitializer<Allocator>` when there would be a `BulkAllocatorInitializer<Allocator>` for every Allocator type (which thus defeats the purpose of the class). Removing that class and its typedef doesn't change its working, and when doing so I found that everything compiles ok.

A possibly quick fix to make the old code thread safe is to wrap a static mutex around the old code, although the mutex might be destructed after exit(), so might crash because of this. In any case such a mutex would cause undefined behaviour. Another fix would be to truly (manually) instantiate a bulkallocator for every allocating-providing type, which is not a lot of work, but pretty ugly.

Personally I think the allocator classes need to be redesigned. Upon checking, it seems to me that there is actually no reason to have a static instance of the allocator, as all allocators are stateless. Their instantiations are only used to polymorphically call the right alloc()... method, but that can easily be done with templates and type traits, similar to how it is done in the newer c++ allocators, which would make everything compile time and therefore lighter.

I'm happy to do this rewrite, although I'm not sure how much code depends directly on the current architecture of the allocator.

So... what would be the best approach to fix this?

tammojan:
Ive added a label to prevent accidental merging. Probably good to discuss with @gervandiepen after the weekend.

aroffringa:
I think the solution that I pushed now is, compared to the old code, IMHO better in all aspects: it is thread safe, more readable and `make check` reports no errors. It is still a ugly hack around the "static initialization order fiasco" though.

helgrind still reports lots of errors in the allocator code, which I don't completely understand, but they might be false positives or come from other issues.

aroffringa:
BTW I think the 'dontmerge' label can be removed. `make check` works with the current branch, and it fixes the race condition. In some cases, helgrind still reports errors in this code, which I was told by helgrind developers are false positives. Their reply:

> I believe Helgrind does not understand modern GCC implementations of the Meyers Singleton. You will see similar warnings for any non-trivial static object.
> 
> See e.g. https://stackoverflow.com/q/44698412/

gervandiepen:
Before merging I want to check how statics in inline static functions behave. I'm not sure if there is one  instance (per template type) or if there is one per compile unit.

aroffringa:
A static method that is part of a class has one instance in the global scope. A static function outside a class has one per compilation unit. Inline doesn't change this.

I think these three questions together provide "proof" of this claim:
https://stackoverflow.com/questions/4992247/how-is-compilation-unit-defined-in-c defines compilation unit
https://stackoverflow.com/questions/1079623/what-is-the-lifetime-of-class-static-variables-in-c gives lifetime of static objects (either inside or outside)
https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.3.0/com.ibm.zos.v2r3.cbclx01/inline_member.htm shows inline makes functions external linkage (i.e. global scope instead of compilation unit scope).

( Note that the respective get-instance method is not a static function (which would have compilation unit scope), but it is a static method of a class, so it has global scope. )

gervandiepen:
IMHO the Allocator stuff is messy and seems needlessly complex. But I can only judge well after having looked at it more thoroughly. It lacks documentation though.
It looks as if no test of BulkAllocator exists (Kohji never wrote one it seems), but maybe it is implicitly used in Array or Block. I don't know if it is used in CASA or ASAP. Shall I merge it or first ask Darrell to check if CASA still builds and runs well?

The problem with templated allocators is that an Array allocated in one way has a different type than an Array allocated another way. So it is harder to mix them (e.g., assign to each other).

aroffringa:
The allocator instances in the bulkallocator in the end are the ones used in Blocks and Arrays, so they are used globally (and probably because of that reasonably covered by tests) and indirectly almost everywhere.

I agree it is messy. This fix is just a workaround to make that mess thread-safe, and it is not intended as a final solution. I'm reasonably confident the fix should have no further side effects, so I don't necessarily see the need to check Casa, but if you think it would be better to check, go ahead.

I don't have an overview why this complexity is required, or why allocators at all are useful in casacore. You're right that templated allocators would make Arrays different types, which is indeed probably not desirable. The singleton-allocator classes have no state and are currently only used to identify what allocator is used, using RTTI. This could be done with a simple `enum` or a polymorphic allocator that is instantiated for each object, probably without needing any statically allocated objects and allocator factories. (I noticed that some of the types named 'allocator' are actually allocatorfactories -- that's also a bit confusingly named).

Anyway, I think this fix could be applied to bridge until we have something better. I don't necessarily need it fixed so happy to wait also. If you want me to work a bit on a proposal that doesn't use allocator-objects, happy to do that, but am a bit busy so will take some time.

PR #879, opened 2019-03-07
adds accessor to underlying HDF5DataSet to HDF5Lattice

veggiesaurus:
For some of our work on CARTA, we need to read large slices of data to an allocated memory buffer directly (in this case, managed by a `std::vector`). The HDF5DataSet object's `get` method allows us to do so, but the HDF5Lattice currently does not expose the underlying HDF5DataSet object associated with that lattice. This PR adds a single accessor function which returns the HDF5DataSet. I have used `array()` rather than `dataset()` to be consistent with the existing `arrayName()` function.

### DIFF
--- a/lattices/Lattices/HDF5Lattice.h
+++ b/lattices/Lattices/HDF5Lattice.h
@@ -194,7 +194,7 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
       { return itsGroup; }
 
 	// Returns the current HDF5DataSet object
-	const CountedPtr<HDF5DataSet> array() const
+	const CountedPtr<HDF5DataSet>& array() const
 	{ return itsDataSet; }
 
     // Returns the name of this HDF5Lattice.
--- a/lattices/Lattices/HDF5Lattice.h
+++ b/lattices/Lattices/HDF5Lattice.h
@@ -193,6 +193,10 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
     const CountedPtr<HDF5Group>& group() const
       { return itsGroup; }
 
+	// Returns the current HDF5DataSet object
+	const CountedPtr<HDF5DataSet> array() const
+	{ return itsDataSet; }
+
     // Returns the name of this HDF5Lattice.
     const String& arrayName() const
       { return itsDataSet->getName(); }


gervandiepen:
As such I'm not opposed to this change, but it is HDF5 specific. It might mean that other image types (such as FITS) cannot be used in CARTA. The nice thing of the casacore images is that it hides how they are stored/constructed (FITS, casacore, HDF5, image expression, ...). It would be great if you would support that in CARTA. Do you work with Christy Reynolds? I've asked her the same thing, but unfortunately did not get an answer. It shouldn't be too hard to extend HDF5Image such that it understands the meta data of your HDF5 image format.

Note there is the function doGetSlice that does exactly what you want. It reads into a buffer, be it an Array object. But it is straightforward to construct that from a std::vector like:
    Array<T> (shape, &vec[0], SHARE)
The shape has to be correct though, because doGetSlice resizes if needed.
May I ask you why you use std::vector for multi-dim arrays?

veggiesaurus:
This method of reading would only be applied to HDF5 files, and would not affect compatibility with any other file types. Our main aim is to prevent additional memory allocations and copies of different array types when using a `std::vector`, which is used widely used in our codebase.

I believe our rationale behind using `std::vector` in our codebase was that `casacore::Array` was not guaranteed to be contiguous, and that random access might be slower than conventional containers (as suggested [here](https://casacore.github.io/casacore/classcasacore_1_1Array.html#af4639e5eaecf7ce267d9dfcb985b576f). An alternative for us might be to switch to `casacore::array` and use `copyToContiguousStorage`  in cases when the array is not contiguous. There were also some concerns about thread-safety, but those seemed speculative, as I can't imagine why the array class wouldn't be thread-safe. 

I've tried the method you've suggested, and it works well, thank you very much. So we could use that instead, but I we may need to directly access `HDF5DataSet` objects for other purposes in future (getting the dataset offset allow us to directly read from contiguous HDF5 files, bypassing thread-safety issues), so I would still like to keep the PR open.

gervandiepen:
Several years ago Array was not thread-safe due to non-atomic reference counting. But since then it uses std::shared_ptr internally which uses atomic operations to in- and decrement the count.

In principle Array elements are always contiguous. Only when taking a window from an Array, the resulting Array is a view on the original and the elements in that view might be non-contiguous.
BTW. Did you consider multi_array?

I'll merge the PR.

PR #881, opened 2019-03-07
Issue #878: use std::regex instead of cregex

### LINKED ISSUE #878: Race condition in Regex
Using helgrind Andre Offringa found a race condition in Regex.cc. It sets a global cregex variable defining the regex syntax in an unprotected way. A solution would be to put this inside a Mutex lock, but that makes regex matching single threaded which is not that nice.
It would be better to base Regex on std::regex, which has to be done one day anyway.

gervandiepen:
Regex has a race condition when setting the allowed grammar in a global cregex variable. It could be solved by introducing a Mutex, but that makes muli-threading impossible.
This PR is using std::regex (its ECMAScript) instead. Its syntax differs slightly from the cregex syntax which is by default adjusted for in the Regex constructor.
Note that this PR will change the Regex ABI.

Close #878


tammojan:
The Travis tests fail with a lot of warnings like
```
Test project /home/travis/build/casacore/casacore/build
        Start   1: tTypes
  1/538 Test   #1: tTypes ...........................***Failed    0.18 sec
terminate called after throwing an instance of 'casacore::AipsError'
  what():  Error in regex [ \*\./]: regex_error
Aborted (core dumped)
FAIL (execution failure): ./tTypes
```

See e.g. https://travis-ci.org/casacore/casacore/jobs/503072861#L4815 and https://travis-ci.org/casacore/casacore/jobs/503074914#L5224

gervandiepen:
gcc-4.8 has a broken std::regex. Grrrr, darn, ...
So for the time being this PR cannot be merged until a newer gcc version is used everywhere. I'll use another way to solve the race condition.

tammojan:
I rebased on top of the current master, and introduced a check on gcc >= 4.9.0 (in which `std::regex` was introduced). @gervandiepen could you review and merge when Travis succeeds?

tammojan:
(In case I messed up with the rebase, I saved the original branch as issue878-beforerebase.)

gervandiepen:
The rebase looks fine to me.
I compare Regex.h and .cc with my version and the only differences are in whitespace.

tammojan:
The LOFAR system CEP3 still has gcc 4.8.5 (on centos 7). However, I think we can use a newer gcc there to compile (using a compiler in singularity or so), so this does not need to be a show-stopper for this PR.

PR #890, opened 2019-04-01
Replace STL-like containers by their STL counterparts

### LINKED ISSUE #118: Discussion: Replace List,Map, etc. by STL counterparts
Casacore contains classes like List, Map, OrderedMap, HashMap. They can be replaced by their STL counterparts std::list and std::map. I'm not sure if that is also true for SimpleOrderedMap used in Tables. I have to check that.
It might well be possible that these Casacore classes are used in CASA code, so they should make the change first. I don't expect it to be used in other code like LOFAR and ASKAPsoft.

gervandiepen:
Classes Map (and variations), List, Queue and Stack have been replaced by their STL counterparts.
These Casacore classes are not built/installed anymore unless -DBUILD_DEPRECATED is given to cmake.

Close #118

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,10 +45,10 @@ set (CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 # By default build only Python2 bindings
 option (BUILD_PYTHON "Build the python bindings" YES)
 option (BUILD_PYTHON3 "Build the python3 bindings" NO)
+option (BUILD_DEPRECATED "build and install deprecated classes (such as Map)" NO)
 
 # By default build shared libraries
 option (ENABLE_SHARED "Build shared libraries" YES)
-
 option (ENABLE_RPATH "Include rpath in executables and shared libraries" YES)
 
 # Check if O_DIRECT is available.
@@ -485,6 +485,10 @@ endif (NOT DATA_DIR)
 set (DATA_DIR "${DATA_DIR}" CACHE PATH "Measures tables root")
 
 
+# Set compiler flag if deprecated Casacore header files are allowed.
+if (BUILD_DEPRECATED)
+    add_definitions(-DAIPS_USE_DEPRECATED)
+endif ()
 # Set compiler flag if no table locking.
 if (NOT ENABLE_TABLELOCKING)
     add_definitions(-DAIPS_TABLE_NOLOCKING)
@@ -563,6 +567,7 @@ message (STATUS "ADIOS2 library? ....... = ${ADIOS2_LIBRARIES}")
 message (STATUS "HDF5 library? ......... = ${HDF5_hdf5_LIBRARY}")
 message (STATUS "FFTW3 library? ........ = ${FFTW3_LIBRARIES}")
 
+message (STATUS "BUILD_DEPRECATED ...... = ${BUILD_DEPRECATED}")
 message (STATUS "BUILD_PYTHON .......... = ${BUILD_PYTHON}")
 message (STATUS "BUILD_PYTHON3 ......... = ${BUILD_PYTHON3}")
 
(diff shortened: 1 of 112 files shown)


dpetry:
MSConcat is making use of simpleOrderedMap a lot. Did you already look at that?

dpetry:
OK, I looked at your changes to MSConcat and you definitely did look at it.  :)
Question: you had to define an inline function getMapValue(). Is not even this simple functionality provided by the STL map? I am wondering why we are abandoning the CASA simpleOrderedMap if the STP version offers so few features ...

gervandiepen:
Less code to maintain and most likely better performance (certainly support
of move semantics).
I noticed, however, that one test (tRowCopier) failed with gcc-4.8, so I
have to look into that problem.

getMapValue is doing something special by returning -1 if not find.
std::map has the functions 'at' and 'find' to lookup a key. Operator[] only
works on non-const maps (unlike the Casacore Maps). I very much like that I
can now use  'for (const auto& x : myMap)'

BTW. An advantage of the Casacore Map classes was the support of a default
value that is used when a new key is defined. It made something like:
     if (value > map[key]) map[key]=value;
possible if a default value of e.g. min_int was used. std::map always has
T() as the default, so in such a case you first have to test if the key
exists.

Also note the somewhat unexpected behaviour of std::map::define which
doesn't do anything (apart from returning a reference to the value) when an
existing key with a new value is defined. No exception and no value
replacement. There is quite some discussion on stackoverflow.

gervandiepen:
Darrell Schiebel has built and tested the CASA package using this PR. All rests pass, so it can be merged.

AlexKurek:
I tried to use this flag in 3.1.2 and I got:

> cmake -DCMAKE_INSTALL_PREFIX=${INSTALLDIR}/casacore/ -DDATA_DIR=${INSTALLDIR}/casacore/data -DBUILD_PYTHON3=ON -DBUILD_PYTHON=OFF -DPYTHON3_EXECUTABLE=/usr/bin/python3 -DENABLE_TABLELOCKING=OFF -DUSE_OPENMP=ON -DUSE_FFTW3=TRUE -DUSE_HDF5=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS="${CXX_FLAGS} -fsigned-char -DNDEBUG" -DBUILD_DEPRECATED ../src/
> 
> Parse error in command line argument: -DBUILD_DEPRECATED
> Should be: VAR:type=value
> CMake Error: No cmake script provided.
> CMake Error: Problem processing arguments. Aborting.

tammojan:
Try with `-DBUILD_DEPRECATED=True`.

sarrvesh:
I am facing the same issue while compiling casarest with casacore 3.2.1. I tried casarest master and 1.5.0. Following the comments above, I tried using the branch useSTL and it fails with a different error:

[ 29%] Building CXX object msvis/CMakeFiles/casa_msvis.dir/MSVis/CalVisBuffer.cc.o
In file included from /casarest/msvis/MSVis/VisBuffer.h:39:0,
                 from /casarest/msvis/MSVis/CalVisBuffer.h:32,
                 from /casarest/msvis/MSVis/CalVisBuffer.cc:28:
/casarest/msvis/MSVis/VisibilityIterator.h:566:17: error: 'ROMSColumns' does not name a type; did you mean 'MSColumns'?
   virtual const ROMSColumns& msColumns() const
                 ^~~~~~~~~~~
                 MSColumns
In file included from /casarest/msvis/MSVis/CalVisBuffer.h:32:0,
                 from /casarest/msvis/MSVis/CalVisBuffer.cc:28:
/casarest/msvis/MSVis/VisBuffer.h:644:19: error: 'ROMSColumns' does not name a type; did you mean 'MSColumns'?
     virtual const ROMSColumns & msColumns() const {
                   ^~~~~~~~~~~
                   MSColumns
msvis/CMakeFiles/casa_msvis.dir/build.make:86: recipe for target 'msvis/CMakeFiles/casa_msvis.dir/MSVis/CalVisBuffer.cc.o' failed

schiebel:
I believe ROMSColumns has been removed. MSColumns should be used even for read-only usage.

bennahugo:
Workaround with ```-DBUILD_DEPRECATED=True ``` as dockerfile of casacore/casarest#31

PR #892, opened 2019-04-03
returns a default tile shape for contiguous HDF5 data sets

veggiesaurus:
A previous PR allowed contiguous HDF5 data sets to be opened correctly. However, we ran into issues when iterating through the data set, because the `HDF5DataSet` object's `tileShape` was left as an empty `IPosition`. 

This PR modifies `HDF5Lattice` to return a default tile shape for contiguous datasets, rather than the empty tile shape defined by the dataset. The default tile shape is calculated using `TiledShape::tiledShape`, similarly to how FITS files' tile shape is determined. 

I initially placed this functionality in `HDF5DataSet`, but then had linking issues, as `HDF5DataSet` belongs to `casa_casa` library, while `TiledShape` belongs to the `casa_lattices` library, and including it creates circular dependencies.

### DIFF
--- a/lattices/Lattices/HDF5Lattice.h
+++ b/lattices/Lattices/HDF5Lattice.h
@@ -266,6 +266,7 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
     CountedPtr<HDF5File>    itsFile;
     CountedPtr<HDF5Group>   itsGroup;
     CountedPtr<HDF5DataSet> itsDataSet;
+    IPosition               itsTileShape;
   };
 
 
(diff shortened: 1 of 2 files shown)


gervandiepen:
Are you sure this is what you want?
FITSImage uses TiledFileAccess::makeTileShape to calculate the tile shape.
The difference is that the latter uses the full axis length for the first axes, so access is sequential when reading tile by tile (which most of the Lattices/Images code do). The first one uses TiledShape::defaultTileShape which makes 'square' tiles, thus not filling up the axes probably making access non-sequential when reading tile by tile.

veggiesaurus:
> Are you sure this is what you want?

@gervandiepen quite sure this is the opposite of what I want! Thanks for pointing me in the right direction. I've updated the PR to use `TiledFileAccess::makeTileShape`. 

It would be better to have this applied to `HDF5DataSet`, but I still have the same issue with circular dependencies if I try to do this, but now it's `casa_casacore` -> `casa_tables` -> `casa_casacore`.

veggiesaurus:
rather than recalculating the tile shape each time `tileShape()` is called, it is now calculated once during `HDF5Lattice` construction.

PR #894, opened 2019-04-24
Allow colon and fix recursive bison parsing

gervandiepen:
In LEL it was not possible to have a colon as part of the file name, because it was always interpreted as a separator for mask or region. Now it is possible to use it in a quoted file name if the colon is escaped with a backslash.

Testing the code revealed that ImageExprGram did not behave properly when a recursive parse had to be done. This has been fixed. The same fix has been applied to TableGram.

### DIFF
--- a/images/Images/ImageExprGram.cc
+++ b/images/Images/ImageExprGram.cc
@@ -73,21 +73,52 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
 static const char*  strpImageExprGram = 0;
 static Int          posImageExprGram = 0;
 
+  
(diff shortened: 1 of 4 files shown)


tnakazato:
I've tested the fix and confirmed that a colon in the filename is properly handled if it is escaped by a backslash. I've also confirmed that the filename with colon combined with the mask worked. Tested LELs are as follows:
* filename with colon: `!ISNAN("test\:colon/working/Ori-KL.spw1.cube.I.iter0.image.sd")`
* filename with colon combined with mask: `IIF(MASK("test\:colon/working/Ori-KL.spw1.cube.I.iter0.image.sd:mask_combine"),1,0)`

PR #895, opened 2019-04-26
Add a public getter for the mask array to LCRegionFixed

confluence:
I am @veggiesaurus's colleague on the CARTA project. We need to access the mask arrays of regions in order to implement our custom low-level code for reading region statistics in a fast and memory-efficient way from a rotated (XYZW -> ZYXW) image dataset in our HDF5 schema.


gervandiepen:
I wonder if it wouldn't be better to use const like:
     const ArrayLattice<T>& getMask() const;
Making a copy of the ArrayLattice object is rather cheap (it  uses referencing), but not necessary.
Certainly the function should be const.

PR #900, opened 2019-05-09
Fix memory leak in TableProxy::getColumnDescription

FedeMPouzols:
This fix deletes the TableDesc pointer allocated internally with new, similarly as done in TableProxy::getTableDescription().

Maybe using unique_ptr would be a better approach, but I'm not sure if there's any policy on this in casacore.

getColumnDescription() is used in CASA in the table tool, exposed via the getcoldesc() function.

### DIFF
--- a/tables/Tables/TableProxy.cc
+++ b/tables/Tables/TableProxy.cc
@@ -952,7 +952,7 @@ Record TableProxy::getColumnDescription (const String& columnName,
 					 Bool actual, Bool cOrder)
 {
   // Get the table description.
-  TableDesc* tableDescPtr;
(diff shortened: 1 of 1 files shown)


gervandiepen:
Using unique_ptr is fine and indeed better.

FedeMPouzols:
Aha, thanks, Ger. I've just pushed a commit that uses unique_ptr.
I assume we cannot yet use make_unique / C++14.

gervandiepen:
Better not to use C++14 features yet. Some systems use compilers that are not up-to-date.
Nice that you also changed getTableDescrption.
I'll wait for the check to finish before I merge.

PR #904, opened 2019-05-16
Changed MadfmFunc so that it inherits from ArrayFunctorBase

steve-ord:
I'm porting ASKAPSoft to casacore3 and some of the components are having issues with minor API changes. One is that the template for the slidingArrayMath. Which now requires an ArrayFunctorBase and MadfmFunc does not inherit from it.

Note my tConvert test fails (all others pass) - but it fails on the current master branch as well.


steve-ord:
Changing the <from> branch for easier edits - will close this and reopen a new one

PR #906, opened 2019-05-16
Hotfix/fit gaussian add

steve-ord:
This exposes the errors in the fit parameters to the FitGaussian method. Currently they are hidden and although the solution itself is returned by the fit the errors are not (although they are exposed by the fitter). All this does is add a couple of members to the FitGaussian class to hold both the solution and the errors - so they can be accessed without exposing the underlying fitter. Probably should have used a smart pointer ...


scimath/Fitting/FitGaussian.h
```diff
@@ -177,6 +177,10 @@ class FitGaussian
                 T maximumRMS = 1.0, uInt maxiter = 1024, 
                 T convcriteria = 0.0001);
 
+  // Allow access to the fit parameters from this class
+  Matrix<T> solution(){return itsSolutionParameters;};
+  Matrix<T> errors(){return itsSolutionErrors;};
```
gervandiepen: Probably better to return both as
const Matrix<T>&

scimath/Fitting/FitGaussian.tcc
```diff
@@ -440,9 +441,10 @@ Matrix<T> FitGaussian<T>::fit(const Matrix<T>& pos, const Vector<T>& f,
             //best fit so far - write parameters to solution matrix
             for (uInt g = 0; g < itsNGaussians; g++) {  
               for (uInt p = 0; p < ngpars; p++) {
-                solutionparameters(g,p) = solution(g*ngpars+p);
+                itsSolutionParameters(g,p) = solution(g*ngpars+p);
               }
             }
+            itsSolutionErrors = errors();
```
gervandiepen: I don't understand since errors() returns itsSolutionErrors, thus assigns to itself.
It looks as if the errors are not set.

scimath/Fitting/FitGaussian.h
steve-ord: Agreed

scimath/Fitting/FitGaussian.tcc
steve-ord: Yup - my fault - I meant to access the getErrors method from the fitter.

steve-ord: I'm about to wrap up for the day - but I'll fix up both of these issues asap. Thx

steve-ord:
Ger,

I have fixed up the const for the gets. Made sure the errors() method of the fitter is actually being called ;-) and add a check to the test case.

Cheers,

Steve

gervandiepen:
Nice.

PR #907, opened 2019-05-16
Extract Python strings via casacore::String not std::string

bmerry:
There is a Unicode converter for casacore strings, but not for
std::string (under Python 2). Rather than define one for std::string
(which would then be foisted on all extensions in the process), change
the few places I could find that use extract<std::string> to use
extract<String> instead (most of them immediately converted the result
to a String anyway).


bmerry:
Seems I can't assign, but this is for @gijzelaerr.

bmerry:
Just realised I'd branched from an older version. I'll rebase and update the PR.

bmerry:
Rebasing is done.

gijzelaerr:
perfect! Thanks a gazillion bruce!

PR #909, opened 2019-05-16
add support for unicode numpy array

gijzelaerr:
This adds support for unicode strings in numpy arrays.

### DIFF
--- a/python/Converters/PycArrayComCC.h
+++ b/python/Converters/PycArrayComCC.h
@@ -359,7 +359,7 @@
         Array<uChar>* uarr = static_cast<Array<uChar>*>(varr);
 	convertArray (res, *uarr);
 	return ValueHolder(res);
-      } else if (PyArray_TYPE(po) == NPY_STRING) {
+      } else if (PyArray_TYPE(po) == NPY_STRING || PyArray_TYPE(po) == NPY_UNICODE) {
 	size_t slen = 0;
 	if (nd > 0) {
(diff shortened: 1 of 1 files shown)


bmerry:
Have you tested what actually gets written to the file? From eye-balling the code, it looks like you're dumping numpy's raw Unicode representation as bytes, and I think numpy uses UCS-4 for Unicode.

gijzelaerr:
no, I didn't test, I'm unfortunately out of time to work on this.

bmerry:
Ok, I'll give it a test when I have a chance to make sure it's doing something sensible.

bmerry:
I tried this and it didn't work. When you read back the column, you get only the first character. I think what's happening is that numpy uses UCS-4 in memory, which means that "Hello" becomes `b"H\0\0\0e\0\0\0l\0\0\0l\0\0\0o\0\0\0"` (on little endian) and then it gets treated as a null-terminated string.

The text will need to be re-encoded as UTF-8. I don't know if casacore has any utilities for this. The numpy Python API has [numpy.core.defchararray.encode](https://docs.scipy.org/doc/numpy/reference/generated/numpy.core.defchararray.encode.html#numpy.core.defchararray.encode), but I didn't see any numpy C API to do the same thing efficiently.

gijzelaerr:
discarding in favor of https://github.com/casacore/casacore/pull/912

PR #912, opened 2019-05-17
Handle numpy Unicode arrays

bmerry:
This converts numpy Unicode arrays by re-encoding from UTF-32 to UTF-8.
It is intended to supersede #909.

### DIFF
--- a/python/Converters/PycArrayComCC.h
+++ b/python/Converters/PycArrayComCC.h
@@ -365,6 +365,12 @@
 	  slen = PyArray_STRIDES(po)[nd-1];
 	}
 	return ValueHolder (ArrayCopyStr_toArray(shp, PyArray_DATA(po), slen));
+      } else if (PyArray_TYPE(po) == NPY_UNICODE) {
+	size_t slen = 0;
+	if (nd > 0) {
+	  slen = PyArray_STRIDES(po)[nd-1];
+	}
+	return ValueHolder (ArrayCopyUnicode_toArray(shp, PyArray_DATA(po), slen));
       }
       break;
     }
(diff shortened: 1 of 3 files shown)


bmerry:
@gijzelaerr do you have time to look at this? Or is there someone else it should be assigned to?

gijzelaerr:
Thanks bruce, looking good! Do you still have the code you used for verifying that it works? We can use that then to make a test for python-casacore.

bmerry:
Yes, it would be good to have a stronger unit test, which basically means reading back the data (and I would say also using more than one row, so that one tests the stride calculations). For now I've just ran a script by hand to eyeball it, but if you have time to update your test PR that would be great.
```
#!/usr/bin/env python3

from casacore.tables import table, makescacoldesc, maketabdesc
import numpy as np

col1 = makescacoldesc('mycol1', 'test', valuetype='string')
col2 = makescacoldesc('mycol2', 'test', valuetype='string')
t = table('blah.ms', maketabdesc([col1, col2]))
t.addrows(2)
t.putcol('mycol1', np.array(['hello', 'world']))
t.putcol('mycol2', ['hello', 'world'])
t.close()

t = table('blah.ms')
print(t.getcol('mycol1'))
print(t.getcol('mycol2'))
```

bmerry:
Should I merge, or should I wait until the unit test in python-casacore is updated?

gijzelaerr:
i'll just merge it!

gijzelaerr:
added the test (which passes) https://github.com/casacore/python-casacore/commit/5e9c50033a54596b7ea6a2577763ec2eef2bfed2

Issue #854, opened 2018-12-21
update_measures broken


gijzelaerr:
When i run it my system i get:
```
Traceback (most recent call last):                   
  File "./update_measures", line 32, in <module>     
    default=detect_domain(),                         
  File "./update_measures", line 13, in detect_domain
    domain = socket.gethostname().split(".")[-2]     
IndexError: list index out of range                  
```
Now you could argue that the machine is wrongly configured since no FQDN is set, but that is the cast with most machines. even if i specify the domain the script crashes since the code still runs.

This works:
```
def detect_domain():                     
    choices = ["atnf", "astron", "nrao"] 
    hostname = socket.gethostname()      
    if '.' in hostname:                  
        domain = hostname.split(".")[-2] 
        if domain in choices:            
            return domain                
    return choices[0]                    
```

But then all options fail again:
```
λ  ./update_measures -l astron data                                                                             
Traceback (most recent call last):                                                                              
  File "./update_measures", line 106, in <module>                                                               
    ftp.cwd(datadirs[OPTS.location])                                                                            
  File "/usr/lib/python2.7/ftplib.py", line 574, in cwd                                                         
    return self.voidcmd(cmd)                                                                                    
  File "/usr/lib/python2.7/ftplib.py", line 256, in voidcmd                                                     
    return self.voidresp()                                                                                      
  File "/usr/lib/python2.7/ftplib.py", line 231, in voidresp                                                    
    resp = self.getresp()                                                                                       
  File "/usr/lib/python2.7/ftplib.py", line 226, in getresp                                                     
    raise error_perm, resp                                                                                      
ftplib.error_perm: 550 Failed to change directory.                                                              
                                                    
λ  ./update_measures -l atnf data                                                                               
Traceback (most recent call last):                                                                              
  File "./update_measures", line 106, in <module>                                                               
    ftp.cwd(datadirs[OPTS.location])                                                                            
  File "/usr/lib/python2.7/ftplib.py", line 574, in cwd                                                         
    return self.voidcmd(cmd)                                                                                    
  File "/usr/lib/python2.7/ftplib.py", line 256, in voidcmd                                                     
    return self.voidresp()                                                                                      
  File "/usr/lib/python2.7/ftplib.py", line 231, in voidresp                                                    
    resp = self.getresp()                                                                                       
  File "/usr/lib/python2.7/ftplib.py", line 226, in getresp                                                     
    raise error_perm, resp                                                                                      
[...]

PR #914, opened 2019-05-21
Remove update measures (#854)

gijzelaerr:


tammojan:
I agree this script is so outdated and apparently unused that it can be removed.

dpetry:
Just for my better understanding: wasn't this script necessary to perform the update of the measures tables at ASTRON? I vaguely remember that Gijs was working on fixing the one problem it had.

gijzelaerr:
I think you are confused with the work @sabourke did to fix `measuresdata`. That one correctly updates the measures data, I assume this is an old, broken and deprecated script that did something similar.

tammojan:
> Just for my better understanding: wasn't this script necessary to perform the update of the measures tables at ASTRON? I vaguely remember that Gijs was working on fixing the one problem it had.

Indeed, the now removed script only was to download measures data from ASTRON (probably similar to the cron job you have running). To create updated measures data, ASTRON uses the script `measuresdata.csh`, which is still there.

dpetry:
Excellent. Thanks.

Issue #899, opened 2019-05-03
Incremental Storage Manager columns don't support slicing


kettenis:
The Incremental Storage Manager (class ISMColumn) does not implement the getSliceTYPEV() methods.  As a result the default implementation in the StManColumn gets used which throws an "invalid operation exception".

Since the CASA calibration framework uses slicing, and LOFAR Measurement Sets use the Incremental Storage Manager for some columns, the result is a fatal error:

2019-04-17 15:53:39     SEVERE  Calibrater::solve       Caught exception: Table DataManager error: Invalid operation: StManColumn::getSlicefloatV not possible for column SIGMA

This can be reproduced with the fringefit task in CASA 5.3 or later:

`fringefit(vis='L1327+5504.ms', caltable='test.fj', solint=60)`

Other CASA calibration tasks (e.g. gaincal) produce similar errors.

A workaround (provided by members of the LOFAR long baseline working group) is to copy the columns that use the ISM before running a CASA calibration task.

gervandiepen:
Mark,
Ik begrijp niet waardoor de fout ontstaat. Zou je
   showtableinfo in='L1327+5504.ms'
kunnen doen en mij de output doen toekomen.

gervandiepen:
I finally managed to reproduce the problem. It occurs in getColumnCells taking a ColumnSlicer which was added by Jim Jacobs quite a while ago.
It does not check if getSlice can actually be used. I'll change it.

kettenis:
Thanks Ger,

That seems to fix the issue. I rebuilt the casacore bundled with CASA and successfully ran the casa fringefit task on a small LOFAR data set provided to us by members of the LOFAR long baseline working group.

PR #915, opened 2019-05-21
prepare for 3.1.1

gijzelaerr:
Everybody ok with this? 

My only worry is that the ABI changed between this release and 3.0. @gervandiepen were there any ABI changes in the shared libraries? I don't see them this quickly.


README.md
```diff
@@ -1,12 +1,22 @@
...
+
+Casacore is now part of Debian and Ubuntu, use apt to search for the package you need:
+```bash
+$ apt-get update
+$ apt-get search casacore
+```
+
+## Building from source
+
+### getting the source code
```
ludwigschwardt: Getting

README.md
```diff
@@ -1,12 +1,22 @@
...
 # Casacore
 
-A suite of c++ libraries for radio astronomy data processing.
+A suite of C++ libraries for radio astronomy data processing.
 
 
 # Installation
 
-## Obtaining the source
+## Debian / Ubuntu
```
ludwigschwardt: We could add a section for macOS, something like:

## macOS

Use Homebrew with the `ska-sa` tap to install casacore:

```bash
$ brew tap ska-sa/tap
$ brew install casacore
```

gijzelaerr:
> For the record, I've improved the cmake Python detection modules in this gist:
> 
> https://gist.github.com/ludwigschwardt/0bfaef7b2c6832fb018332742e14924e
> 
> Unfortunately it needs at least cmake 3.12 (although it can be made to work with cmake 3.7 if we copied the relevant modules into our local stash). This is probably too new for things like CentOS 6 and other places where casacore shows up.

can you open a PR for that change? We can then hopefully accept it at some point.

CHANGES.md
```diff
@@ -1,3 +1,27 @@
+# 3.1.1
+
+## General
+ - Only use DataManager.get/putSlice if possible (#901)
+ - Incremental Storage Manager columns don't support slicing (#899)
```
gervandiepen: This line could be removed (#901 solves issue #899)

gervandiepen:
The most important change is the removal of classes Map, List, etc. which changes the ABI (and API). However, they were only used internally and by CASA, so that change does not really affect the ABI.

gervandiepen:
I've created PR#916 to solve the PyLong/TpInt issue. You could merge it before making the release.

gervandiepen:
Could you merge PR177 of python-casacore before making a python-casacore release?

gijzelaerr:
done!

PR #916, opened 2019-05-22
Always use TpInt64 for PyLong

### LINKED ISSUE #911: tConvert test fails on OSX
boost-python can't be found on non-standard location.

The test has been disabled on travis.

gervandiepen:
Do not treat PyLong as TpInt for Python3

Close #911

### DIFF
--- a/casa/Containers/RecordInterface.cc
+++ b/casa/Containers/RecordInterface.cc
@@ -347,89 +347,67 @@ void RecordInterface::get (const RecordFieldId& id, String& value) const
 }
 void RecordInterface::get (const RecordFieldId& id, Array<Bool>& value) const
 {
-    Int whichField = idToNumber (id);
-    const Array<Bool>& array = *(const Array<Bool>*)
-	                            get_pointer (whichField, TpArrayBool);
+    Array<Bool> array = toArrayBool (id);
     value.resize (array.shape());
     value = array;
 }
 void RecordInterface::get (const RecordFieldId& id, Array<uChar>& value) const
 {
-    Int whichField = idToNumber (id);
-    const Array<uChar>& array = *(const Array<uChar>*)
-	                            get_pointer (whichField, TpArrayUChar);
+    Array<uChar> array = toArrayuChar (id);
     value.resize (array.shape());
     value = array;
 }
 void RecordInterface::get (const RecordFieldId& id, Array<Short>& value) const
 {
-    Int whichField = idToNumber (id);
-    const Array<Short>& array = *(const Array<Short>*)
-	                            get_pointer (whichField, TpArrayShort);
+    Array<Short> array = toArrayShort (id);
     value.resize (array.shape());
     value = array;
 }
 void RecordInterface::get (const RecordFieldId& id, Array<Int>& value) const
 {
-    Int whichField = idToNumber (id);
-    const Array<Int>& array = *(const Array<Int>*)
-	                            get_pointer (whichField, TpArrayInt);
+    Array<Int> array = toArrayInt (id);
     value.resize (array.shape());
     value = array;
 }
 void RecordInterface::get (const RecordFieldId& id, Array<uInt>& value) const
 {
-    Int whichField = idToNumber (id);
-    const Array<uInt>& array = *(const Array<uInt>*)
-	                            get_pointer (whichField, TpArrayUInt);
+    Array<uInt> array = toArrayuInt (id);
     value.resize (array.shape());
     value = array;
 }
 void RecordInterface::get (const RecordFieldId& id, Array<Int64>& value) const
 {
-    Int whichField = idToNumber (id);
-    const Array<Int64>& array = *(const Array<Int64>*)
-	                            get_pointer (whichField, TpArrayInt64);
+    Array<Int64> array = toArrayInt64 (id);
     value.resize (array.shape());
     value = array;
 }
 void RecordInterface::get (const RecordFieldId& id, Array<float>& value) const
 {
-    Int whichField = idToNumber (id);
-    const Array<float>& array = *(const Array<float>*)
-	                            get_pointer (whichField, TpArrayFloat);
+    Array<float> array = toArrayFloat (id);
     value.resize (array.shape());
     value = array;
 }
 void RecordInterface::get (const RecordFieldId& id, Array<double>& value) const
 {
-    Int whichField = idToNumber (id);
-    const Array<double>& array = *(const Array<double>*)
-	                            get_pointer (whichField, TpArrayDouble);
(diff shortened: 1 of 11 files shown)


gijzelaerr:
@gervandiepen did you actually run the `python-casacore` test suite on this code? it fails miserably on my side.

gervandiepen:
I did run the test on a Liunux machine with Python2.7 which went fine.
Does it fail for you with python3?

gervandiepen:
All 6 PR builds succeeded.

gijzelaerr:
they fail for both py2 and py3 for me:

https://gist.github.com/gijzelaerr/7169d9907357967aa1c1535174395c96

when i build against current master only the long int test fails.

gervandiepen:
How do you run the python-casacore test suite? I cannot run ctest or so.

gijzelaerr:
install and run nosetest or pytest

gervandiepen:
When running tConvert.py (in casacore) with python, it runs well. Using
pytest it finds the value type to be int, while when using python itself it
has type long. So it is the question if pytest is doing something special?

gijzelaerr:
first of all, are you running the `python-casacore` test suite?

secondly, maybe pytest is associated with a different python interpreter linked to a different casacore/python-casacore?

gervandiepen:
I ran
  pytest test_convert.py
and
  python tConvert.py
The scripts are basically the same; I assume you cloned test_convert.py
from tConvert.py.

They definitely use the same casa_python library as I added a few print
statement to it which appear in both tests.

pytest seems to use the same python interpreter. It uses python2 and
/usr/bin/python is a symlink to python2 (and further to python2.7). So I
don't know.
I'll try to understand more tomorrow. Also what pytest is doing exactly.

-bash-4.2$ more /usr/bin/pytest
#!/usr/bin/python2

# -*- coding: utf-8 -*-
import re
import sys

from pytest import main

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
-bash-4.2$
-bash-4.2$ which python
/usr/bin/python
-bash-4.2$ ls -l /usr/bin/python
lrwxrwxrwx. 1 root root 7 May 16  2017 /usr/bin/python -> python2

gijzelaerr:
pytest collects and runs all (unit) tests in a Python project. If you just run it without arguments in the `python-casacore` folder it should collect all tests and give output similar to:

https://gist.github.com/gijzelaerr/7169d9907357967aa1c1535174395c96

gijzelaerr:
To clarify, i've ran the `python-casacore` test suite with py2 and py3 for the master branch and the issue 911 branch. py2 + master, py3 + master, py2 + issue 911 all give 1 failed test: `test_convert.py:140: OverflowErrow`.

py3 + issue 911 branch has 26 failed tests:

https://gist.github.com/gijzelaerr/7169d9907357967aa1c1535174395c96

gervandiepen:
I've submitted a fix.
Now pytest runs fine (for Python2.7) for most tests, but test_unicode fails for a numpy array of Unicode strings. This is because the converters do not support an array of unicode strings. I doubt if it's worth the effort. What do you think?

gijzelaerr:
That unicode in numpy arrays problem has been fixed in this PR: https://github.com/casacore/casacore/pull/912

Which has been merged into master. If you merge master into this branch the py2 test suite completes successfully now! For py3 I still have 26 failing tests, mostly in fitting and functional:
```
tests/test_convert.py ....                                                                                                                                                                                                                                                  [  5%] 
tests/test_fitting.py FFFF.F.FF                                                                                                                                                                                                                                             [ 19%] 
tests/test_functionals.py FFFFFFFFFFFFFFFF.                                                                                                                                                                                                                                 [ 44%] 
tests/test_image.py ............                                                                                                                                                                                                                                            [ 62%] 
tests/test_measures.py .                                                                                                                                                                                                                                                    [ 64%] 
tests/test_quanta.py .                                                                                                                                                                                                                                                      [ 65%] 
tests/test_table.py .......F..FF......                                                                                                                                                                                                                                      [ 92%] 
tests/test_unicode.py ...                                                                                                                                                                                                                                                   [ 97%] 
tests/test_util.py ..                  
```
I don't see these errors when i test py3 against the latest master.

gervandiepen:
I fixed the test_tables failures.
It looks as if the many fitting and functionals failures are due to the object not being constructed.
I think both issues have the same reason which is that the expected data type in the C++ code mismatches the data types given by Python. I've changed the code to convert the data types as needed. Hopefully all tests pass now for all cases.

gijzelaerr:
We are down to 23 failing tests now for python3, python2 works.

```
23 failed, 44 passed in 2.13 seconds


tests/test_convert.py ....                                                                                       [  5%]
tests/test_fitting.py FFFF.F.FF                                                                                  [ 19%]
tests/test_functionals.py FFFFFFFFFFFFFFFF.                                                                      [ 44%]
tests/test_image.py ............                                                                                 [ 62%]
tests/test_measures.py .                                                                                         [ 64%]
tests/test_quanta.py .                                                                                           [ 65%]
tests/test_table.py ..................                                                                           [ 92%]
tests/test_unicode.py ...                                                                                        [ 97%]
tests/test_util.py ..                                                                                            [100%]
```

gervandiepen:
Does pytest show if the construction of the functionals object succeeds,
and if not, why it fails?
Since all failing tests report None.something, it looks as if the
construction fails.

gijzelaerr:
pytests doesn't do much special apart from gathering the tests and running them. Nothing magical, it is just python code.

gervandiepen:
It captures all output, so you don't see why something fails.
   pytest --show-capture=all
gives you the output of a failing test, but not of a failing constructor in
setupClass. And I want to see that part.

rtobar:
Just saw this discussion, and I thought I might dive in.

From experience, if something fails in pytest at setup time you would get a clear printing of the error. If one of the modules fails to load at import time you should also get a clear error about it.

If the testing code doesn't rely on pytest itself you could also try using a different unit test runner. The built-in unittest module already provides something very much like via `python -munittest discover`.

gervandiepen:
I think (and now really hope) I found the problem which is in FunctionHolder. It is very picky on the data types that can be given and does not accept a long where an int is expected. Python3 will usually have a long.
I have to check a bit more tomorrow though.

gervandiepen:
I've committed the changes and ensured FunctionHolder works well with data type PyLong. Give it another try.

gijzelaerr:
Yes! Well done Ger. Lets wait for the merged travis build to finish and then this is good to go! Thanks!

PR #919, opened 2019-05-22
Improve CMake FindPython routines

ludwigschwardt:
CMake 3.12 introduced the FindPython2 and FindPython3 modules which
supercede the old and buggy FindPythonLibs and FindPythonInterp modules.
It finds consistent versions of the interpreter, development artifacts
and NumPy include directories in one go, and does it separately for
Python 2 and 3.

I've kept the original CMake variable names since I'm not sure where
they are used, but some of them may probably be removed or refactored.
This has been tested on a Homebrew setup (but not on Linux), where it
fixes detection errors using the previous commit.

By copying the appropriate modules from CMake, this functionality could
also be made to work on the older CMake 3.7 (although this might not be
old enough for all installations of casacore).

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -1,32 +1,13 @@
 message(STATUS "Looking for python2 specific environment...")
 
-# tempororarly set variables used by findpython
-if (PYTHON2_EXECUTABLE)
-    set(PYTHON_EXECUTABLE ${PYTHON2_EXECUTABLE} CACHE FILEPATH "")
-endif()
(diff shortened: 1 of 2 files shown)


ludwigschwardt:
I added this for posterity on @gijzelaerr's request, but it will probably only be useful to casacore in the medium term (unless one can somehow do a conditional CMake module that also supports both old and new versions).

ludwigschwardt:
I'm attempting the conditional CMakeLists option to have a once-size-fits-all solution, so withdrawing this for now...

PR #922, opened 2019-05-24
Improve CMake FindPython

ludwigschwardt:
CMake 3.12 introduced the `FindPython2` and `FindPython3` modules which supercede the old and buggy `FindPythonLibs` and `FindPythonInterp` modules. It finds consistent versions of the interpreter, development artifacts and NumPy include directories in one go, and does it separately for
Python 2 and 3.

I've kept the original CMake variable names since I'm not sure where they are used, but some of them may probably be removed or refactored. This has been tested on a Homebrew setup (but not on Linux), where it fixes detection errors using the previous commit.

By copying the appropriate modules from CMake, this functionality could also be made to work on the older CMake 3.7 (although this might not be old enough for all installations of casacore).

I've also kept the original versions of the CMakeLists.txt files and switch to them if CMake is older than 3.12, so this should be backwards compatible (although having four copies of everything is suboptimal...).


ludwigschwardt:
BTW, the tests passed - the first one timed out after 50 minutes though, hence the red cross...

ludwigschwardt:
I've rerun the tests on Travis and now they pass (maybe we should increase the timeout to 60 minutes per test?)

ludwigschwardt:
Can we try this out for 3.1.1 too? It will really help my Homebrew formula along (and maybe even your Travis build).

gijzelaerr:
@tammojan @gervandiepen what do you guys think? Include in 3.1.1? Looks good to me.

gervandiepen:
Looks ok to me

ludwigschwardt:
I tried to maintain the original script for older CMakes and at least Travis is happy, so this should hopefully be fine 🤞

gervandiepen:
My build with Python2.7 was happy as well.

Issue #920, opened 2019-05-24
Memory leak in MSSelection::initErrorHandler()


FedeMPouzols:
As MSSelection::setErrorHandler() says:
"We make a copy (clone) of the supplied error handler pointer and manage that pointer internally."

The caller is responsible for the handler passed to it. But in the same class, MSSelection::initErrorHandler() forgets to release the handlers that it allocates with new:
https://github.com/casacore/casacore/blob/master/ms/MSSel/MSSelection.cc#L502-L536.
Rather than allocating the temporary objects dynamically I 'd opt for creating them as a local ver on the stack, as they do in tMSSelection.

This shows up in valgrind checks that I'm doing on some leaky tclean commands. For example:

```
==11172== 56 bytes in 1 blocks are definitely lost in loss record 11,802 of 19,304
==11172==    at 0x4C2A1E3: operator new(unsigned long) (vg_replace_malloc.c:334)
==11172==    by 0x2828553D: casacore::MSSelection::initErrorHandler(casacore::MSSelection::MSExprType) (in /lustre/naasc/users/scastro/casa/el7/casa-prerelease-5.6.0-15.el7/lib/libcasa_ms.s
o.2637.195.194)
==11172==    by 0x2828800F: casacore::MSSelection::MSSelection() (in /lustre/naasc/users/scastro/casa/el7/casa-prerelease-5.6.0-15.el7/lib/libcasa_ms.so.2637.195.194)
==11172==    by 0x24D3C248: casa::SynthesisImagerVi2::selectData(casa::SynthesisParamsSelect const&) (in /lustre/naasc/users/scastro/casa/el7/casa-prerelease-5.6.0-15.el7/lib/libsynthesis.s
o.2637.195.194)
==11172==    by 0x2CB73905: casac::synthesisimager::selectdata(casac::record const&) (in /lustre/naasc/users/scastro/casa/el7/casa-prerelease-5.6.0-15.el7/lib/libtools.so.2637.195.194)
==11172==    by 0x3B421737: _wrap_synthesisimager_selectdata (in /lustre/naasc/users/scastro/casa/el7/casa-prerelease-5.6.0-15.el7/lib/python2.7/__casac__/_synthesisimager.so)
==11172==    by 0x4F1B485: PyEval_EvalFrameEx (in /lustre/naasc/users/scastro/casa/el7/casa-prerelease-5.6.0-15.el7/lib/libpython2.7.so.1.0)
[...]

PR #924, opened 2019-05-25
Fixing tMSSelections leaks found by valgrind

gervandiepen:
Running valgrind on tMSSelection revealed a few leaks and an array overrun error.
This PR solves the leaks. It also removes some needless comparisons in some MSxxxGram.yy files.


FedeMPouzols:
This fixes #920 and more issues.

PR #932, opened 2019-06-11
Feature/regrid wcsmany

steve-ord:
This is a feature that we have added locally as a patch to ImageRegrid.tcc that allows the coordinate lookups to be done as bulk (ToMany) calls which are <much> faster than performing this a pixel at a time.

I have to admit that this patch predates me and I am not completely familiar with its workings - which is one of the reasons this is a draft request. It basically it turns the useMachine test into a conditional for a choice between the original method or a bulk coordinate conversion. 

All the tests pass. But they likely do not test this case.
 
I've put this up as a draft to give those a chance to look at it while I add a test that covers the new branch.


steve-ord:
Closing this one - to make a new one - with the same fix off the current master so there is only a single checkin so no-one gets to worried.

PR #936, opened 2019-07-01
Small TaQL improvements

gervandiepen:
A few improvements to TaQL:
- Improved error messages
- Better state handling in TableGram.ll/yy, so unquoted table name can be given in concatenation
- function name 'ampl' as synonym for 'amplitude'

### DIFF
--- a/tables/TaQL/TableGram.yy
+++ b/tables/TaQL/TableGram.yy
@@ -268,7 +268,7 @@ topcomm:   topcomm1
          | topcomm1 SEMICOL
          ;
 
-/* A command can be preceeded by the TIME keyword and style arguments */
+/* A command can be preceded by the TIME keyword and style arguments */
 topcomm1:  command
          | sttimcoms command
          ;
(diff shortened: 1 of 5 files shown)


tables/TaQL/TableGram.yy
```diff
@@ -1333,16 +1343,19 @@ concslist: NAME {
          ;
 
 /* Concat table persistency using GIVING or INTO is optional */
+concgiven: GIVING {
+               setTABLENAMEstate();
+           }
+         | INTO tabname {
+               setTABLENAMEstate();
+           }
```
tammojan: I would say either both these lines should have `tabname`, or none..

tables/TaQL/TableGram.yy
```diff
@@ -1294,14 +1305,13 @@ tfnamen:   tfname {
    concatenate and the GIVING/INTO to make the concat table persistent.
 */
 tfname:    tfcommand {
-	       theFromQueryDone = True;
```
tammojan: How did you eliminate the need for `theFromQuery`? To which other change does this correspond?

gervandiepen:
The functions setEXPRstate and setTABLENAMEstate do a much better and
clearer job.
The parser sets those states in the lexer at appropriate points.

PR #938, opened 2019-07-03
CAS-12548

dmehring:
* added UInt64 type case check for miscInfo record
* fixed log message not being logged when miscInfo key-val
      pair not being written because of sloppy coding
* replaced the multitude of image.miscInfo() calls by calling
      image.miscInfo() once and capturing value, and then
      using that, because you know, why call the same method
      over and over when you can just capture and store the
      value once?
* fixed some atrocious indentation, the rest of the file really
      needs to be indented properly. It is a mess to maintain as is.

### DIFF
--- a/images/Images/ImageFITS2Converter.cc
+++ b/images/Images/ImageFITS2Converter.cc
@@ -1074,101 +1074,111 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
     //
     // Add in the fields from miscInfo that we can
     //
-    const uInt nmisc = image.miscInfo().nfields();
+    const auto miscInfo = image.miscInfo();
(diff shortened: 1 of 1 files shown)


dpetry:
I would like to mention that I did not write this code. This was written before my time.
I suspect the bad indentation might have come from porting this through different editors in the more distant past.

dmehring:
> I would like to mention that I did not write this code. This was written before my time.
> I suspect the bad indentation might have come from porting this through different editors in the more distant past.

Of course @dpetry . I didn’t mean to imply that you had. I figured that this was written long, long ago and you just inherited it, much as I’ve inherited the abominable UVFITS code. 

Thanks for the quick review and merge.

PR #945, opened 2019-07-16
Change row numbers to 64 bits

gervandiepen:
CTDS has been changed to use 64-bit instead of 32-bit row numbers. It has been typedef-ed as rownr_t.
Of course it is backward compatible, thus existing tables can still be used. Although the DataManager API has changed, it is done in a backward compatible way, thus external data managers should build without problems.
New tables will be written as 32-bit row numbers if the table contains < 2**32 rows, so for most tables it will also be forward compatible.

Most Casacore software has been adapted, but some ms and msfits classes need to be visited. Also an external package such as CASA should be build against this branch to ensure it builds well. The preprocessor variable IMPLICIT_CTDS_32BIT can be defined in such a package to allow for implicit conversion to/from Vector<uInt> row numbers.


tammojan:
@rtobar I can imagine that in the recent Adios2 fixes also migrations from `unsigned int` to `rownr_t` are necessary. I have not tested this branch on a system with Adios2. Could you perhaps try this branch with Adios2?

rtobar:
@tammojan yes, fixes will indeed be needed. I'll work on this these coming days and will let you know once I'm finished with the porting.

tammojan:
Thanks @gervandiepen for this enormous effort! This makes casacore ready for the really large datasets!

PR #949, opened 2019-07-24
More efficient ToMany WCS coordinate conversions in ImageRegrid

steve-ord:
Re-opened this with a commit against the latest master. 

This is a feature that we have added locally as a patch to ImageRegrid.tcc that allows the coordinate lookups to be done as bulk (ToMany) calls which are faster than performing this a pixel at a time.

I have to admit that this patch predates me and I am not completely familiar with its workings - which is one of the reasons this has taken a while to put in - and bounced around a bit. It basically it turns the useMachine test into a conditional for a choice between the original method or a bulk coordinate conversion.

All the tests pass. It has been used routinely against casacore 2+ in ASKAPSoft for many years.

I've checked whether the current test case (tImageRegrid) hits this code and it does.


steve-ord:
Possible problem with a regression test. <DONT MERGE> till I confirm whether this patch is as fault

tammojan:
Hi Steve, do you have an update on this?

steve-ord:
Hi,

Just got back to this. On my local machine I have a failing regression test (tConvert) but I don't think it has anything to do with this patch - looking at it now. I've just merged master in and we'll see how the latest Travis build goes.

steve-ord:
A bit of trouble with your osx travis set up - in particular the python3 - I think I've sorted this and the checks <should> pass

steve-ord:
This is a failing tHostInfo test on one of your platforms. I suspect all your jobs are failing with this. I'll have a quick look

tammojan:
Travis is indeed broken on MacOS, probably due to a change in the build environment. You're more than welcome to fix that, but it doesn't need to be in this pull request.

If the WCS change works on Linux test machines, I'm happy to review/merge it; the change should be quite OS independent.

steve-ord:
Hi - I've fixed the build environment problems - and all the linux test pass. One of the osx tests failed on your gcc builder (apparently the 5 second sleep in tHostInfo took substantially less or more than 5 seconds). For completeness I'm just having a look at that. I'll wrap it up today.

rtobar:
This would probably be fit for merging now that the MacOS environment has been stabilized (#1007) and the tests made more robust (#1008).

tammojan:
I rebased the relevant commits on the current master, this PR is superseded by #1009 .

PR #950, opened 2019-07-24
Fit2D::estimate using a Masked array in the minMax call. Needs the right template

steve-ord:
Ok this is a simple one.

One of my tests was failing as Fit2D could not find the correct minMax template for a MaskedArray. I found I had to include the correct header.

Its a one line addition. All the current tests pass. None of them seem to test estimate. Hang on. I'll extend the tFit2D test to do that.....


steve-ord:
Also added some test coverage for the estimate method.

PR #952, opened 2019-07-25
Improve error when executing `cp` fails

aroffringa:
In lofar-astron/DP3#177 it is reported that casacore sometimes fails with
std exception detected: (/opt/lofar/casacore/src/casa/OS/Directory.cc : 304) Failed AlwaysAssert system(command.chars()) == 0

To avoid having to check the code what this means, I've improved that error msg a bit. The cause is a failing `cp` command (which could still be for various reasons...)

### DIFF
--- a/casa/OS/Directory.cc
+++ b/casa/OS/Directory.cc
@@ -301,7 +301,10 @@ void Directory::copy (const Path& target, Bool overwrite,
     String command("cp -r '");
     command += itsFile.path().expandedName() + "' '" +
                targetName.expandedName() + "'";
-    AlwaysAssert (system(command.chars()) == 0, AipsError);
+    int result = system(command.chars());
+    if(result != 0)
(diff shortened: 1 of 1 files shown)


casa/OS/Directory.cc
```diff
@@ -301,7 +301,10 @@ void Directory::copy (const Path& target, Bool overwrite,
     String command("cp -r '");
     command += itsFile.path().expandedName() + "' '" +
                targetName.expandedName() + "'";
-    AlwaysAssert (system(command.chars()) == 0, AipsError);
+    int result = system(command.chars());
+    if(result != 0)
```
tammojan: Sorry to nag about this, but casacore uses curly braced in if statements. Please keep the code style consistent.

aroffringa: Sure, fair point.

PR #955, opened 2019-09-09
CAS-12627 

dmehring:
* bug fix ImageBeamSet::setBeam() for chan >= 0 && stokes < 0
* made ImageBeamSet::fromRecord() more efficient for images with multiple beams

### DIFF
--- a/images/Images/ImageBeamSet.cc
+++ b/images/Images/ImageBeamSet.cc
@@ -48,7 +48,7 @@ ImageBeamSet::ImageBeamSet() :
 
 ImageBeamSet::ImageBeamSet(const Matrix<GaussianBeam>& beams) :
     _beams(beams) {
-    _calculateAreas();
+        _calculateAreas();
 }
 
 ImageBeamSet::ImageBeamSet(const GaussianBeam& beam) :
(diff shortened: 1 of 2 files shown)


dpetry:
Hi Dave, looks like some clang compilation problem?
Maybe a fault of Travis, not your changes ...

tammojan:
> Hi Dave, looks like some clang compilation problem?
> Maybe a fault of Travis, not your changes ...

It's again a travis fault; this time a problem with Boost-python on MacOS. Clang works fine in the Travis Linux build.

dpetry:
So, what next?

gervandiepen:
I noticed you changed line 248 in the .cc file (which seems right to me), but shouldn't line 227 be changed in the same way?

dmehring:
Yes I think you are right Ger. It appears that branch will never be run because chan must be negative but the two values against which it is compared are always non-negative. I will have a look. Thanks for raising this point.

dmehring:
I did the bug fix and a refactor in a single commit. I probably should have split those into separate commits for clarity. Let me know if you'd rather I do that and I'll give it a try.

tammojan:
Two tests now timed out, I've restarted the travis jobs. I'm fine with the joint commit. I'll wait an hour for the tests to finish, and then merge.

PR #956, opened 2019-09-11
Change MSIter sorting algorithm to ParSort

cquike:
This PR changes the sorting algorithm of the underlying tables in the MSIter class from the current 
QuickSort to ParSort.

QuickSort algorithm performs significantly slower than the
ParSort algorithm added a few years ago to CASA.
The following benchmarks which simply run a MSIter loop with different
sorting columns show the performance on a relatively small MS
(1M rows) and a larger one (11M rows) and show that ParSort always
performs better by at least a factor of 2x in the case of using
1 thread but it can reach up to a factor of 10x in some cases if
all cores are used.
For comparison, also the STL std::stable_sort algorithm has been
hooked into the CASA Sort class. It improves QuickSort and is close
to the ParSort 1-thread case, but it cannot compete with ParSort
multi-thread case.
The benchmarks have been performed on a 16 core Xeon E5-4620 machine
with 64 GB and HyperThreading enabled running on a CentOS 7 container.

| Large MS. 11M rows, 55GB| - | - | - |
|------------------------------------| - | - | - |
|             |      Sort1  |   Sort2    |   Sort3|
|ParSort |         1100 ms  |   2100 ms  |   450 ms|
|ParSort 1 THR  |  2800 ms  |   8200 ms  |  1000 ms|
|STL stable_sort | 6800 ms  |   7800 ms  |  2100 ms|
|QuickSort     |  14000 ms  |  15000 ms  |  4500 ms}

|Medium MS. 1M rows, 3.4GB| - | - | - |
|------------------------------------| - | - | - |
|        |           Sort1  |   Sort2    |   Sort3|
|ParSort  |         130 ms  |    170 ms  |    70 ms|
|ParSort 1 THR |    400 ms  |    700 ms  |   130 ms|
|STL stable_sort|   600 ms  |    900 ms  |   200 ms|
|QuickSort      |  1100 ms  |   1300 ms  |   370 ms|

Sort1: Default sorting: ARRAY_ID, FIELD_ID, DATA_DESC_ID, MS::TIME
Sort2: ANTENNA1, ANTENNA2, TIME
Sort3: DATA_DESC_ID, ARRAY_ID

### DIFF
--- a/casa/Utilities/Sort.cc
+++ b/casa/Utilities/Sort.cc
@@ -234,30 +234,44 @@ uInt Sort::unique (Vector<uInt>& uniqueVector, uInt nrrec) const
 }
 
 uInt Sort::unique (Vector<uInt>& uniqueVector,
-		   const Vector<uInt>& indexVector) const
+                   const Vector<uInt>& indexVector) const
+{
+    Vector<uInt> changeKey;
+    return unique(uniqueVector, changeKey, indexVector);
+}
+
+uInt Sort::unique (Vector<uInt>& uniqueVector,
+                   Vector<uInt>& changeKey,
+                   const Vector<uInt>& indexVector) const
 {
     uInt nrrec = indexVector.nelements();
     uniqueVector.resize (nrrec);
+    changeKey.resize (nrrec);
     if (nrrec == 0) {
         return 0;
     }
     // Pass the sort function a C-array of indices, because indexing
     // in there is (much) faster than in a vector.
-    Bool delInx, delUniq;
+    Bool delInx, delUniq, delChange;
     const uInt* inx = indexVector.getStorage (delInx);
     uInt* uniq = uniqueVector.getStorage (delUniq);
+    uInt* change = changeKey.getStorage (delChange);
     uniq[0] = 0;
     uInt nruniq = 1;
+    uInt idxComp;
     for (uInt i=1; i<nrrec; i++) {
-        Int cmp = compare (inx[i-1], inx[i]);
-	if (cmp != 1  &&  cmp != -1) {
-	    uniq[nruniq++] = i;
-	}
+        Int cmp = compareChangeIdx (inx[i-1], inx[i], idxComp);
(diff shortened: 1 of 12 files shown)


cquike:
Looking more in detail of class MSIter I have found an inconsistency in the sorting options used for the sort() method of the ms and the TableIterator constructor.

The MSIter class was calling the sort() method of MeasurementSet and then calling the constructor of TableIterator with the NoSort option, assuming that the sorting was previously done.
However the sort parameters (columns, comparison functions and sorting order) weren't the same one, leading to wierd behaviour in cases where the comparison functions or sorting order are not the default ones. 

In particular, if timeInterval=0 and MS::TIME is part of the columns, the sort() method of the ms would get the default comparison function for double, effectively sorting in time. However this is not what the user has requested.

cquike:
Also, a rounding error has been fixed in the case timeInterval=0. In that case MSIter sets the interval to DBL_MAX, and the comparison function for time intervals (MSInterval) tries to figure out to which interval each timestamp belongs. Due to rounding errors the interval sometimes is -1 or 1, rather than 0, which is the interval index were all the timestamps should be.
Now a large number in interval, comparable to DBL_MAX is shortcut and will simply return that the two timestamps match in the comparison.

cquike:
One more thing: when reviewing turn on the  Hide whitespace changes option in the diff for better clarity. Unfortunately the original file had some end of line ^M characters that have now been removed.

tammojan:
As discussed in person at ADASS, it would be good to split up this PR into smaller ones (separate the performance improvements from the API change if possible).

Before merging those PRs I'd like to test them against python-casacore, DPPP and wsclean.

tammojan:
Apart from this, the optimizations and changes make a lot of sense to me as you explained in code comments and oral explanation.

cquike:
This PR is now obsolete, since its functionality has been taken over by #992 , #994 and #1106.
Hence closing.

PR #957, opened 2019-09-13
Use ADIOS2 Table engine for Adios2StMan

JasonRuonanWang:
The Table engine in ADIOS2 is developed specifically for casacore. It largely reduces the metadata size for cases where a large number of writer ranks write small data chunks simultaneously.


JasonRuonanWang:
@tammojan I have some of the CI tests running out of time, and the Mac compilations are having issues with boost library. It doesn't look like a problem that I have introduced in this PR. Wondering what you usually do to get around these issues. Thanks.

tammojan:
Hi @JasonRuonanWang , I've restarted the timed-out build, and am looking into the failing python test (#958). Testing on MacOS systems in the cloud is unfortunately not easy, as it's not possible to run a virtual mac inside a container.

When #958 is done, you could rebase onto the (then working) master.

JasonRuonanWang:
@tammojan Okay it looks good now.

JasonRuonanWang:
Just added back the option to select different ADIOS2 engines, so it will allow optimizations such as in #966 by @rtobar  to keep using the original BP4 engine.

PR #961, opened 2019-09-19
Travis: Disable building python on macos

### LINKED ISSUE #958: Travis Mac builds fail (again)
The current master fails, because the mac builds cannot find boost-python. Apparently, something changed in the homebrew boost-python installation. I'm looking into this.

tammojan:
It's taking a bit longer to fix #958, let's temporarily disable the python tests on Travis' macos. In that way at least the other parts of casacore get tested on macos.


tammojan:
Sorry for merging my own PR, this is a quick workaround to get a 'green' master.

PR #962, opened 2019-09-19
Prevent setting OpenMP num_threads to zero

mpokorny:
If a calculation of the number of threads previously yielded a value of zero,
replace it by a value of one.

AFAICT, the OpenMP specification states that a "num_threads" value must be a
positive integer or else the behavior is implementation defined. While the
meaning of "positive integer" appears undefined in the specification, it's usage
in that document suggests a value that is strictly greater than zero, which also
happens to agree with my intuitive sense of a minimum valid value of
"num_threads".

This bug originally appeared (for me) in Conversion.cc when using the OPARI2
tool (https://www.vi-hps.org/projects/score-p/) to instrument the casacore code,
while the bug apparently remains latent when using one of the standard compilers
without the tool.

Note that StatisticsDataset::initThreadVars() also computes a value for a number
of threads, and it does not check for zero. However, I have not changed that
code because I am uncertain that a value of zero would occur in practice, and
there is further initialization done in that method after computing the number
of threads.

### DIFF
--- a/casa/OS/Conversion.cc
+++ b/casa/OS/Conversion.cc
@@ -461,8 +461,9 @@ size_t Conversion::bitToBool (void* to, const void* from,
     const size_t bits_per_loop = 8;
     const size_t nwords = nvalues / bits_per_loop;
 #ifdef _OPENMP
-    size_t nthr = std::min((size_t)omp_get_max_threads(),
-                           nwords / (16 * 1024));
+    size_t nthr =
+        std::max((size_t)1,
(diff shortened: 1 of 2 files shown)


mpokorny:
@dmehring, assuming that you're responsible for the StatisticsDataset module, I wanted to bring your attention to my original comments in this pull request. Feel free to decide for yourself whether or not the initThreadVars() method needs any changes, as I don't know enough to make any such decision.

dmehring:
Thanks @mpokorny , AFAIK, nthreads in initThreadVars() should never be zero. So, I'll have a look at that code and make any necessary changes.

PR #963, opened 2019-09-19
Cas 12689 check for things that can't be 0 + PtrHolder -> std::unique_ptr

dmehring:


dmehring:
@tammojan Looks like CI passed a few days ago. Can this please be merged? Perhaps @schiebel needs to review?
Thanks

PR #966, opened 2019-10-03
Adding more I/O patterns to Adios2StMan

rtobar:
In Adios2StMan so far we have supported putting scalars and arrays into single cells, or into cell slides, but haven't supported single operations that span across either the whole column, or a slice of a column. Since we haven't advertised such support, the storage manager classes have resorted so
far to perform multiple cell-wise I/O operations, even when at the application level users perform a single write or read. This, in turn, has led us to have poor performance when using ADIOS2 as we have been hit by its bigger metadata-to-data ratio.

This commit adds support for column-wise operations. Whole column slices are fully supported, but partial column slices are supported only when a single slice is requested (requests spanning multiple slices will currently receive an exception).

This is currently enough to improve our use case (writing visibilities from OSKAR), and now single writes are generated at the data manager level.

### DIFF
--- a/tables/DataMan/Adios2StManColumn.cc
+++ b/tables/DataMan/Adios2StManColumn.cc
@@ -111,8 +111,30 @@ void Adios2StManColumn::arrayVToSelection(uInt rownr)
 
 void Adios2StManColumn::sliceVToSelection(uInt rownr, const Slicer &ns)
 {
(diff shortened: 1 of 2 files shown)


rtobar:
Mentioning @JasonRuonanWang so he is aware of this pull request.

PR #968, opened 2019-10-15
CAS-10543 Add workaround for SMA SYSTEM_TEMPERATURE tabledwa 

Pietromezza:
FITS-IDI files from the SMA have scalar TSYS_1/2 columns instead of
array columns as required by the FITS-IDI standard.  Interpret these
as a single value that applies to all bands like we already do for
POLAA/B in the ANTENNA table.fwe

### DIFF
--- a/msfits/MSFits/FitsIDItoMS.cc
+++ b/msfits/MSFits/FitsIDItoMS.cc
@@ -3138,16 +3138,32 @@ Bool FITSIDItoMS1::fillSysCalTable()
 
   Int nVal=nrows();
   Bool dualPol=False;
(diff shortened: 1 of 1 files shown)


gervandiepen:
What is this PR about?
Mark Kettenis commented this code was already merged in 2017. Hence git complains about merge conflicts.

tammojan:
This was indeed already merged, have a look at the [corresponding section in the current master](https://github.com/casacore/casacore/blob/master/msfits/MSFits/FitsIDItoMS.cc#L3178). It was introduced in 1aeae6cde0ed2866715133d390dea3a6c622cd42.

Issue #970, opened 2019-10-18
Cache invalidation deadlock with user locks


bennahugo:
While debugging a deadlock in dask-ms I noticed that when a running all calls to python-casacore in a single threaded executor pool the library deadlocks with no exception when putcol is called without setting a writable user lock. It always seems to hang on a ISMColumn operation. The bug likely affects only this storage manager:
```
#0  0x00007f12c8d3c6d1 in casacore::ISMColumn::getScalarColumnCellsdoubleV (this=0x7f12c006f1c0, rownrs=..., values=0x7f12c4a77670) at /scratch/bhugo/oms-subset/casacore/tables/DataMan/ISMColumn.cc:464
#1  0x00007f12c8d86ad8 in casacore::StManColumn::getScalarColumnCellsV (this=0x7f12c006f1c0, rownrs=..., dataPtr=0x7f12c4a77670) at /scratch/bhugo/oms-subset/casacore/tables/DataMan/StManColumn.cc:227
#2  0x00007f12c8c7bbf2 in casacore::ScalarColumnData<double>::getScalarColumnCells (this=0x7f12c0066960, rownrs=..., val=0x7f12c4a77670)
    at /scratch/bhugo/oms-subset/casacore/casacore/tables/Tables/ScaColData.tcc:151
#3  0x00007f12c8c99375 in casacore::ScalarColumn<double>::getColumnCells (this=0x7f12c4a77450, rownrs=..., vec=..., resize=false)
    at /scratch/bhugo/oms-subset/casacore/casacore/tables/Tables/ScalarColumn.tcc:205
#4  0x00007f12c8c990d4 in casacore::ScalarColumn<double>::getColumnRange (this=0x7f12c4a77450, rowRange=..., vec=..., resize=false)
    at /scratch/bhugo/oms-subset/casacore/casacore/tables/Tables/ScalarColumn.tcc:180
#5  0x00007f12c8c9921b in casacore::ScalarColumn<double>::getColumnRange (this=0x7f12c4a77450, rowRange=...) at /scratch/bhugo/oms-subset/casacore/casacore/tables/Tables/ScalarColumn.tcc:165
#6  0x00007f12c8cd343c in casacore::TableProxy::getValueFromTable (this=0x7f12c0001b30, colName=..., rownr=100, nrow=100, incr=1, isCell=false)
    at /scratch/bhugo/oms-subset/casacore/tables/Tables/TableProxy.cc:2245
[...]

PR #971, opened 2019-10-22
Issue 970

bennahugo:
- fixes longstanding bug in incremental storage manager slice reading
 - clean up code duplication and simplifies getcolcells ~~static~~ preprocessor
definition
Incremental storage manager test ran successfully:
```
Start 200: tIncrementalStMan
200/530 Test #200: tIncrementalStMan ................   Passed    2.15 sec
```

### DIFF
--- a/tables/DataMan/ISMColumn.cc
+++ b/tables/DataMan/ISMColumn.cc
@@ -26,18 +26,6 @@
 //# $Id$
 
 #include <casacore/tables/DataMan/ISMColumn.h>
-#include <casacore/tables/DataMan/ISMBase.h>
-#include <casacore/tables/DataMan/ISMBucket.h>
-#include <casacore/tables/Tables/RefRows.h>
-#include <casacore/casa/Arrays/Array.h>
-#include <casacore/casa/Arrays/Vector.h>
-#include <casacore/casa/Utilities/ValType.h>
-#include <casacore/casa/Utilities/Assert.h>
-#include <casacore/casa/Utilities/Copy.h>
-#include <casacore/casa/BasicMath/Math.h>
-#include <casacore/casa/OS/CanonicalConversion.h>
-#include <casacore/casa/OS/LECanonicalConversion.h>
-
 
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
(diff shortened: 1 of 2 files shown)


bennahugo:
Specifically this bit of code is missing a else case.
So the logic should flow something along the lines for each slice read values directly  with the getvalue call if the cache is invalid, otherwise read values **up** to the end of the cache
```
            while (rownr <= end) { \
                if (rownr < cache.start()  ||  rownr > cache.end()) { \
                    aips_name2(get,NM) (rownr, valptr); \
                    DebugAssert (cache.incr() == 0, AipsError); \
                } \
                const T* cacheValue = (const T*)(cache.dataPtr()); \
                uInt endrow = min (end, cache.end()); \
                while (rownr <= endrow) { \
	            *valptr++ = *cacheValue; \
                    rownr += incr; \
```
replaced with
```
                while (rownr <= end) {
                    if (rownr < cache.start()  ||  rownr > cache.end()) {
                        get<T>(rownr, valptr);
                        DebugAssert (cache.start() == 1, AipsError); // assert cache invalid
                        DebugAssert (cache.end() == 0, AipsError);
                        DebugAssert (cache.incr() == 0, AipsError);
                        rownr += incr;
                        valptr++;
                    } else { // cache not invalid, read up to the end of the cache
                        const T* cacheValue = (const T*)(cache.dataPtr());
                        uInt endrow = min (end, cache.end());
                        DebugAssert(cache.incr() == incr, AipsError);

                        while (rownr <= endrow) {
                            *valptr++ = *cacheValue;
                            rownr += incr;
                        } 
                    }
```

bennahugo:
tagging @gervandiepen here. From the blame log it looks like you were the original author. Could you please review.

sjperkins:
I'd like  to expand on @bennahugo 's analysis of the issue:

```cpp
while (rownr <= end) {
    // There are two cases handled by this if statement:
    //    (1) The cache doesn't contain the requested data
    //    (2) The cache is in an invalidated state (start=1, end=0, incr=0)

    // I think this logic handles (1) and (2) correctly
    if (rownr < cache.start()  ||  rownr > cache.end()) {
       // extract data from the storage manager
        aips_name2(get,NM) (rownr, valptr);
        // Assert cache is in invalidated state?
        DebugAssert (cache.incr() == 0, AipsError);
    }

    // This logic only handles (1) correctly.
    const T* cacheValue = (const T*)(cache.dataPtr());
    // The problem occurs here. If the cache was previously invalidated,
    // cache.end() == 0
    // which imples endRow == 0
    uInt endrow = min (end, cache.end());
    // which implies the following while loop never increments rownr
    while (rownr <= endrow) {
    *valptr++ = *cacheValue;
        rownr += incr;

    // which implies that the outer while loop starts again with the same rownr.
```

and the loop continues forever.

sjperkins:
@tammojan @gervandiepen Is there anything else required from this PR?

tammojan:
> I wonder why you want to make this change. Did you get some error somewhere?

@gervandiepen The issue is described in #970.

sjperkins:
> I wonder why you want to make this change. Did you get some error somewhere?

There is an example that reproduces this for me in this comment: https://github.com/casacore/casacore/issues/970#issuecomment-544850460

We're encountering an infinite while loop in the above code. It occurs when the ColumnCache is in an invalidated state: `(start=1, end=0, incr=0)`:

https://github.com/casacore/casacore/blob/740f336cb9f343ce5a139585d8fe4a0d691df447/tables/Tables/ColumnCache.h#L129-L132

I feel that the problem with this case is sufficiently described in https://github.com/casacore/casacore/pull/971#issuecomment-545903279. Would you mind having another look at that? If that's not sufficient, I'll try again.

tammojan:
This PR was superseded by #974, which fixes #970 in a more fundamental way.

PR #973, opened 2019-10-31
nparameters -> nparameters()

dmehring:


dmehring:
If this merge could be expedited, I'd appreciate it. It is pretty small and straight forward.
Thanks

PR #975, opened 2019-11-06
Fix measuresdata.cc with new URLs

apschoenmakers:
Certain tables that used to be hosted by https://maia.usno.navy.mil/ser7/,
which has shut down for maintenance without replacement (thank you...)

### DIFF
--- a/measures/apps/measuresdata/measuresdata.cc
+++ b/measures/apps/measuresdata/measuresdata.cc
@@ -425,7 +425,7 @@ const tableProperties allProperties[] = {
     "TAI_UTC difference obtained from USNO",		"leapSecond",
     False,			"",			vector<String>(),
     0, 				vector<formatDescr>(),
-    { "maia.usno.navy.mil", "ser7", "tai-utc.dat" } },
+    { "ftp.astron.nl", "outgoing/Measures", "tai-utc.dat" } },
 
   //**********************************************************************//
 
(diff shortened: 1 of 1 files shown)


dpetry:
Hi Arno, I can review this if you like.
Any idea what the Travis failure is about?

tammojan:
The travis error is a timeout on one of the jobs. All other jobs succeeded; I've restarted the failing one.

tammojan:
I think the fix is ok currently.

One thing noteworthy is that TAI-UTC tables are not available at maia.usno.navy.mil . We now host such a table at astron, making us responsible for updating this table when a leap second is announced. Arno and I are both subscribed to the announcements list, so this should not be a big problem.

dpetry:
Arno and Tammo, concerning the leap second table: can you give me the announcement list
 URL so I can subscribe as well? Thanks.
Also, I suppose the USNO will change all the URLs again in 6 months, right?

tammojan:
The URL is https://www.iers.org/SharedDocs/News/EN/BulletinC.html . A description of the bulletin is at https://www.iers.org/IERS/EN/Publications/Bulletins/bulletins.html . You can subscribe at http://hpiers.obspm.fr/eop-pc/products/bulletins/bulletin_registration.html .

Indeed the table may reappear at maia.usno.navy.mil in 6 months, so hosting at astron should be a temporary solution.

dpetry:
Thanks. I subscribed.

PR #988, opened 2020-01-07
Force CRLF -> LF conversion

cquike:
This PR converts the few files with CRLF (Windows) end of lines to unix ones. Additionally, it creates a .gitattributes file to ensure that even if a developer uses a Windows machine the file will be written in the repository with the proper LF (Unix) end of line.
See https://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/


cquike:
Forget about this, it included unrelated commits

PR #990, opened 2020-01-09
Remove CHANNEL_SELECTION  keyword support

cquike:
The CHANNEL_SELECTION keyword was used by CASA imager to convey some information between cycles. This is not used since time ago and it complicates unnecessarily the logic in MSIter.
The startChan() method which delivers the first "selected" channel is also removed, since there is no longer the concept of channel selection.

I believe that the MSIter class is only used by CASA, so I think that this shouldn't have an impact in other software using casacore.


cquike:
Note that I have based this PR on  PR#989, so you might want to merged that before.

tammojan:
Should this be reviewed by someone from the CASA team?

schiebel:
@gervandiepen do you know of any other groups that might be affected by this?

cquike:
I have spoken with Kumar and George about this. For the related CASA ticket (CAS-12864) I already pointed out to Kumar that he can review this PR if he wants. I can ask him again.

kgolap:
As the LF changes makes it difficult to see the relevant changes quickly...I'll be able to review this only on Monday as i start traveling today back to Socorro

tammojan:
Hi Kumar, thanks for having a look. Pro tip: you can ignore the whitespace warnings when viewing the diff through github by appending `?w=1` to the url.

cquike:
Hi,
the LF come from PR #989 which has already been merged. I thought that after being merged into master github was smart enough to recompute the diff but it seems that is not :)
I have nevertheless merged master into the branch so now the diff is clean.

In addition to pro-tip from Tammo, you can also use the tool wheel in the "Files changed" tab and select "Hide whitespace changes".

cquike:
Hi Kumar,
I agree that it would be better to remove startChan() call from casa and eventually remove that from casacore. I don't know which is the policy for deprecation within casacore. Other projects issue a deprecation warning for a given number of releases, include the proper documentation on release notes and after a relevant number of releases the functions are removed. Is this also how casacore manages API deprecations?
From what I can see in casarest there is a one call to startChan in the VisibilityIterator class, which I presume is the same code of the old VisibilityIterator that we have in CASA.

tammojan:
For casarest, this was fixed in casacore/casarest#32

PR #991, opened 2020-01-09
MSIter to propagate full metadata columns

cquike:
The array, field, data description metadata has been assumed to be constant in one given iteration. However, for applications that need to get several SPWs (or several fields) in one single iteration this is not true (those ones won't add those columns in the sorting columns).
    
This change propagates the metadata columns for array, field, dd so that the client can see the full metadata columns rather than only the first value.
    
Additionally, the logic for the new*() methods has changed. If the relevant column (or instance DataDescription) is not in the sorting columns then one cannot be sure that all the elements in one iteration have the same Id, therefore it is safer to simply signal that the DataDescription is new. In case the column is part of the sorting columns the logic remains as before.

### DIFF
--- a/ms/MeasurementSets/MSIter.cc
+++ b/ms/MeasurementSets/MSIter.cc
@@ -166,15 +166,15 @@ void MSIter::construct(const Block<Int>& sortColumns,
     cols=sortColumns;
   }
 
-  Bool timeSeen=False, arraySeen=False, ddSeen=False, fieldSeen=False;
+  timeInSort_p=False, arrayInSort_p=False, ddInSort_p=False, fieldInSort_p=False;
   Int nCol=0;
   for (uInt i=0; i<cols.nelements(); i++) {
     if (cols[i]>0 &&
 	cols[i]<MS::NUMBER_PREDEFINED_COLUMNS) {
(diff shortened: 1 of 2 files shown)


cquike:
I have asked George Moellenbrock to review this PR.

cquike:
Hi,
I have rebased this branch on top of current master. The diff should show the differences of this branch only and not showing the commits from already merged PR #990

tammojan:
The commit in this PR also ended up in #994. So if we merge #994 then this PR is not necessary anymore I guess.

cquike:
Hi,
it is correct that merging #994 would make this PR unnecessary. However, since the code review of George Moellenbrock might happen at different stages for the two PRs I still think we can keep this one open. I will ask George anyway if he prefers to do the review together, in which case we can close this PR.

gmoellen:
Enrique and I have discussed this PR, and I hereby endorse merging it.  This commit enables better support for full-column access to traditionally "required" sort columns when they are not used as sort columns.   (Many other diffs are just improved semantics.)

cquike:
The last push has rebased the branch on top of master.

tammojan:
I'm closing this PR since the commit is also part of #994

PR #992, opened 2020-01-15
Supports ParSort sorting algorithm in BaseTabIterator class

cquike:
Currently, the TableIterator class has a parameter Option in all the constructors which defaults to TableIterator::ParSort. This parameter is propagated trough several classes:    TableIterator::TableIterator() -> BaseTable::makeIterator -> BaseTabIterator::BaseTabIterator()

However the later constructor BaseTabIterator::BaseTabIterator() simply ignores the option ParSort (whereas QuickSort, HeapSort and InsSort are supported.
    
This basically adds the support for ParSort as it is intended by adding the proper "else if".

### DIFF
--- a/tables/Tables/BaseTabIter.cc
+++ b/tables/Tables/BaseTabIter.cc
@@ -43,10 +43,10 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
 
 BaseTableIterator::BaseTableIterator (BaseTable* btp,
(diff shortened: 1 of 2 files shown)


cquike:
There a couple of indentation fixed, so hiding the spaces in the diff would be better.

tammojan:
Looks good to me, well found. I'm happy to merge this when CI finishes.

PR #993, opened 2020-01-20
Allow user control of readline usage

mpokorny:
Although readline and sofa are advertised as optional, users have no way to actually control whether or not readline and sofa are used. The unguarded call to `find_package(Readline)` will create casacore libraries that link to readline libraries if they are found (and a similar situation holds with sofa). In some environments, this behavior is not desirable.

This PR allows users to state whether or not to build against and link casacore to the readline and sofa libraries through the use of `USE_READLINE` and `USE_SOFA` CMake options. I have verified that setting `USE_READLINE=OFF` results in casacore libraries that do not link to libreadline.so, even on systems where that library is present.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,7 @@ endif()
 
 # By default do not use ADIOS2, HDF5, FFTW3
 option (ENABLE_TABLELOCKING "Make locking for concurrent table access possible" YES)
+option (USE_READLINE "Build readline support" YES)
 option (USE_ADIOS2 "Build ADIOS2 " NO)
 option (USE_HDF5 "Build HDF5 " NO)
 option (USE_FFTW3 "Use FFTW instead of FFTPack" NO)
@@ -323,7 +324,9 @@ include (CTest)
 #    "Maximum ctest time allowed before CTest will kill the test." FORCE)
 
 find_package (DL)
-find_package (Readline)
+if (USE_READLINE)
+    find_package (Readline REQUIRED)
+endif (USE_READLINE)
 find_package (SOFA)
 if (USE_ADIOS2)
     set(USE_MPI ON)


tammojan:
Out of curiosity: in which environment would you not want to link to readline? Compiling for a different machine?

tammojan:
For SOFA I'm inclined to leave the situation as is. The linking is only done in a test (the test that verifies that Measures gives the same output as the reference SOFA library). Setting the default to `USE_SOFA=NO` would mean that this is not tested anymore. This could be fixed by setting `USE_SOFA=YES` in .travis.yml , but just looking for SOFA would maybe get some 'free' testing on other systems, while not hurting other use cases.

mpokorny:
I've been working on the casacore package for Spack (https://spack.io), which means that casacore may be built on a machine that has several versions of readline installed. If readline is indeed an optional dependency, the user ought not to be surprised by having expected to build casacore without readline, only to find that the library is in fact a dependency at link-time. I don't have an objection to simply asserting that readline is a required dependency, but I thought that it could remain optional with the changes in this PR. I'll admit that I don't entirely know what the readline library is used for in casacore.

As for the default for `USE_SOFA`, I don't see any issue with using a default value of `ON`, as you suggest.

mpokorny:
FYI, in the casacore Spack package, I declared readline to be a required dependency to avoid unintentional linkage with readline when the dependency was optional and the option was set to "off". So there are workarounds. SOFA is less of an issue, as it isn't nearly as widespread as readline.

mpokorny:
I see now the problem with declaring `USE_SOFA=YES` by default along with `find_library(SOFA REQUIRED)`, as the Travis CI build then fails. If I use instead `find_library(SOFA)`, still guarded by `if (USE_SOFA)` then we should get something like the current behavior, I think, on Travis and other systems. In that case `USE_SOFA=YES` would amount to saying "use SOFA if it's present, but it's OK to proceed if it isn't", while `USE_SOFA=NO` would at least avoid linking to SOFA if it existed. That ought to be OK for my use case, too. What do you think?

tammojan:
I think linking to sofa if it exists is not a problem, since only a test case will link to it. The test case should never be installed.

So I suggest to keep the logic around SOFA as it was (i.e. remove the `USE_SOFA` option).

I think readline is only used in the taql executable for interactive command line exploration of tables. Readline helps with more user friendly user input (navigating to beginning of word, keeping history of commands, etc).

mpokorny:
This last version includes just the changes to readline usage.

I agree that leaving the SOFA dependency management as it is would create no issues for an  installed casacore package if only test cases link with SOFA.

tammojan:
Thanks Martin.

PR #1005, opened 2020-03-31
CAS-11726 uvfits import VLA archive AN positions bug fix

dmehring:


dmehring:
The OSX CI failures appear to be due to configuration errors and have nothing to do with my code change.

PR #1007, opened 2020-04-06
Fixing MacOS Travis builds

rtobar:
This might the cause of cmake not being able to build python3 (i.e., the Headers/ directory is still pointing to python@2's). An alternative would be to pass down the include directory explicitly.

I will be force-pushing into this branch until things work -- so expect a couple of failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,17 +2,6 @@ language: cpp
 cache: ccache
 sudo: false
 
-addons:
-  homebrew:
-    packages:
-    - fftw
-    - hdf5
-    - ccache
-    - boost-python3
-    - cfitsio
-    - wcslib
-    update: true
-
 env:
   global:
     - CCACHE_COMPRESS=1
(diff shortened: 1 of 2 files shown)


rtobar:
One of the Mac builds (but not the other) had [one failure](https://travis-ci.org/github/casacore/casacore/jobs/671453200#L5075) in [this test assertion](https://github.com/casacore/casacore/blob/41c6d211a4c9b733042c263017116af00e32d2d7/casa/OS/test/tHostInfo.cc#L85), which is a completely unrelated issue that needs fixing (i.e., the issue are the steps the test is performing to check time granularity, not the time granularity itself necessarily). There was a second error in one of the Linux builds which seems to be again coming from an invalid check in the tObjectID test; I'll address this one on a separate PR.

In any case the purpose of this PR has now been achieved, as MacOS builds are back to a working state. Feel free to merge as-is, or to wait until the other two issues have been addressed.

tammojan:
Thanks so much for fixing the MacOS tests @rtobar!

PR #1008, opened 2020-04-06
Adjusting tests expectations

rtobar:
This is a follow-up to the problems found during #1007

In tHostInfo an assertion was being made on the amount of time slept on the system as a way to ensure HostInfo::secondsFrom1970 worked as intended. This combination of sleeping plus checking the time difference was very sensible to how computers sleep, their current workload, etc, and the assertion actually broke in a Travis test under MacOS because the sleep(5) call slept either less than 4.9 or more than 5.1 seconds. A less precise but more correct assertion is simply sleeping for less time and assert that the difference is greater or equal than 0, which should always be true.

On the other hand, in tObjectID the byte in the hash coming from the PID can actually be zero 1/256th of the times when the PID of the process running the test is divisible by 256. We are thus again removing the assertion, which was not always correct to make.

### DIFF
--- a/casa/OS/test/tHostInfo.cc
+++ b/casa/OS/test/tHostInfo.cc
@@ -79,10 +79,9 @@ int main()
     AlwaysAssertExit(memory == HostInfo::memoryTotal( )); // make sure no chang
 
     Double now = HostInfo::secondsFrom1970();
-    sleep(5);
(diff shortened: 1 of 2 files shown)


tammojan:
Thanks @rtobar!

PR #1010, opened 2020-04-08
Replace incorrect implementation of complex asin/acos with std versions

aroffringa:
I noticed that the complex versions of asin and acos in casacore return the wrong sign for the imaginary part. In casacore, these two functions are custom implementations and don't use std::asin/std::acos. The std versions do return the right values.

I wrapped my head a bit around the question whether a flipped sign of asin/acos imaginary return value is really incorrect or not, or whether 'both answers are correct' and I believe it is really wrong: for the complex trig functions, for any input phase, the values have a unique output phase, hence an imaginary sign swap is wrong. Even if that were not the case, cos(acos) and sin(asin) do not return the same value in casacore's implementation, while they do for std's version.

I checked and I don't see any usage of complex asin/acos in casacore. I added a small test set to test whether casacore::asin/acos (as well as atan) return the same values as std::asin/acos, which before the fix reports this:

```
for z=(0.7,-0.3), casacore::asin(z)=(0.708903,0.385591), std::asin(z)=(0.708903,-0.385591)
for z=(0.7,-0.3), casacore::acos(z)=(0.861893,-0.385591), std::acos(z)=(0.861893,0.385591)
terminate called after throwing an instance of 'casacore::AipsError'
  what():  (/home/anoko/projects/casacore-fork/casa/BasicSL/test/tComplex.cc : 56) Failed AlwaysAssert !fail
Aborted
```

This fix removes the incorrect custom implementation and returns the value of std::asin/acos.

### DIFF
--- a/casa/BasicSL/Complex.cc
+++ b/casa/BasicSL/Complex.cc
@@ -149,36 +149,16 @@ Complex atan(const Complex &in) {
 		 0.25*std::log((1.0+n+2*imag(in))/(1.0+n-2*imag(in))));
 }
 DComplex asin(const DComplex &in) {
-  const Double n = norm(in);
-  Double a = 0.5*std::sqrt(1.0+n+2*real(in));
-  const Double c = 0.5*std::sqrt(1.0+n-2*real(in));
-  const Double b = a-c;
-  a += c;
-  return DComplex(std::asin(b), std::log(a+std::sqrt(a*a-1.0)));
+  return std::asin(in);
 }
 Complex asin(const Complex &in) {
-  const Float n = norm(in);
-  Float a = 0.5*std::sqrt(1.0+n+2*real(in));
-  const Float c = 0.5*sqrt(1.0+n-2*real(in));
-  const Float b = a-c;
-  a += c;
-  return Complex(std::asin(b), std::log(a+std::sqrt(a*a-1.0)));
+  return std::asin(in);
 }
 DComplex acos(const DComplex &in) {
-  const Double n = norm(in);
-  Double a = 0.5*std::sqrt(1.0+n+2*real(in));
-  const Double c = 0.5*std::sqrt(1.0+n-2*real(in));
-  const Double b = a-c;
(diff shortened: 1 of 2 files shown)


gervandiepen:
I looked at the definitions on https://en.wikipedia.org/wiki/Inverse_trigonometric_functions
Calculating asin this way yields the std::asin result. However, for acos it is somewhat strange. Calculating it with the ln, it yields -std::acos, but when using pi/2-asin, it yields std::acos.
It looks as if a minus sign is wrong in the derivation of the formula. When doing it myself it results in asin-pi/2.
I believe wikipedia is wrong. http://scipp.ucsc.edu/~haber/archives/physics116A10/arc_10.pdf derives the acos formula (17) which yields the std::acos result.
To a certain extent the sign does not matter too much; in all cases the sin or cos of the result yields the original. So maybe you can say all answers are correct, but the std:: answers correspond exactly to the formal definition.

tammojan:
I'm in favor of being consistent with the `std::` implementation.

aroffringa:
Thanks Ger! I didn't look at the definition acos. Just note that I was talking about the wrong sign for the imaginary value only. Making an error in the imaginary value is wrong, and does result in that cos(acos(x)) is no longer x. Whether the "full" sign of the acos result is also wrong, is indeed less of an issue, since returning -acos would not result in a correct but not standard value. But for that I agree with tammo-jan, probably better to use std::acos' definition. Maybe Ger can fix Wikipedia? :D

gervandiepen:
I think the test of fabs(complex) should not be removed.
fabs is defined in Complex.h, so should work fine for all compilers.

You're right about the imaginary sign.

aroffringa:
I didn't remove already existing tests. I had added a test for `std::fabs`, but `std::fabs` is not a standard function for complex values, I think it is a gnu extension. Hence it failed to compile on the Mac. The standard C++ function name is std::abs, so casacore::fabs(Complex) has no equally named counterpart.

PR #1012, opened 2020-04-09
Refactor of Arrays

aroffringa:
This is a major refactoring of the Arrays subdirectory making it more in line with C++11. The test [tCpp11Features.cc](https://github.com/aroffringa/arrays2/blob/master/test/tCpp11Features.cc) is a good demonstration of some of the new features. Summary of changes / goals:

- The Array directory is now completely independent of all other casa headers: it can be compiled and tested by itself, and doesn't include any other casacore header.
- An Array now supports moving its elements. Therefore, it can now contain uncopyable classes like `std::unique_ptr`
- Move constructors and move assignments were added to the most relevant types
- Removed the use of CountedPtr, used unique_ptr instead.
- I've added iterator constructors (i.e. Array(begin, end)), and removed constructors taken a 'Block'.
- I've for now removed the option to skip initialization. I doubt the way it is done was standard C++ compliant due to rules of type aliasing -- and it is in any case not always allowed. This also didn't seem used anywhere in casacore.
- I've removed the allocators from Array. They were not used anywhere in casacore, and at least to me highly confusing. Instead, I've added the possibility for C++STL-like allocators, i.e. you can define something like an `Array<int, PoolAllocator>`. It wasn't hard to add, but not sure if ever useful.
- I've added support for initializer lists, like Vector<int> a{1,2,3} and IPosition ip{2,5}.
- Added fast swap operators for some of the classes like IPosition and Array.
- Added `to_string()` implementations for Array, IPosition.
- I rewrote the Array tests using Boost unit testing platform. This makes the tests more debuggable.
- Array-exceptions are now based on std::runtime_error instead of AipsError.
- Casacore's "Uppercase-types" like Int, Char, Bool etc. were replaced by their lower-case C++ counterparts. Array indices now use `size_t` instead of uInt. Use of String was replaced by std::string, use of Regex was replaced by std::regex.
- There were many other small things like sorting and medium calculation that had manual implementation, which I replaced by calls to equivalent std calls.
- Add `final` and `override` to methods/classes where appropriate

One of the biggest issues I have with arrays is its confusing copy & assignment semantics. Unfortunately this is not fixable without a major overhaul of the rest of casacore (and casa) so I didn't change any semantics so far. Might be a consideration for the future.

I was using my own Git repository here https://github.com/aroffringa/arrays2/ , which can be compiled and tested without casacore.

Due to the fact that many of the Array headers originally included other parts of casa, like BasicMath, BasicSL and Utils, many files outside Arrays needed extra includes, because they were depending on Array headers to also include those files. Hence, this has become a really major pull request touching lots of files, and will like also require equal amount of changes in casa.

Other required "global" changes were to catch std::exception instead of AipsError. Also the use of size_t for nr of elements instead of uInt needed a few changes, for example because in some cases there was no exact overload match anymore. There are more incompatibilities, but they were mostly minor.

I've put the few methods for AipsIO and LogIO for arrays in casa/IO/ArraysIO.h . A new file was added, `casa/Arrays/ArrayStr.h` that has the string handling routines.


aroffringa:
Still to do:
- Read possibility to make an uninitialized storage
- Think about "compatibility headers"
- Compile DP3 against this branch
- Describe implications for Casa

aroffringa:
I've added a text file that summarizes the changes that are going to be necessary in Casa:

https://github.com/aroffringa/casacore/blob/arrays2/casa/Arrays/Changes-May-2020.txt

schiebel:
This looks good, thanks for your patience and help @aroffringa 
@tammojan is now a good time to merge this pull request?

schiebel:
I'll just assume this is a good time... thanks @tammojan and @aroffringa

tammojan:
Thanks for merging @schiebel!

Issue #1019, opened 2020-04-22
MSFitsIDI antenna swapping is broken


kettenis:
When importing a FITS-IDI file that has baseline antenna pairs in non-canonical order
MSFitsIDI does not correctly swap cross-polarization products.  This is very similar to issue #977 .

Seen with full-polarization e-MERLIN data as reported by Anita Richards in CASA ticket CAS-12989.  Fix in progress,

gijzelaerr:
This issue has been resolved in #1020

PR #1020, opened 2020-04-22
Fix issue #1019

kettenis:
Swap the data, weight and flags corresponding to cross-polarization
products if necessary when swapping antennas for a baseline.


dpetry:
Hi, please describe the tests you have done.

kettenis:
Hi Dirk,

So far I've verified tat:
1. This fixes the issue with the e-MERLIN data that Anita made available to me. The polarization labels now match the data that has been exported from AIPS.
2. This does not have an effect on a full-polarization EVN data set.
3. Still passes the CASA6 testsuites for both casatools and casatasks  This includes the test_importfitsidi.py test.

Cheers,
Mark

PR #1022, opened 2020-04-24
Port Adios2StMan to use 64bit row numbers

rtobar:
This patch ports the Adios2StManColumn class hierarchy to inherit
directly from StManBaseColumn (the new, 64bit-only interface) instead of
StManColumn (the old, 32bit-backwards-compatible one) and adjusts the
implementation to comply with the new methods taking rownr_t instead of
uInt. The Adios2StMan class itself (and its pimpl) also needed to be
adjusted, although the changes were minimal.

A major change that happened together with this porting was a small
refactoring of the internal Adios2StManColumn class hierarchy. The
Adios2StManColumn has always been the base class used by the many
AdiosStManColumnT<T> class template specializations, but the latter
contained a fair bit of code that could be moved down to the base class
with minimal changes. In this commit I thus adjusted and moved all the
common code down to Adios2StManColumn, making the derived template class
much leaner. A consequence of this internal refactoring is that now the
std::string specialization of this template class is not used directly
anymore, by instead it is inherited by Adios2StManColumnString, which is
what Adios2StMan instantiates. This was done to ensure that string
columns properly fail to implement certain operations that are not
supported by the storage manager yet.

In doing this porting I also caught some of the I/O patterns not being
fully supported (e.g., we had getArrayColumnV but no putArrayColumnV) or
not properly implemented (e.g., getArrayColumnV did not set the ADIOS2
selection objects correctly). This obviously means that our unit test
coverage is not great, and that our other external use cases are not
touching those remaining parts of the code. This is a limitation that we
should try to improve in future releases.

As per expectations outlined in the DataManagerColumn documentation, the
implementation of Adios2StManColumn is now simpler thanks to it deriving
from StManBaseColumn.

### DIFF
--- a/tables/DataMan/Adios2StMan.cc
+++ b/tables/DataMan/Adios2StMan.cc
@@ -74,19 +74,19 @@ String Adios2StMan::dataManagerName() const
 	return pimpl->dataManagerName();
 }
 
-void Adios2StMan::create(uInt aNrRows)
+void Adios2StMan::create64(rownr_t aNrRows)
 {
-	pimpl->create(aNrRows);
+	pimpl->create64(aNrRows);
 }
 
-void Adios2StMan::open(uInt aRowNr, AipsIO &ios)
+rownr_t Adios2StMan::open64(rownr_t aRowNr, AipsIO &ios)
 {
-	pimpl->open(aRowNr, ios);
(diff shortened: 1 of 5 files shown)


tammojan:
Thanks @rtobar!

PR #1023, opened 2020-05-01
must check getIndex return (CAS-12818)

schiebel:
I believe this resolves the memory violation with uvfits. It was unrelated to either the 64bit row change or the lldb inspired potential fix.


tammojan:
Thanks @schiebel!

PR #1026, opened 2020-05-05
Improve markdown in CTDS document [ci-skip]

tammojan:


gervandiepen:
The changed look fine to me.

PR #1027, opened 2020-05-06
Improved printing of MVTime

gervandiepen:
The ISO format of date/time printing in MVTime was using a T instead of space, so the USE_SPACE option was added. Furthermore, it is better to use the CLEAN as well in case seconds or minutes are not printed.
For TaQL date/time formatting in the str function the comma was added as optional separator next to the vertical bar.


tammojan:
`tTableGramFunc` fails on all builds.

tammojan:
Thanks for fixing the test @gervandiepen . One minor comment: please adhere to git commit message guidelines, so `Fix print difference` rather than `Fixed print difference`. See https://chris.beams.io/posts/git-commit/

PR #1024, opened 2020-05-03
Fix Fortran errors in gfortran-10

aroffringa:
This PR is supposed to fix #1017 . gfortran 10 has changed some behaviour, and some of the things that were previously warnings are now errors and thus cause compilation to fail.

Issue #1017, opened 2020-04-20
Compilation problem with gcc-10


olebole:
On Debian, we are currently testing to build our packages with the new gcc release 10. For casacore, this leads to an errror as reported in [Debian#957065](https://bugs.debian.org/957065):

```
[ 50%] Building Fortran object scimath_f/CMakeFiles/casa_scimath_f.dir/dfftpak.f.o
cd /<<PKGBUILDDIR>>/obj-x86_64-linux-gnu/scimath_f && /usr/bin/gfortran -DCFITSIO_VERSION_MAJOR=3 -DCFITSIO_VERSION_MINOR=470 -DHAVE_FFTW3 -DHAVE_FFTW3_THREADS -DHAVE_O_DIRECT -DHAVE_READLINE -DUSE_THREADS -DWCSLIB_VERSION_MAJOR=7 -DWCSLIB_VERSION_MINOR=1 -Dcasa_scimath_f_EXPORTS -I/<<PKGBUILDDIR>> -I/<<PKGBUILDDIR>>/obj-x86_64-linux-gnu -I/usr/include  -g -O2 -fdebug-prefix-map=/<<PKGBUILDDIR>>=. -fstack-protector-strong -fPIC   -c /<<PKGBUILDDIR>>/scimath_f/dfftpak.f -o CMakeFiles/casa_scimath_f.dir/dfftpak.f.o
/<<PKGBUILDDIR>>/scimath_f/dfftpak.f:732:72:

  732 |       CALL DRFTB1 (N,R,WSAVE,WSAVE(N+1),WSAVE(2*N+1))
      |                                                                        1
Error: Type mismatch in argument ‘ifac’ at (1); passed REAL(8) to INTEGER(4)
/<<PKGBUILDDIR>>/scimath_f/dfftpak.f:214:72:

  214 |       CALL DRFTF1 (N,R,WSAVE,WSAVE(N+1),WSAVE(2*N+1))
      |                                                                        1
Error: Type mismatch in argument ‘ifac’ at (1); passed REAL(8) to INTEGER(4)
/<<PKGBUILDDIR>>/scimath_f/dfftpak.f:82:72:

   82 |       CALL DRFTI1 (N,WSAVE(N+1),WSAVE(2*N+1))
      |                                                                        1
Error: Type mismatch in argument ‘ifac’ at (1); passed REAL(8) to INTEGER(4)
/<<PKGBUILDDIR>>/scimath_f/dfftpak.f:1284:72:

[...]

PR #1029, opened 2020-05-07
Use FFTW instead of FFTPack in Fits-IDI to MS table converter

aroffringa:
The FitsIDItoMS class is the only class that was still dependent on the FFTPack
Fortran code. By rewriting it to use FFTW, the FFTPack code can be removed.
This required implementing a wrapper for the real even FFT in the FFTW class,
as that was not available.

The FFTW class should be rewritten to have a more sensible interface. I've added
some TODO comments for that. Despite the use of a mutex, the FFTW wrapper is
also not thread safe.

This is related to the work in #1024 and #1017.

### DIFF
--- a/msfits/MSFits/FitsIDItoMS.cc
+++ b/msfits/MSFits/FitsIDItoMS.cc
@@ -2216,7 +2216,7 @@ void FITSIDItoMS1::fillMSMainTable(const String& MSFileName, Int& nField, Int& n
 	      }
 	      fftOut[nChan] = 0.0;
 
-	      // Cosine trandorm back to frequency domain
+	      // Cosine transform back to frequency domain
 	      redftPlan.Execute(fftOut.data(), fftIn.data());
 
 	      for (Int chan=0; chan<nChan; chan++)
--- a/msfits/MSFits/FitsIDItoMS.cc
+++ b/msfits/MSFits/FitsIDItoMS.cc
@@ -104,7 +104,7 @@
 
 #include <casacore/casa/iomanip.h>
 
-#include <casacore/scimath/Mathematics/FFTPack.h>
+#include <casacore/scimath/Mathematics/FFTW.h>
 
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
@@ -1850,11 +1850,8 @@ void FITSIDItoMS1::fillMSMainTable(const String& MSFileName, Int& nField, Int& n
   Matrix<Float> sigmaSpec(nCorr, nChan);
   Matrix<Float> weightSpec(nCorr, nChan);
 
-  Vector<Float> tmp(nChan + 1);
-  Vector<Float> work(3 * (nChan + 1) + 15);
-  Bool worksave;
-  Float *workptr = work.getStorage(worksave);
-  FFTPack::costi(nChan + 1, workptr);
+  std::vector<float> fftIn(nChan + 1), fftOut(nChan + 1);
+  FFTW::Plan redftPlan = FFTW::plan_redft00( IPosition(1, nChan+1), fftIn.data(), fftOut.data() );
 
   const Int nCat = 3; // three initial categories
   // define the categories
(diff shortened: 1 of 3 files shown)


dpetry:
Hi, what tests did you run to verify your changes?

aroffringa:
Hi @dpetry . I've discussed with Mark Kettenis (who if I understood correctly wrote the fits-to-ms) class, that he will test the changes. That still needs to be done, that's why the PR is a draft.

dpetry:
The FITS2MS class was written by a number of people including Mark and myself.
Mark was the last person working on it.
Let's wait for the outcome of his tests then.

kgolap:
Just an FYI...FFTPack is very safe to use in a multi-thread environment especially where each thread are dealing with their own FFT's ...I have never been able to get FFTW to work safely that way.

aroffringa:
@kgolap  FFTW is not so hard to make threadsafe; WSClean uses it in a multi-threaded way.

kettenis:
@aroffringa When I drop this change in my CASA6 tree it doesn't build:

In file included from /opt/rh/devtoolset-4/root/usr/include/c++/5.3.1/memory:81:0,
                 from casacore/casacore/scimath/Mathematics/FFTW.h:35,
                 from casacore/scimath/Mathematics/FFTW.cc:28:
/opt/rh/devtoolset-4/root/usr/include/c++/5.3.1/bits/unique_ptr.h: In instantiation of ‘void std::default_delete<_Tp>::operator()(_Tp*) const [with _Tp = casa6core::FFTWPlanf]’:
/opt/rh/devtoolset-4/root/usr/include/c++/5.3.1/bits/unique_ptr.h:236:17:   required from ‘std::unique_ptr<_Tp, _Dp>::~unique_ptr() [with _Tp = casa6core::FFTWPlanf; _Dp = std::default_delete<casa6core::FFTWPlanf>]’
casacore/scimath/Mathematics/FFTW.cc:286:14:   required from here
/opt/rh/devtoolset-4/root/usr/include/c++/5.3.1/bits/unique_ptr.h:74:22: error: invalid application of ‘sizeof’ to incomplete type ‘casa6core::FFTWPlanf’
  static_assert(sizeof(_Tp)>0,
                      ^
/opt/rh/devtoolset-4/root/usr/include/c++/5.3.1/bits/unique_ptr.h: In instantiation of ‘void std::default_delete<_Tp>::operator()(_Tp*) const [with _Tp = casa6core::FFTWPlan]’:
/opt/rh/devtoolset-4/root/usr/include/c++/5.3.1/bits/unique_ptr.h:236:17:   required from ‘std::unique_ptr<_Tp, _Dp>::~unique_ptr() [with _Tp = casa6core::FFTWPlan; _Dp = std::default_delete<casa6core::FFTWPlan>]’
casacore/scimath/Mathematics/FFTW.cc:286:14:   required from here
/opt/rh/devtoolset-4/root/usr/include/c++/5.3.1/bits/unique_ptr.h:74:22: error: invalid application of ‘sizeof’ to incomplete type ‘casa6core::FFTWPlan’

I think you went a step too far in modernizing the C++...

aroffringa:
Hi @kettenis , thanks for testing so quickly. You get this error when you are compiling without FFTW. Can you enable it / make sure it is found by cmake?

( I've added code to make it at least compile without FFTW for now, but when doing that and it is being used in the code, you'll get a runtime exception )

kettenis:
Oh, that was interesting. I dropped the changes into my CASA6 build which appears to include FFTW but actually doesn't because of a typo in the setup script.
 Builds now so I'll get the testing going.

kettenis:
Initial testing suggests this is working correctly. Differences in amplitudes of the autocorrelations between FFTW and FFTPack are of the order of 1e-6, with zero mean.  Since the autocorrelations are (approximately) normalized to unity that is a different one would expect for single-precision floating point.
I'd like to do a bit more testing and we will need to sort out the FFTW situation on the CASA side before this can be merged.

aroffringa:
Thanks @kettenis . What implications for CASA do you mean? This PR isn't changing the interface of FFTW and is also not yet removing FFTPack, so for CASA there's no change. For PR #1024 , the FFTPack Fortran code is planned to be removed. So wouldn't this PR be safe to merge now, and check the CASA situation before merging #1024 ?

(it's good that you check the CASA situation though)

kettenis:
Hi @aroffringa, the FitsIDItoMS class is used by the CASA importfitsidi taskso the lack of FFTW support would break that task. The NRAO folks are aware though and the lack of FFTW support in CASA6 was unintentded, so I think we're all good.

kettenis:
Done some more extensive testing. Now tested VLBA data (DiFX), GMVA data (DiFX) and EVN data (SFXC). Everything seems to behave as expected. The autocrrelation amplitudes show small variations for VLBA and GMVA data (< 1e-6) and in some cases there is a clear discretization effect that you often see in numerical noise. No changes in the cross-correlations at all which is good since the Cosine transform is only used in the code that touches the autocorrelations. And phase are identical as well. The EVN data is completely unaffected since the code thatmakes these corrections is bypassed completely.

aroffringa:
Ok, great! Ah I hadn't realized that FFTW was not available in Casa, I thought your problem was 'local'. Thanks again for the tests. Is the PR now ready for review?

kettenis:
@aroffringa Not sure if you're asking me, but I suppose so.

aroffringa:
Thanks @kettenis . @dpetry do you want to review the changes?

aroffringa:
@dpetry You put yourself as reviewer on this PR. I assume this can be merged for you?

dpetry:
Sorry for delaying this. I was assured by Mark that he is verifying the changes.
If Mark has verified that the changes have not changed the functionality, they are fine with me.

dpetry:
@kettenis , you say that your tests actually have not exercised the modified code, correct?
How were the code changes tested at all?

dpetry:
@aroffringa , I reread Mark's comments and see now that his tests did exercise your code sufficiently. I will merge the request now.

PR #1035, opened 2020-06-18
CAS-11514 fix bug that required SOUrCE::REST_FREQUENCY column

dmehring:
which is optional according to the MS spec


dmehring:
@schiebel @tammojan Pinging re: merge. I'm working on another uvfits issue, so it would be useful to have this code on master. The change is pretty minimal and straight forward.
Thanks

schiebel:
@dmehring on the [MS spec](https://casa.nrao.edu/Memos/229.html#SECTION000615000000000000000) it looks like REST_FREQUENCY is required for spectral line information, but optional for doppler tracking and source information. How is this change related to the required vs optional uses of REST_FREQUENCY (I only see one rest frequency #define, and I'm not sure exactly sure of how the required/optional pieces are checked).

dmehring:
Hi @schiebel . My reading is that REST_FREQUENCY is an optional column in the SOURCE table. The use case [here](https://open-jira.nrao.edu/browse/CAS-11514) is that a user wants to export an apparently valid MS, that has no REST_FREQUENCY column in the SOURCE table, to uvfits.

PR #1036, opened 2020-07-16
Update the support for 64bits containers sizes.

cquike:
The containers Vector and Block used size_t for nelements() since more than 5 years ago. 
MSIter now properly supports size_t as the type for container sizes.


aroffringa:
The Arrays part (Vector) is already covered partially by  #1012 .

tammojan:
Thanks!

Issue #1039, opened 2020-08-04
Import FITS-IDI GAIN_CURVE tables


kettenis:
For amplitude ca;ibration of VLBI arrays it is important to know the sensitivity of each telescope
and have a description of how the antenna gain changes as a function of antenna pointing.
This information is provided by the GAIN_CURVE table in FITS-IDI files.  Having a straightforward way
to import this data into a Measurement Set is essential for providing a good user experience.

Below you found a proposal for an (ad-hoc) table that can be included in a Measurement Set to carry
this data.  I plan to use this table in CASA to improve the implementation of amplitude calibration for VLBI.
But the table format should be generally usable for other software packages and non-VLBI dish arrays.

The proposal includes various gain curve types/parametrizations that are currently in use for VLBI antennas and the VLA, but
leaves out some gain curve types that have been used in the past for antennas of a more esoteric design.
Other gain curve types/parametrizations can be added later if needed.

## GAIN_CURVE: Gain Curves

Name | Format | Units | Measure | Comments
-------|--------|-------|-------|--------
ANTENNA_ID | Int | | | Antenna id
FEED_ID	| Int	| | |	 	Feed id
SPECTRAL_WINDOW_ID |	Int	| | | 	 	Spectral window id
TIME_RANGE	| Double(2) |	s |	EPOCH |	Start, end time
TYPE |	String | | |	 	 	Gain curve type
NUM_POLY |	Int | | |	 	 	Series order
GAIN |	Float(N_r, NUM_POLY) | | |	 	 	Gain
SENSITIVITY |	Float(N_r) | 	K/Jy | |	 	Sensitivity

### ANTENNA_ID

Antenna number, as indexed from ANTENNAn in MAIN.

### FEED_ID

Feed identifier, as indexed from FEEDn in MAIN.

References the feed this gain curve is associated with. Can be -1 if
association between gain curve and feed is not known.

Rationale: Gain curves are in principle feed dependent. For example
dual S/X-band observations will almost certainly need different gain
curves for S-band and X-band since in most cases different receivers
are used. In practice this will often be set to -1 as the
association is often lost when data is converted.

### SPECTRAL_WINDOW_ID

References the spectral window this gain curve is associated with.
Can be -1 if association between gain curve and spectral window is
not known.

Rationale: While gain curves are in principle feed dependent, the
association between a particular feed and its gain curve is often
lost when data is converted. However, the association between gain
curve and spectral window typically survives.

### TIME_RANGE

[...]

PR #1040, opened 2020-08-04
Add support for importing FITS-IDI GAIN_CURVE tables

kettenis:
See issue #1039

### DIFF
--- a/msfits/MSFits/FitsIDItoMS.cc
+++ b/msfits/MSFits/FitsIDItoMS.cc
@@ -58,6 +58,9 @@
 #include <casacore/ms/MeasurementSets/MSSpWindowColumns.h>
 #include <casacore/ms/MeasurementSets/MSTileLayout.h>
 
+#include <casacore/measures/TableMeasures/TableMeasValueDesc.h>
+#include <casacore/measures/TableMeasures/TableMeasDesc.h>
+#include <casacore/measures/TableMeasures/TableQuantumDesc.h>
 #include <casacore/measures/Measures/MDirection.h>
 #include <casacore/measures/Measures/MDoppler.h>
 #include <casacore/measures/Measures/MEpoch.h>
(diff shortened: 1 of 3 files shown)


dpetry:
Did you look at the implications of adding this table to the MS for CASA tasks like mstransform and concat?

Also, comparing with the MS FEED table (which seems to be using the same key structure), you are using a TIME_RANGE while the FEED table uses TIME and INTERVAL. Maybe you should align you design with that of MS FEED and also mark the fields which form the key like in the MS Definition document.

If the structure of the key would be the same as for the FEED table, it would have the additional advantage that the reindexing code etc. in mstransform and concat could be re-used. 

(I'm going to add this comment also to the Issue.)

dpetry:
OK, I'm going to merge this now.
But NOTE that the CASA importfitsidi task uses ms.concatenate() if the input is a list of FITSIDI files!
So the next step will have to be to make sure that concat handles the new GAIN_CURVE tables properly.

dpetry:
@kettenis and @aroffringa , there are reports of compilation problems related to this PR from CASA developers.
@kettenis, could you try and compile the latest CASA Master with a casacore version which contains this PR.

kettenis:
I can confirm that CASA doesn't build against the current casacore master.  This is obviously related to #1029 since the problem is with code in CASA that still uses FFTPACK through the FFTPack class. I guess the problem is that the change I tested for @aroffringa didn't actually remove the FFTPack code from casacore yet, whereas the final PR did.

I think the easiest way out of this mess is to reinstate the FFTPack class and the FFTPACK Fortran code in casacore (but keep the changes to make FFTServer and the FITS-IDI conversion code to use FFTW unconditionally). Compilation of the FFTPACK code should probably be conditional such that other casacore users that wish to use modern compilers can do so.

Long-term, the code in CASA should probably be converted to use FFTW by the corresponding code owners as FFTPACK is no longer maintained and can't be compiled by GCC 10.

tammojan:
Since André is on vacation, I can pick this up. I hope Wednesday is early enough. Mark, can you point me to instructions on compiling casa?

cquike:
Hi,
thank you for looking at this! Indeed, the problem is in #1029 rather than here. I just discovered when using the tip of master. The proposal from @kettenis sounds quite good: make FFTPack optional so that different applications can decide. Eventually CASA might also move away from FFTPack, but that will obviously require quite some time and we are not there yet.

cquike:
Curiously enough the commit that removed FFTPack (https://github.com/casacore/casacore/commit/33c8d5f8468860239800ab1592cd4636a80108e7) is not part of #1029 ... Was yet another PR that was dealing with this?

kettenis:
Ah, that explains why I didn't see this when I initially tested the changes proposed by @aroffringa. That commit is part of PR #1024 which references issue #1017. I think that is exactly the commit that needs to be reverted and then made conditional using some cmake magic.

@tammojan it is probably easiest for me to provide you with an account on our casa development machine at JIVE if you want to take this on (my ability to do cmake magic is fairly limited).

PR #1041, opened 2020-08-10
CAS 11860 support export/import of SYSPOWER table to/from uvfits

dmehring:


kettenis (commented):
I'm not really an expert on VLA amplitude calibration but my understanding is that this should indeed work in a very similar way to writeTY.

The way the code explicitly opens the SYSPOWER table as a file is a bit odd. Wouldn't it make more sense to open it as a table and pass it as such to _writeSY?

Probably should double-check that the science case doesn't require support for combineSpw = true. From a VLBI standpoint I would expect that to be what users would actually use, but this may very well be different for the VLA. And since SYSPOWER is VLA-specific, it is fine to leave that unimplemented if there is no science requirement.

dmehring:
Thanks @kettenis I've changed so syspower is passed as a table, not a file. I don't have a good example I can use for testing for multiple IFs, so would prefer to leave that unimplemented for now and only implement if and when there is a clear use case to do so.

dmehring:
PS because this is an ad hoc table, it is not part of the MS interface, so cannot be directly accessed as a table like MSSysCal can; ie there is no MSSysPower table class because the SYSPOWER table is not part of the MS spec.

kettenis:
That is probably better. Note that the code in CASA that uses the SYSPOWER (code/synthesis/MeasurementComponents/EVLASwPow.cc) actually traverses the keywordlist to determine the name of the table:

`sysPowTabName_ = ms.rwKeywordSet().asTable("SYSPOWER").tableName();`

and then checks if the table is readable:

```
if (!Table::isReadable(sysPowTabName_))
    throw(AipsError("The SYSPOWER subtable is not present in the specified MS."));
```

That probably does a more thorough check than checking whether the directory exists...

I guess if multiple IF support is required it will show up during validation on the CASA side and you can always add it later. Proof of the pudding really is whether AIPS can read the resulting SY table.

dmehring:
Thanks @kettenis , I guess I'm missing the improvement of the first code statement you list, since the string "SYSPOWER" still has to be hard-coded, but I've made the changes you've suggested.

dmehring:
Hello? Anyone? Can someone please merge this for me?
Thanks

kettenis:
Looks good to me. But I don't have the powers to merge this...

dmehring:
Thanks @kettenis and @schiebel

PR #1043, opened 2020-08-13
Fix matching empty strings

### LINKED ISSUE #1042: Regex no longer matches empty strings
PR #881 introduced a change of behaviour in matching empty strings. In particular, the regular expression ".*" no longer matches empty strings. As a result using "*" to select all spectral windows when using the MSSelection class will fai in most cases. This causes several CASA regression tests to fail.

kettenis:
Fixes #1042

### DIFF
--- a/casa/Utilities/Regex.cc
+++ b/casa/Utilities/Regex.cc
@@ -73,7 +73,7 @@ String::size_type Regex::match(const Char* s,
   if (ps < 0) {
     ps += len;
   }
-  if (ps >= static_cast<Int>(len)) return String::npos;
+  if (ps > static_cast<Int>(len)) return String::npos;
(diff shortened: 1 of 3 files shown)


schiebel:
Thanks @kettenis I think this will fix the last CASA test failure for the casacore Array refactor that  @aroffringa has done.

It would probably be good to add this case to one of the tRegex*.cc test cases.

kettenis:
Good idea @schiebel. Seems tRegex_1.cc is the most convenient one to add a couple of tests.

FYI, this isn't fallout from the Array refactor, but from the switch to std::regex that @gervandiepen did.

schiebel:
Thanks for the test @kettenis, the array refactor branch has merged master since the regular expression change.

PR #1047, opened 2020-08-18
Add fftpack deprecated

tammojan:
As mentioned in https://github.com/casacore/casacore/pull/1040#issuecomment-671429341 , removing FFTPack in 33c8d5f caused the CASA build to break. This PR follows the proposal of @kettenis to add back FFTPack, but with a deprecation warning. It will not be built by default, except in the CASA build setting. Also, it will print a deprecation warning when including `FFTPack.h`.

Building FFTPack can be enabled at cmake time by setting the option `-DBUILD_FFTPACK_DEPRECATED=True`. The default is `False`. When `CASA_BUILD=True`, it will currently be `True` so CASA should build.


kettenis:
Does what it says on the tin.  I have successfully built an (almost) current CASA master against this.

schiebel:
Great, thanks very much @tammojan and @kettenis

PR #1051, opened 2020-08-24
Restore minimaxNC behaviour

### LINKED ISSUE #1045: Change in behaviour of MUString::minimaxNC()
Commit d2e04b9e made by @gervandiepen changed the behaviour of MUString::minimaxNC(). As a result partial matching of MDirection reference strings no longer works if the provided string is longer than the string from the list.  In particular MDirection::setRefString("J2000 ") (note the extra space) no longer matches J2000 and now returns false.

As a result of this changes some tests in the CASA testsuite now fail (@schiebel the test_simobserve and probably also the test_req_task_simanalyze)

The commit message does not suggest that this is a deliberate change.

kettenis:
Restores behaviour of this method that was changed in
commit d2e04b9ed56c59c5daf3a030d2c29ddca1e79bb2.

Fixes #1045

### DIFF
--- a/casa/Utilities/MUString.cc
+++ b/casa/Utilities/MUString.cc
@@ -475,20 +475,22 @@ uInt MUString::minimaxNC(const String &in, Int N_name,
     if (i >= N_name) {
 	size_t ia = a.length();
 	for (i=0; i<N_name; i++) {
-          if (ia < tname[i].length()) {
 	    String b = upcase(tname[i]);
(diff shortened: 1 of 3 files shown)


tammojan:
I added a test, that passes in v3.0.0, and v3.3.0 but failed in the master. With your fix it passes again, so thanks 👍

kettenis:
Thanks @tammojan, that test looks good to me.

PR #1052, opened 2020-08-25
Solve issue when assigning from lower to higher dim array

aroffringa:
In psuedo code, this didn't work anymore, causing Casa errors:

```
Cube cube(5, 1, 1);
cube = Vector(5);
```

It did "by accident" work for a Matrix, because Matrix didn't implement a specialization of assignment, causing
an implicit call conversion via the Matrix(const Array&) constructor which did allow this.

This "feature" allowing assignment from lower dim arrays isn't yet documented and is also not consistent. For example,
this isn't allowed:

```
Array arr(IPosition{5, 1, 1});
arr = Vector(5);
```

So in the nearby future it would be good if i) this were documented; and ii) this would be made consistent
between Array and Matrix/Cube classes. By doing so, the specialized assignment methods from Matrix and Cube could
be removed, so it would also simplify the code.

This PR also adds unit tests for this.

### DIFF
--- a/casa/Arrays/Cube.h
+++ b/casa/Arrays/Cube.h
@@ -122,17 +122,49 @@ public:
     // Note that the assign function can be used to assign a
     // non-conforming cube.
     // <group>
-     Cube<T, Alloc> &operator=(const Cube<T, Alloc>& source)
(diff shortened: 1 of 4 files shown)


schiebel:
Travis CI build failure was due to exceeding max runtime.

PR #1053, opened 2020-08-27
Fix empty dim-changing assignment like matrix = Vector()

aroffringa:
Code like this used to work, but didn't after the array refactor:
  Matrix<int> matrix;
  matrix = Vector<int>();
The same holds for Vector to Cube and Matrix to Cube assignment.

This has been fixed, and test cases have been added for these assignments.


tammojan (approved):

aroffringa:
I don't think the one failing test on one platform is due to my changes... can the CI be restarted?

schiebel:
Thanks @aroffringa and @tammojan

PR #1054, opened 2020-09-09
Remove boost dependency in DataMan tests

rtobar:
This commit removes the dependency that DataMan tests have on Boost's
unit testing framework. This was added in cd54d948f, where boost is
effectively used for unit testing some of the new array changes, but
it's definitely not needed by the DataMan tests, as none of them
includes boost or links against it.


aroffringa:
Thanks @rtobar . The dependency on Boost in DataMan was my fault, I hadn't meant to commit that during the Arrays refactor. Thanks for the fix.

PR #1055, opened 2020-09-15
Correct dependency management

rtobar:
This set of changes introduces a few fixes to how dependencies are managed by casacore at configuration time, specially when building a subset of the casacore modules.

Firstly, `scimath` introduces a hard dependency on BLAS, LAPACK, FFTW and the Fortran language (thus a compiler). When `scimath` is *not* built though (e.g., when building up to `tables`) these dependencies were still required at configuration time, posing an unnecessary complication on users.

Similarly, bison and flex are required only by the `tables` module and later. Users compiling only the core `casa` module would still need to have both tools installed unnecessarily.

Finally, since FFTW3 has turned into a hard dependency I've updated the Travis scripts to stop specifying the now unused `USE_FFTW` cmake variable on the `cmake` command line, and the README file to reflect the change of the dependency from optional to hard.


tammojan (approved):
Thanks! Currently we do not test whether building individual modules works. I don't think we officially support this: most users install the whole of casacore. Anyhow, your changes are all improvements, so I'll merge them.

PR #1057, opened 2020-09-15
Remove user provided methods

rtobar:
In C++11 generation of implicitly-declared copy assignment operator is deprecated if a user-provided copy constructor is given, and vice-versa. In new-ish versions of gcc (9+) a very visible set of warnings is issued because of this, and therefore the situation needed to be tackled.

These two commits remove a few user-provided methods that prevent the generation of other implicitly-defined ones. In all cases removing these user-defined methods leads to no loss, as the compiler will still implicitly define them, and with the expected behavior.

There was also a misleading comment in one of the constructors of the classes *deriving* from `MVPosition`: the comment above the constructor read "Copy constructor", but in reality the constructor took always an `const MVPosition &` argument. I updated the comment and the class documentation to indicate that these classes can be created from an `MVPosition`, while the actual copy constructor is now implicitly defined.

### DIFF
--- a/casa/Quanta/MVAngle.h
+++ b/casa/Quanta/MVAngle.h
@@ -286,8 +286,6 @@ class MVAngle {
     // Construct from type and precision (present due to overlaoding problems)
     Format(uInt intyp, uInt inprec) :
       typ((MVAngle::formatTypes) intyp), prec(inprec) {;};
-    Format(const Format &other) :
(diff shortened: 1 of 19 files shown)


rtobar:
Added an extra commit with similar changes producing similar warnings regarding implicitly-defined methods being deprecated.

tammojan:
Thanks!

PR #1059, opened 2020-09-18
Use H5free_memory instead of free

rtobar:
H5free_memory was introduced in 1.8.13 and is the way one should free memory allocated via the HDF5 library. Using free has worked until now, but local tests against HDF5 1.12 are broken due to this (they segfault).

When building against HDF5 versions older than 1.8.13 we simply define H5free_memory = free. There is no clear statement in casacore about which HDF5 versions are supported, so I played it conservatively and avoided using the H5_VERSION_LE macro introduced in 1.8.7, and used a boolean logic combination to check the major/minor/release components of the HDF5 version at compile time.


tammojan:
Thanks!

PR #1060, opened 2020-09-23
Improve error message when chmod fails

aroffringa:
Copying a dir causes a call to chmod. When this fails
(e.g. because a second process has deleted the dir in the mean time), Casacore
causes a very cryptic message. This PR improves the error message
somewhat.

### DIFF
--- a/casa/OS/Directory.cc
+++ b/casa/OS/Directory.cc
@@ -315,7 +315,10 @@ void Directory::copy (const Path& target, Bool overwrite,
 	command = "chmod -Rf u+w '";
 #endif
 	command += targetName.expandedName() + "'";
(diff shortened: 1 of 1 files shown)


aroffringa:
The CI failure seems to be a random failure not due to my changes.

tammojan:
Thanks. Indeed the `tObjectID` failure on MacOS g++ is unrelated; however I'm afraid not random.

rtobar:
Yes, that's beucase an ObjectID's creation time (int #secs since UNIX epoch) is a multiple of 256 and the test fails. This is part of the assertions that are done over the ObjectIDs hash, but it turns out there's very little one can for sure say about the hash.

PR #1061, opened 2020-09-25
allow for direct specification of IERS data with AppState

schiebel:
CASA is moving toward supporting user configuration via a Python config file that is evaluated. These configuration settings are made available in C++ via the AppState class. There is already a dataPath( ) option which provides a list of directories which MeasIERS searches for IERS data. In general this has worked well, but some CASA stakeholders want to explicitly set the specific path that is used instead of leaving it up to a path search. This modification adds a measuresDir( ) option to AppState which take precedence over the dataPath( ) in MeasIERS.

This change should have no effect outside of CASA. The default value returned by measuresDir( ) is an empty string which MeasIERS ignores. To change this, a casacore user (like CASA) derives from AppState and provides a non empty string to direct MeasIERS to a specific path to look for the IERS tables. Since measuresDir( ) is specific to the measures module, an exception is thrown if the returned directory does not exist.

### DIFF
--- a/casa/System/AppState.h
+++ b/casa/System/AppState.h
@@ -65,6 +65,14 @@ public:
         return result;
     }
 
+    // get directory containing IERS measures data
+    // an exception is thrown if it does not exist or
+    // does not contain the IERS tables
+    virtual std::string measuresDir( ) const {
+        static std::string result;
+        return result;
+    }
+
     virtual bool initialized( ) const { return false; }
 
     virtual ~AppState( ) { }
--- a/casa/System/AppState.h
+++ b/casa/System/AppState.h
@@ -65,9 +65,11 @@ public:
         return result;
     }
 
-    // get directory containing IERS measures data
-    // an exception is thrown if it does not exist or
-    // does not contain the IERS tables
+    // Get AppState specified directory for (IERS) measures data.
+    //
+    // If data is not found in the specified directory and the
+    // specified directiory name is nonempty (size > 0), an
+    // exception will be thrown in findTab.
     virtual std::string measuresDir( ) const {
         static std::string result;
         return result;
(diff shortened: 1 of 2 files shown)


schiebel:
The failed test was due to exceeding allotted time for this ticket.

casa/System/AppState.h
```diff
@@ -65,6 +65,14 @@ class AppState {
         return result;
     }
 
+    // get directory containing IERS measures data
+    // an exception is thrown if it does not exist or
+    // does not contain the IERS tables
```
tammojan, line 70: Suggested comment: get AppState specified directory for (IERS) measures data. If data is not found in the specified directory and the specified directiory name is nonempty, an exception will be thrown in findTab.

casa/System/AppState.h
```diff
@@ -65,6 +65,14 @@ class AppState {
...
+    virtual std::string measuresDir( ) const {
+        static std::string result;
+        return result;
```
tammojan, line 73: This function in the current form does not make sense, is it overridden by a CASA specific override?

tammojan (commented):
Setting the measures directory through the API was about the only way was not yet possible, so a good addition.

Can this be tested?

casa/System/AppState.h
schiebel: Yes, CASA derives a class from AppState which includes more configuration settings (that casacore does not care about) and then registers an object created from that derived class for use as the AppState object. casacore gets the base class object and casa gets the derived class object.

casa/System/AppState.h
schiebel: Thanks! Sounds good to me. I'll push up the change.

schiebel:
> Can this be tested?

Are there any tests which fetch and unpack the measures data tar file from Astron? The test could fetch, unpack, configure the AppState and then try to initialize measures. I think this would introduce a good bit of uncertainty and use time for something that currently only CASA uses. I can't think of a quick test (other than one that just tests C++ virtual function invocation).

tammojan:
If I understand correctly the new code does not allow to set the measures directory from casacore; so then there would be nothing to test. Ok with me if the test lives at the CASA side.

tammojan (approved):

PR #1064, opened 2020-10-01
Merge GAIN_CURVE tables.

kettenis:
Adjusting ANTENNA_ID and SPECTRAL_WINDOW_ID numbers when appropriate.

### DIFF
--- a/ms/MSOper/MSConcat.cc
+++ b/ms/MSOper/MSConcat.cc
@@ -354,6 +354,9 @@ IPosition MSConcat::isFixedShape(const TableDesc& td) {
   // WEATHER
   copyWeather(otherMS.weather(), newAntIndices);
 
+  // GAIN_CURVE
+  copyGainCurve(otherMS, newAntIndices);
+
   /////////////////////////////////////////////////////
 
   // copying all subtables over to otherMS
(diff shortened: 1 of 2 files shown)


kettenis:
This addresses some of the concerns raised by @dpetry in pull request #1040 .

dpetry:
Hi Mark, I guess you essentially copied the copyFeed method?

kettenis:
Not exactly; there is no copyFeed() method.  The copyAntennaAndFeed() handles feeds in conjunction with antennas, which doesn't really apply to copyGainCurve().
Instead I looked at copySysCal() for the code to renumber the antennas and added code to renumber spectral windows that is loosely based on how copyAntennaAndFeed() does things.

There is no attempt to avoid duplication of gain curves. I don't think that's necessary as the number of gain curves is small, and having the same gain curve for an antenna covering overlapping time ranges isn't a problem.

dpetry:
Yeah, right, copyAntennaAndFeed(). Sorry (I wrote it very long ago, ca. 10 years, I think).
In that case, how did you test it and did you also test the case where there are no gain curves?
I.e. did you also run the concat task unit test in CASA?

kettenis:
Tested this by running importfitsidi on EVN experiments EP103A and EP103B, where I added a GAIN_CURVE table to the FITS-IDI files. Since those two epochs use different bands, this excercises the spectral window renumbering.

I also ran the full CASA casatools testsuite and casatasks testsuite, which didn't reveal any regressions. That includes tests for importfitsidi and concat that run on data without gain curves.

PR #1065, opened 2020-10-06
Restore Quantum user-defined copy constructor

rtobar:
The removal of the copy-constructor in the `Quantum` class brought unintended consequences, breaking one of the CASA tests  (original report [here](https://github.com/casacore/casacore/commit/e5d8484b5108f0a890ddd9d494c2efcab738ce7c#r42799565) by @dmehring). This test is present in the casacore codebase, but is skipped by default as it requires some extra data hosted by NRAO, making the problem invisible to the Travis CI environment.

After some investigation, it turned out this was due to the different semantics of the copy constructor and copy assignment of the casacore Array classes (see [this explanation](https://github.com/casacore/casacore/commit/e5d8484b5108f0a890ddd9d494c2efcab738ce7c#r42992327)).

This PR brings back the user-defined copy-constructor, which needs to *assign* `other.qVal` rather than direct initializing it via copy construction if one wants `Quantum` to perform deep-copying of Vectors at copy-construction time. The copy-assignment operator is also brought back (if one is defined, the other should also be -- that was the whole point of #1057), but defaulted since the default behavior is already what we need.

The existing `tQuantum` test is also being augmented to include deep-copying checks for the Vector class, thus preventing this bug is introduced again.

### DIFF
--- a/casa/Quanta/Quantum.h
+++ b/casa/Quanta/Quantum.h
@@ -276,6 +276,8 @@ template <class Qtype> class Quantum : public QBase{
   //# Constructors
   // Default constructor, generates '0'
   Quantum();
+  // Copy constructor (deep copy)
+  Quantum(const Quantum<Qtype> &other);
   // Construct undimensioned quantum (i.e. unit="")
   Quantum(const Qtype &factor);
   // Construct dimensioned quantum (e.g. '1.23 km/Mpc')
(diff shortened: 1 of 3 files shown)


dmehring:
Would someone be kind enough to merge this please? We really need it for CASA.
Thanks

dpetry (approved):
Looks good to me. I assume you have tested it well.

PR #1066, opened 2020-10-23
Handle FITS-IDI GAIN_CURVE tables for a single band

kettenis:
When only a single band is present, certain columns become scalar.
Handle this in the code and warn about this if we it is unexpected.

While there, suppress a similar warning for the SYSTEM_TEMPERATURE
table when there is only one band.


kettenis:
This issue was uncovered by a test observation at the VLBA that we intend to use as a test data set for the CASA importfitsidi that happened to have only a single band since we wanted something small.
The bigger BA123A observation that I used for initial testing of this feature still produces the same results.

PR #1067, opened 2020-10-26
Only skip over a subtable if there still is data to skip

kettenis:
This improves upon commit 8f1127361ea745c9eecc78adc4c4c45e9f677f8b
by handling the case where the subtable contains a small amount
of data that has already been read when parsing the table header.

This fixes skipping the GATEMODL subtable found in VLBA pulsar
bin data.


kettenis:
Thanks @dpetry

PR #1062, opened 2020-09-30
Fix compilation warnings with g++ 10.2

cquike:
This PR fixes some warnings that I have seen while compiling with g++ 10.2.

The warnings that are now fixed are:
```
casacore_github/casacore/casa/OS/HostInfo.cc: In static member function ‘static casacore::Double casacore::HostInfo::secondsFrom1970()’:
/home/cgarcia/PPS/src/casacore_github/casacore/casa/OS/HostInfo.cc:114:28: warning: ‘int ftime(timeb*)’ is deprecated [-Wdeprecated-declarations]
  114 |     AlwaysAssert(ftime(&ftm) >= 0, AipsError);

casacore/msfits/MSFits/MSFitsInput.cc:1740:9: warning: unused variable ‘nrows’ [-Wunused-variable]
 1740 |     Int nrows = bt.nrows();

casacore/tables/DataMan/test/tIncrementalStMan.cc:563:34: warning: comparison of integer expressions of different signedness: ‘int’ and ‘casacore::uInt’ {aka ‘unsigned int’} [-Wsign-compare]
  563 |         AlwaysAssertExit (vec[i] == row/10000*10000);

casacore/tables/TaQL/test/tTableExprData.cc:144:27: warning: catching polymorphic type ‘class std::exception’ by value [-Wcatch-value=]
  144 |   } catch (std::exception x) {

casacore/measures/Measures/test/tMeasMath.cc:250:39: warning: type qualifiers ignored on cast result type [-Wignored-qualifiers]
  250 |      nt1.getEqoxAngle(45882.2,"''") - (const Double) 0.2 *
      |                                       ^~~~~~~~~~~~~~~~~~

casacore/measures/Measures/test/tMeasMath.cc:253:40: warning: type qualifiers ignored on cast result type [-Wignored-qualifiers]
  253 |      nt1.getEqoxAngle(45882.04,"''") - (const Double) 0.04 *

casacore/casacore/ms/MeasurementSets/MSIter.h: In constructor ‘casacore::MSIter::MSIter(const casacore::MeasurementSet&, const std::vector<std::pair<casacore::String, casacore::CountedPtr<casacore::BaseCompare> > >&)’:
/home/cgarcia/PPS/src/casacore_github/casacore/casacore/ms/MeasurementSets/MSIter.h:471:26: warning: ‘casacore::MSIter::timeComp_p’ will be initialized after [-Wreorder]
  471 |   CountedPtr<MSInterval> timeComp_p; // Points to the time comparator.
[...]

PR #1069, opened 2020-11-10
Fix missing initialization of MSIter members in the constructor

cquike:
This PR fixes a bug where a few members of MSIter weren't properly initialized in the constructor.
Additionally a couple of compiler warnings have been fixed.


cquike:
I have rebased this branch on top of #1062  to avoid a conflict when merging to master. So it should be better to merge #1062 before this branch.

PR #1070, opened 2020-11-11
Fix further compiler warnings

rtobar:
These are in addition to those fixed in #1069 and #1062, yielding a cleaner compilation of casacore in newer compilers.

Most of the changes are making switch case fallthrough explicit. Most of them looked legit; one looked like it needed a break though, so I'd appreciate if someone double-checked this is correct.


aroffringa:
I have merged this branch into #1080. Therefore, this PR can be closed. Thanks for creating this @rtobar .

Issue #1077, opened 2020-12-08
Table system update of keywordset does not respect lock and set table.dat to 0 size


kgolap:
When multiple processes try to update keywords in a disk based table  the lock on the on the table is not respected.
So the keywords update suffers from race conditions ..sometimes part gets updated or not.
Also, more importantly,  it has been noted that when updating the keywords the other processes that are trying to access the table would throw an exception that table.dat size is now 0. @cquike traced it to the file being open with O_TRUNC somehow.
This appears for example when one process is trying to do PagedImage::setimageinfo and others are trying to read part of the data.

cquike:
Note that table.dat is open with O_TRUNC and then filled again with the proper information. So the problem of the race condition happens when a process just opens with O_TRUNC and the second process tries to access table.dat *before* the first process had the chance to properly fill table.dat.

cquike:
I can think of two possible solutions:

1. The constructor of Table checks for the existence of a lock *before* trying to open table.dat. Right now it tries to open table.dat to verify that this is actually a table, by reading the first bytes of the file and looking for some "magic number". This test could easily be done after grabbing the lock. One can even say that the existence of the lock is in itself enough to guarantee that we are dealing with a table.
2. When updating the keywords, rather than opening with O_TRUNC one could open normally and seek the file descriptor to the place where the keywords are written. That should leave the first bytes where the magic number is intact and the check performed by the second process will succeed.

cquike:
I attach a code snippet that nicely shows the problem (github doesn't allow me to upload the file with .cc extension, therefore the .txt, sigh....). You have to compile with mpi support. To run it:
$ mkdir MYTMP
$ /opt/casa/02/bin/mpirun -np 4 tPagedImage
I am rank 2 of 4
I am rank 3 of 4
I am rank 0 of 4
I am rank 1 of 4
Exception caught: FilebufIO::readBlock - incorrect number of bytes read for file /scratch/cgarcia/build/casa-el7-ts2-dev02/casa/casacore/MYTMP/tPagedImage_tmp.img1/table.dat

Basically the program updates the image info in process0 and the other ones try to open the image. If you increase the time each secondary process waits (line 78) the problems disappears (in my case with 300000 nsec), which demonstrates that this is a race condition.


[tPagedImage.cc.txt](https://github.com/casacore/casacore/files/5667820/tPagedImage.cc.txt)

cquike:
Hi,
[...]

PR #1078, opened 2021-01-21
Avoid race condition when a process updates table.dat

cquike:
When updating table.dat file because table has changed, write before to a temporary file and the do an atomic renaming. Otherwise, the update of table.dat is done with open() flag O_TRUNC, which effectively empties the file for a small time window. 
If a second process tries to open the table exactly during that small time window then it fails (#1077).

Note a couple of additional remarks:

- The fix is done at the BaseTable level, in the function that writes the table.dat file. The fix is in two functions: writeStart() and writeEnd(). Those functions take a AipsIO& ios which is assumed to be the same object in both calls. This is the case for all the uses of writeStart/writeEnd in PlainTable, ConcatTable and RefTable.
- This fixes the code snippet that I posted in #1077. That code snippet uses in PagedImage, which in turn uses the PlainTable subclass. I have not tested whether this works properly in ConcatTable and RefTable, but in all three cases the use of writeStart/writeEnd is pretty straightforward: within the same code section writeStart() is called, additional things are done and then writeEnd() is called.
- Note that a second process might have already open table.dat but not started to read (or just read a few bytes). This is not a problem since the POSIX rename() function guarantees that any file descriptors already using the old version of the file will continue to have access to the old content, while new open() will see the new content. Since rename() is atomic there is no chance that a process sees a partially updated version of the file, which was the problem reported in #1077. Note also that the part of the file table.dat being read by the second process Table constructor is just the beginning of the file which contains the magic header that tells that this is indeed a table. This content should be the same both for the old and new version of the file.

### DIFF
--- a/tables/Tables/BaseTable.cc
+++ b/tables/Tables/BaseTable.cc
@@ -338,18 +338,24 @@ void BaseTable::flushTableInfo()
 
 void BaseTable::writeStart (AipsIO& ios, Bool bigEndian)
 {
-    //# Check option.
+    // Check option.
     if (!openedForWrite()) {
-	throw (TableInvOpt ("BaseTable::writeStart",
-			    "must be Table::New, NewNoReplace or Update"));
+        throw (TableInvOpt ("BaseTable::writeStart",
+                            "must be Table::New, NewNoReplace or Update"));
     }
-    //# Create table directory when needed.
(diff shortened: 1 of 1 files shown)


tammojan:
Thanks a lot for this fix, and sorry to keep you waiting so long!

PR #1080, opened 2021-02-10
Fix several warnings

aroffringa:
This fixes in particular a lot of 'fallthrough' warnings inside switch statements. To suppress the warning, I use the (C++14) attribute '[[fallthrough]]' to annotate / suppress fallthrough. Although this is a C++14 attribute, C++11 compilers ignore unknown attributes, which I think is therefore fine. A few warnings remain of which I'm not 100% sure they are intended to fallthrough. Also one fallthrough seemed clearly unintended (in SDPolarizationHandler), so this fixes a bug.

Some classes also declared either an explicit default copy constructor or copy assignment operator, causing a warning if the non-declared counter part is used. In FitsBase some of these warnings remain, but since it's not a trivial copy in those cases, it looks wrong to explicitly declare a default constructor.

BTW the test tFITSImage fails locally on my computer on master (already before the fix)

### DIFF
--- a/fits/FITS/fits.h
+++ b/fits/FITS/fits.h
@@ -91,15 +91,14 @@ class FitsLogical {
 	friend ostream & operator << (ostream &o, const FitsLogical &);
     public:
 	FitsLogical() : v('\0') { }
-	FitsLogical(Bool x) { v = (x == True ? 'T' : 'F'); }
-	FitsLogical(const FitsLogical &x) : v(x.v) { }
-	FitsLogical & operator = (const FitsLogical &x) { 
-		v = x.v; return *this; }
+	FitsLogical(Bool x) : v(x == True ? 'T' : 'F') { }
 	FitsLogical & operator = (Bool x) { 
 		v = (x == True ? 'T' : 'F'); return *this; }
-	Bool isdefined() const { return v == '\0' ? True : False; }
+  ///ARO 2021-02-20:
+  ///Removed the following function, because it seems incorrectly implemented and isn't used
+	///Bool isdefined() const { return v == '\0' ? True : False; }
 	void undefine() { v = '\0'; }
-	operator Bool() { return (v == 'T' ? True : False); }
+	operator Bool() const { return v == 'T'; }
     protected:
 	char v;
 };
@@ -118,11 +117,8 @@ class FitsBit {
     public:
 	FitsBit() : bit_array(0) { }
 	FitsBit(unsigned char x) : bit_array(x) { }
-	FitsBit(const FitsBit &x) : bit_array(x.bit_array) { }
-	FitsBit & operator = (const FitsBit &x) { 
-		bit_array = x.bit_array; return *this; }
 	FitsBit & operator = (unsigned char x) { bit_array = x; return *this; }
-	operator unsigned char() { return bit_array; }
+	operator unsigned char() const { return bit_array; }
     protected:
 	unsigned char bit_array;
 };
--- a/fits/FITS/hdu2.cc
+++ b/fits/FITS/hdu2.cc
@@ -862,13 +862,6 @@ int *FitsBase::vdim() {
 	return &no_elements; 
 }
 //====================================================================================
-/*FitsBase & FitsBase::operator = (FitsBase &x) {
-	if (fieldtype() == x.fieldtype() &&
-	    nelements() == x.nelements())
-	    memcpy(data(),x.data(),localfieldsize());
-	return *this;
-}*/
-//====================================================================================
 FitsBase * FitsBase::make(const FITS::ValueType &type,int n) {
 	switch (type) {
 	    case FITS::LOGICAL: return (new FitsField<FitsLogical> (n));
--- a/fits/FITS/hdu2.cc
+++ b/fits/FITS/hdu2.cc
@@ -862,12 +862,12 @@ int *FitsBase::vdim() {
 	return &no_elements; 
 }
 //====================================================================================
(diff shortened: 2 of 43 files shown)


rtobar:
I haven't checked the contents, but sounds a lot like #1070

aroffringa:
Ay, they're quite overlapping indeed.

I'm happy to merge the two -- a good check whether we agree on deliberate fall throughs :).

I wouldn't declare a CASACORE_FALLTHROUGH symbol; it's described how attributes are to be handled by the C++ compilers, so I think it's better to just use straight the C++ standard for it instead of making it more complicated.

rtobar:
The `fallthrough` attribute was only introduced in C++17, and I'm not sure what compilers *not* supporting that (casacore requires only 11, right?) would output if we used it without any checks; that's why I defined the `CASACORE_FALLTHROUGH`macro differently depending on the occasion (with the first option always being to use the standard).

No strong feelings though; if older compilers don't complain about an unsupported attribute being there then there's no issue.

aroffringa:
Agree, yes I think the target compiler requirement is >= c++11 for casacore. Since C++17, compilers must ignore unknown attributes. Older compilers might still warn: gcc <7 still warns, while >= 7 do not, e.g. https://godbolt.org/z/KKGEn6 . I think it's acceptable that older compilers warn, but I'm not sure everyone agrees...

cquike:
Some builds of CASA are still using fairly old compilers (5.3.1) so it would be good to keep warning free compilations with those ones, if possible.

aroffringa:
I think that's @rtobar 's approach should avoid warnings on older compilers (as well as newer ones), I'll compare this branch with his and will take his approach of defining CASACORE_FALLTHROUGH.

aroffringa:
I have merged the branch of @rtobar into this branch, making use of his CASACORE_FALLTHROUGH definition (thanks @rtobar !) and I've solved a few more compiler warnings. A handful of warnings remain; some bison/flex warnings remain, and a few final warnings in ancient code.

fits/FITS/fits.h
```diff
@@ -91,15 +91,14 @@ class FitsLogical {
 	friend ostream & operator << (ostream &o, const FitsLogical &);
     public:
 	FitsLogical() : v('\0') { }
-	FitsLogical(Bool x) { v = (x == True ? 'T' : 'F'); }
-	FitsLogical(const FitsLogical &x) : v(x.v) { }
-	FitsLogical & operator = (const FitsLogical &x) {
```
tammojan, line 96: Quite a lot of functions disappear here, why?

fits/FITS/hdu2.cc
```diff
@@ -862,12 +862,12 @@ int *FitsBase::vdim() {
 	return &no_elements; 
 }
 //====================================================================================
-FitsBase & FitsBase::operator = (FitsBase &x) {
```
tammojan, line 865: Better just delete this.

scimath/Fitting/FitGaussian.tcc
```diff
@@ -134,7 +134,7 @@ void FitGaussian<T>::setNumGaussians(uInt numgaussians)
   itsRetryFctr.resize();
   itsFirstEstimate.resize();
   itsMask.resize();
-  if (itsDimension*3 && itsNGaussians) {
+  if (itsDimension*3!=0 && itsNGaussians) {
```
tammojan, line 137: (Needs spaces around operator.)

scimath/Functionals/Function.h
```diff
@@ -219,14 +219,6 @@ class RecordInterface;
      template <class W, class X>
      Function(const Function<W,X> &other) : param_p(other.parameters()),
      arg_p(0), parset_p(other.parsetp()), locked_p(False) {}
-     Function(const Function<T,U> &other) :
```
tammojan, line 222: Why does this disappear?

scimath/Functionals/Function1D.h
```diff
@@ -84,7 +84,6 @@ template<class T, class U=T> class Function1D : public Function<T,U> {
   explicit Function1D(const uInt n) : Function<T,U>(n) {}
   explicit Function1D(const Vector<T> &in) : Function<T,U>(in) {}
   Function1D(const FunctionParam<T> &other) : Function<T,U>(other) {}
-  Function1D(const Function1D<T,U> &other) : Function<T,U>(other) {}
```
tammojan, line 87: Why does this disappear?

fits/FITS/fits.h
aroffringa: For copy constructors, they are redundant: the default copy constructor is better. A manual copy constructor disables the assignment operator (or a move constructor), although gcc so far accepted us of an assingment operator like that. Defining only one of a copy constructor or copy assignment operator causes a warning when the other one is used.

scimath/Functionals/Function.h
aroffringa: Same reason as before, this copy constructor causes warnings, the default is fine.

scimath/Functionals/Function1D.h
aroffringa: Same reason as before, this copy constructor causes warnings, the default is fine.

fits/FITS/fits.h
tammojan: Thanks, clear.

tammojan:
It's ok with me if you merge this.

aroffringa:
I can't merge, but maybe @tammojan you can give me rights? (Or someone else at ASTRON?)

PR #1079, opened 2021-02-10
Fix two erros reported by valgrind in IPosition

cquike:
This PR fixes two small errors reported by valgrind in IPosition:
- uninitialized values in IPosition constructor. This has been triggered by the use of the IPosition constructor with just the length without specifying the values of the vector. I guess this constructor is not used much, but the unit test tCpp11Features uses it in lines 350 and 394.
- memory leak in IPosition move constructor.

PR #1081, opened 2021-02-10
Revert zeroing IPosition and some clean-up

aroffringa:
This reverts one part of #1079 (by @cquike). An IPosition does not need to be zeroed; it's part of its interface to leave it uninitialized by default. The valgrind error that cquike reported was actually an invalid use of IPosition, which I've corrected in the test.

It's open to discussion whether the IPosition *should* be like that or whether we better chose for the safe option, but currently it's often declared and then filled.

On top of that I also went through IPosition and used std::copy, std::fill and std::move where appropriate to replace for loops, and this simplifies/removes a lot of for statements.

### DIFF
--- a/casa/Arrays/IPosition.cc
+++ b/casa/Arrays/IPosition.cc
@@ -41,7 +41,6 @@ IPosition::IPosition (size_t length)
     if (length > BufferLength) {
 	allocateBuffer();
     }
-    memset(data_p, 0, size_p * sizeof(ssize_t));
 }
 
 IPosition::IPosition(std::initializer_list<ssize_t> list)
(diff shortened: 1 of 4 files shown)


cquike:
Hi,
the rationale behind my change was to avoid undefined behaviour. Maybe the documentation of the constructor should be even more explicit, in the sense that it stresses that the object cannot be fully used until all its values are properly initialized.
However I agree that your changes make more sense, since my change had an implicit change in semantics. As you say, whether that is desired is a separate discussion.

gervandiepen:
I think we should leave it uninitialised. std::vector behaves the same way.

gervandiepen:
Furthermore, clearly leaving it undefined showed an error which may have
gone unnoticed otherwise.

tammojan:
Thanks Ger, so it is good that we reverted the zeroing.

aroffringa:
@gervandiepen `std::vector` default constructs its elements. This means it does initialize build-in types like int to zero. The behaviour of `IPosition` does therefore not match `std::vector`.

PR #1086, opened 2021-02-12
Switch to github actions (#1085)

gijzelaerr:


gijzelaerr:
This is mostly done, only OS X is failing since it can't find the installed fortran compiler. It can on my MacBook, so maybe you know what is wrong @tammojan  ?

PR #1090, opened 2021-02-17
Cas 11860 Undo incomplete SYSPOWER table support for exportuvfits

dmehring:


dmehring:
@tammojan these failures appear unrelated to my changes

```
anyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

The command "./.travis/before_script.sh" failed and exited with 1 during .

Your build has been stopped.
```

tammojan:
Ouch, this is what we get for using free services that eventually start charging.

Luckily @gijzelaerr is already working on migrating to Github actions, maybe (hopefully) that will also solve the docker limit.

I suggest that we fall back to testing locally for now.

cquike:
FYI: we have setup regular casacore builds at the ESO Jenkins. For the time being it is only master but I will try to figure out how to do it for the PRs as well.
By the way, we are also using valgrind on the unit tests to have a better QA. It will also help to avoid false positives when we run valgrind with CASA. For the time being there are still a few errors reported by valgrind, but I will post patches on this ones soon.

cquike:
I forgot to comment: our Jenkins is internal and unfortunately cannot be accessed from outside. If IT ever gives us a solution on how to expose the Jenkins results to the outside I will inform you. Sorry for that inconvenience.

dmehring:
Thanks for the responses @tammojan and @cquike . @tammojan is there anything I can or need to do at this point? This change is to revert work I did a few months ago that is now on master. As it turns out, that work was incomplete, and so if not reverted, improper behavior may result. So, getting the reversion on master asap is the best course of action as I won't have time to complete implementing the full, correct behavior in the near future.
Thanks

tammojan:
I've built and tested locally, I think the change is innocent enough to skip the full cross-platform test.

dmehring:
Great! Thank you.

PR #1091, opened 2021-02-18
Use setenv() rather than putenv() to set environmental variables.

cquike:
setenv() is prefered over putenv() as descirbed in the POSIX standard:
https://pubs.opengroup.org/onlinepubs/9699919799/functions/putenv.html#

Addtionally this fix a small memory leak, since with putenv one
cannot modify the input buffer and therefore the buffer must remain allocated. With setenv() that's not the case and is therefore cleaner.


aroffringa (approved):
Looks good to me.

PR #1094, opened 2021-02-20
Rewrite the IPosition(size, args...) constructor to use a parameter pack

aroffringa:
The old constructor was inefficient. To prepare for deprecation of this constructor,
I've added a Make() factory function which is to replace this constructor for
cases where a initializer list isn't possible (i.e. inside a macro).

### DIFF
--- a/casa/Arrays/IPosition.cc
+++ b/casa/Arrays/IPosition.cc
@@ -115,47 +115,6 @@ IPosition::IPosition (size_t length, ssize_t val)
     std::fill_n(data_p, size_p, val);
 }
 
-// <thrown>
-//    <item> std::runtime_error
-// </thrown>
-IPosition::IPosition (size_t length, ssize_t val0, ssize_t val1, ssize_t val2,
-                      ssize_t val3, ssize_t val4, ssize_t val5, ssize_t val6,
-                      ssize_t val7, ssize_t val8, ssize_t val9)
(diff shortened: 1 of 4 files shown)


aroffringa:
These changes also highlighted a bug in the test tables/DataMan/test/tTiledShapeStMan.cc: this test provided an incorrect size to the var arg IPosition constructor. This is fixed.

tammojan:
Nice refactor! It would be good if some tests were changed to use the new constructor.

I wonder why you enforce >=3 parameters: the original IPosition could also have length 1 - I think `IPosition(1, 2)` was a valid constructor call, and I'm not sure it still is with the your rewrite.

aroffringa:
The reason for requiring >=3 parameters is that otherwise there is ambiguity between
```
IPosition(size_t length, ssize_t val);
```
and the new vararg constructor,
```
 template<typename... Vals> IPosition (size_t /*dummy*/, ssize_t val1, ssize_t val2, Vals... vals)
```
If val2 would be left out, IPosition(3, 5) would match both constructors, because `Vals... vals` may be empty.

I can add some tests, will take a moment.

aroffringa:
I've now added a unit test for Make(). The constructor is actually replaced and is therefore already thoroughly tested, so I haven't added new ones for that.

tammojan:
Thanks!

Issue #1092, opened 2021-02-19
Fix OSX build on github actions


gijzelaerr:

derwentx:
I'll be attempting this in my own fork as a time-boxed spike over the next few days. If anyone has any advice, I'm all ears :) I'll add my running notes to this ticket.

Currently the osx.yml GitHub Actions CI is failing at the `configure` stage for me. 

```txt
Run cd build
-- The C compiler identification is AppleClang 12.0.0.12000032
-- The CXX compiler identification is AppleClang 12.0.0.12000032
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Applications/Xcode_12.3.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Applications/Xcode_12.3.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Deprecation Warning at CMakeLists.txt:6 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


CMake Deprecation Warning at CMakeLists.txt:14 (cmake_policy):
  The OLD behavior for policy CMP0048 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.


-- Performing Test HAVE_O_DIRECT
-- Performing Test HAVE_O_DIRECT - Failed
-- Using namespace casacore.
-- Performing Test HAS_CXX11
-- Performing Test HAS_CXX11 - Success
-- Looking for pread
-- Looking for pread - found
-- Looking for pwrite
-- Looking for pwrite - found
-- The Fortran compiler identification is unknown
CMake Error at CMakeLists.txt:292 (enable_language):
  No CMAKE_Fortran_COMPILER could be found.

  Tell CMake where to find the compiler by setting either the environment
[...]

PR #1097, opened 2021-02-22
[casacore#1092] 👷 Fix OSX build on github actions

derwentx:
What I did:

- fix https://github.com/casacore/casacore/issues/1092
- re-enabled the build script in `.github/workflows/osx.yml` (removed `continue-on-error`)
- set Fortran compiler with `-DCMAKE_Fortran_COMPILER=gfortran-10` (GitHub Actions quirk)
- made minimal syntax changes to the build script so that it would work with GitHub Actions (e.g. `cd build` in each job)
- removed job count limit with `make -j` (reduces build time by 3 minutes!)

### DIFF
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -16,14 +16,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: install homebrew deps
-        run: |
-          brew install fftw hdf5 boost-python3 cfitsio wcslib wget gfortran
-
-      - name: debug gfortran
-        continue-on-error: true
-        run: |
-          set -ex
-          which gfortran-10
+        run: brew install fftw hdf5 boost-python3 cfitsio wcslib wget gfortran
 
       - name: pip install numpy
         run: pip3 install numpy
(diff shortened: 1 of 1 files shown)


.github/workflows/osx.yml
```diff
@@ -34,13 +33,28 @@ jobs:
...
+            -DBUILD_PYTHON3=ON \
+            -DBoost_NO_BOOST_CMAKE=True \
+            -DCMAKE_Fortran_COMPILER=gfortran-10 \
+            -DDATA_DIR=. ..
 
       - name: build
-        run: make -j3
+        run: |
+          cd build
+          make -j
```
gijzelaerr, line 50: i think this still forks unlimited processes, which might slow down the compilation. the github actions documentation states that the osx vm's have 3 cores available.

gijzelaerr (commented):
great work, thanks!

.github/workflows/osx.yml
derwentx: ah ok, feel free to change it back. For me, it was 40 minutes with `-j3` and 37 with `-j`

gijzelaerr:
thanks!

PR #1100, opened 2021-03-01
Remove the custom typing system of casacore

aroffringa:
Casacore was using a manual way to identify runtime type information,
using many virtual type() methods and Register() overloads taking
dummy variables. All this functionality was easy to replace with
'normal' run-time type info using dynamic_cast or typeid(), saving
a lot of code, and improving performance in several places.

Removing these type() and variant methods highlighted several bugs
with measures like MDirection where it was
supposed to call mDirection.getRefPtr()->getType()
(which returns an enum describing something like ITRF), but
instead called mDirection.type(), returning an int that
was assigned to the MDirection type by the custom type system.

Many classes still have an 'assure()' method whose 
purpose it is to throw an exception when the object is not of a specified
type. These are now useless, as one can just perform
dynamic_cast<T&>, which throws when the type is not of the
specified type. However, I left those here for now.


aroffringa:
I think this MR is ready for review. It removes a nice amount of code :). It might be good to check if this affects Casa in anyway.

aroffringa:
@gervandiepen Is this an MR that you can review?

gervandiepen:
Yes, I can review it. It has to wait till next week.

aroffringa:
Thanks Ger. I've opened a new MR, because this MR was still in my private casacore repo which I removed in the mean time. The new MR is #1172, I'll close this one.

gervandiepen:
In #1100 you comment that getRefPtr()->getType() returns an enum, but isn't it returning a uInt (from MRBase or MeasRef)?

PR #1102, opened 2021-04-22
MSIter: Fix call to table sort.

cquike:
The MSIter class calls the table sort() method and later on calls TableIterator without sorting because the sort already took place. However the sorting criteria for both calls is different, which might lead to inconsistencies. In the call to table sort() no sorting functions are specified, which would mean that the default sorting functions are used. This is fixed by this PR.

Since until now almost no code sets sorting functions different from the default ones there has been no impact from this bug. However in CASA there will be more use of non-default sorting functions in the future (mainly to support the SPW combination in VI/VB2).

### DIFF
--- a/ms/MeasurementSets/MSIter.cc
+++ b/ms/MeasurementSets/MSIter.cc
@@ -343,7 +343,7 @@ void MSIter::construct(const Block<Int>& sortColumns,
     if (!useIn && !useSorted) {
       // we have to resort the input
       if (aips_debug) cout << "MSIter::construct - resorting table"<<endl;
-      sorted = bms_p[i].sort(columns, Sort::Ascending, Sort::QuickSort);
+      sorted = bms_p[i].sort(columns, objComp, orders, Sort::QuickSort);
     }
 
     // Only store if globally requested _and_ locally decided


tammojan:
Interestingly, the tests fail on your change only on MacOS:
https://github.com/casacore/casacore/pull/1102/checks?check_run_id=2409188926#step:9:675

I have no idea what causes this, but this looks like a genuine test failure since it's tMSIter that fails:
```
        Start 335: tMSIter
335/487 Test #335: tMSIter ..........................***Failed    0.83 sec
39,41c39,41
< nrow=1 a1=0 a2=0 time=[30] keyCh=TIME
< nrow=2 a1=0 a2=0 time=[90, 150] keyCh=TIME
< nrow=2 a1=0 a2=0 time=[210, 270] keyCh=ANTENNA2
---
> nrow=2 a1=0 a2=0 time=[30, 90] keyCh=TIME
> nrow=2 a1=0 a2=0 time=[150, 210] keyCh=TIME
> nrow=1 a1=0 a2=0 time=[270] keyCh=ANTENNA2
FAIL (output not verified): ./tMSIter
```

cquike:
Thank you for catching that. I will drop this PR and will create a different one, since I think to rethink about it.

PR #1103, opened 2021-05-01
MSIter: Compute DDId information on demand.

cquike:
Some metadata related to Data Description that has been computed at every iteration (i.e, when calling next()) is now computed less often to improve performance. This is driven by a change in CASA that now uses the MSIter class in the subchunk iteration loop of the VI/VB2. Usually each subchunk has single timestamp and there is no need to query metadata such as the DDID, SpwID or PolID, since this is done in the chunk iteration. With this PR only if the application asks for such information will MSIter compute (and cache) it, rather than computing it unconditionally in each next() call. In a test MS that saves 50 ms per subchunk iteration on a total budget of 400 ms.

Specifically:

1) If DDId is in the sorting columns then the information is always computed, since it likely that the application will request DD related metadata. This is typically the case in the CASA VI/VB2 chunk iteration.

2) If not, then the DDId related information is computed only if any of the accesors to retrieve it is called. This is typically the case in the subchunk iteration of the CASA VI/VB2. In this case  the information is cached internally for the next time. The relevant accessors are frequency(), frequency0(), colDataDescriptionIds(), spectralWindowId(), polarizationId(), dataDescriptionId(), polFrame().

 As a reference, it has been verified that after this change the following CASA tests pass: test_agentflagger, test_bandpass, test_calanalysis, test_concat, test_coordsys,  test_cvel, test_fixvis, test_flagcmd, test_flagdata, test_flagmanager, test_hanningsmooth, test_listpartition, test_mstransform, test_mstransform_mms, test_partition, test_plotcal, test_plotms, test_req_task_listobs, test_req_task_statwt, test_sdbaseline, test_sdfit, test_sdfixscan, test_sdgaincal, test_sdsmooth, test_sdtimeaverage, test_split, test_tclean, test_uvcontsub, test_uvcontsub3.

### DIFF
--- a/ms/MeasurementSets/MSIter.cc
+++ b/ms/MeasurementSets/MSIter.cc
@@ -413,7 +413,7 @@ MSIter::operator=(const MSIter& other)
   lastFieldId_p = other.lastFieldId_p;
   curSpectralWindowIdFirst_p = other.curSpectralWindowIdFirst_p;
   lastSpectralWindowId_p = other.lastSpectralWindowId_p;
-  curPolarizationId_p = other.curPolarizationId_p;
+  curPolarizationIdFirst_p = other.curPolarizationIdFirst_p;
   lastPolarizationId_p = other.lastPolarizationId_p;
   curDataDescIdFirst_p = other.curDataDescIdFirst_p;
   lastDataDescId_p = other.lastDataDescId_p;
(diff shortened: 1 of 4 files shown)


tammojan:
I see no test coverage of the added caching. Would it be possible to test this in test/tMSIter.cc ?

cquike:
A unit test has been added to tMSIter that retrieves the DD related information in a number of different cases:
- when the DD is part of the sorting columns (in this case it is always computed)
- when the DD is not part of the sorting column. In this case it is also exercised that the query of the DD related information is not done for every row. That's really where the lazy feature will be exercised, since for a number of rows there won't be any computation.

tammojan:
Thanks for including the tests. The speed improvements are very nice!

PR #1104, opened 2021-05-05
 MSIter: Compute Feed information on demand.

cquike:
Some metadata related to the feed that has been computed at every iteration (i.e, when calling next()) is now computed only when some feed related metadata is requested.

Similar to #1103 this is driven by a change in CASA that now uses the MSIter class in the subchunk iteration loop of the VI/VB2. Usually each subchunk has single timestamp and there is no need to query metadata related to the feed.
With this change, the MSIter next() method reduces its execution time from ~400 microseconds to  ~100 microseconds, which is a big improvement since the next() is called very often in each subchunk.

In particular, the feed related metadata is computed (and cached) when the following methods are called: CJones(), CJonesAll(),
receptorAngle(), receptorAngles(), getBeamOffsets(), allBeamOffsetsZero().

As a reference, it has been verified that after this change the following CASA tests pass: test_agentflagger, test_bandpass, test_calanalysis, test_concat, test_coordsys, test_cvel, test_fixvis, test_flagcmd, test_flagdata, test_flagmanager, test_hanningsmooth, test_listpartition, test_mstransform, test_mstransform_mms, test_partition, test_plotcal, test_plotms, test_req_task_listobs, test_req_task_statwt, test_sdbaseline, test_sdfit, test_sdfixscan, test_sdgaincal, test_sdsmooth, test_sdtimeaverage, test_split, test_tclean, test_uvcontsub, test_uvcontsub3.

Note that this PR depends on #1103, so it probably should be merged only after that one.

### DIFF
--- a/ms/MeasurementSets/MSIter.cc
+++ b/ms/MeasurementSets/MSIter.cc
@@ -148,6 +148,15 @@ void MSIter::construct(
         sortColumnNames[iCol] = element.first;
         sortCompareFunctions[iCol] = element.second;
         ++iCol;
+        if (element.first == "DATA_DESC_ID") {
+            ddInSort_p = true;
+        } else if (element.first == "TIME") {
+            timeInSort_p = true;
+        } else if (element.first == "FIELD_ID") {
+            fieldInSort_p = true;
+        } else if (element.first == "ARRAY_ID") {
+            arrayInSort_p = true;
+        }
     }
 
     // Create the table iterators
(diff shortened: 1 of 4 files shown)


cquike:
A unit test has been added to tMSIter that retrieves the Feed related information in a number of different cases:
- when the Feed is part of the sorting columns (in this case it is always computed).
- when the Feed is part of the sorting columns but not the fastest changing keyword. In this case the feed related metadata is always computed but it might not change for every iteration.
- when the Feed is not part of the sorting column. In this case it is also exercised that the query of the Feed related information is not done for every row. That's really where the lazy feature will be exercised, since for a number of rows there won't be any computation. There is a loop on the number of iterations for which no information is retrieved, so different cases are tested.

tammojan:
Thanks, also for adding the unit tests!

PR #1106, opened 2021-05-06
TableIter: Add support for caching the iteration boundaries that have been created while sorting the tables

cquike:
This PR allows the caching of the iteration boundaries by the table iterators. Those boundaries are cheap to obtain while sorting the table in the first place.  

This avoids calling the sorting functions again when doing a next() iteration call as it has been done until now, since a vector with the iteration boundaries is cached and hence the information of which rows are part of the next iteration is readily available.

The mechanism is disable by default and controlled by a boolean switch in the  TableIterator constructor. For the time being only MSIter makes (partial) use of it.

As a reference, it has been verified that after this change the following CASA tests pass: test_agentflagger, test_bandpass, test_calanalysis, test_concat, test_coordsys, test_cvel, test_fixvis, test_flagcmd, test_flagdata, test_flagmanager, test_hanningsmooth, test_listpartition, test_mstransform, test_mstransform_mms, test_partition, test_plotcal, test_plotms, test_req_task_listobs, test_req_task_statwt, test_sdbaseline, test_sdfit, test_sdfixscan, test_sdgaincal, test_sdsmooth, test_sdtimeaverage, test_split, test_tclean, test_uvcontsub, test_uvcontsub3.

### DIFF
--- a/casa/Utilities/Sort.cc
+++ b/casa/Utilities/Sort.cc
@@ -181,7 +181,7 @@ Sort::Sort (const Sort& that)
 
 Sort::~Sort()
 {
-    for (uInt i=0; i<nrkey_p; i++) {
+    for (size_t i=0; i<nrkey_p; i++) {
 	delete keys_p[i];
     }
 }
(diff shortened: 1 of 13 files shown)


gervandiepen:
A good idea to use the sorted values for getting the boundaries. It indeed
saves you reading the comparison column(s) again, which can be expensive as
they are read in non-sequential order.
I think it can be used in all cases, so the new boolean argument is not
needed.

One little remark.
When having an if (or for) statement split over multiple lines, it is good
practice to always use braces.

cquike:
Thank you @gervandiepen  for your comments. I have added the braces as suggested.

cquike:
Some unit tests have been added for the following new functionality:
- The unique() method in Sort that allows to retrieve the sort key that will change in the next sorting group.
- The new constructor for TableIterator that allows to explicitly set whether to cache the results of the iteration or not is tested in both cases to make sure that the results are the same with and without caching.

@gervandiepen: Since this is needed for some urgent performance fixes in CASA coming late in the release cycle I have decided to make it an optional feature so that the changes are minimized. I agree that eventually the option can go away and make it the default. I will probably create a new PR in view of including that in the next CASA release.

PR #1108, opened 2021-05-26
Table iterator: Reuse a RefTable for each iteration

cquike:
This PR will use an existing RefTable created at construction time in the BaseTableIterator class  to avoid paying the cost of a RefTable creation every time the next() method is called to advance iteration.

 When next() is called then the RefTable is reset with a call to RefTable::removeAllRow() and then the rows for this iteration are added. For that, also a convenient function addRownrRange() has been added to RefTable, to avoid calling addRownr() in a tight loop.

In a dataset with 10 rows per iteration the cost goes down from ~80 usecs per next() call to ~13 usecs.

Note that this PR depends on PR #1106 and should merged after that one.

### DIFF
--- a/tables/Tables/BaseTabIter.cc
+++ b/tables/Tables/BaseTabIter.cc
@@ -56,7 +56,8 @@ BaseTableIterator::BaseTableIterator (BaseTable* btp,
   lastVal_p (keys.nelements()),
   curVal_p  (keys.nelements()),
   sortIterBoundaries_p   (nullptr),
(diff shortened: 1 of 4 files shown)


tammojan:
Sorry if I merged in the wrong order, but this now has a merge conflict. Could you have a look at that?

cquike:
Thank you for merging the other PRs!
I have pushed the changes in this branch that solve the conflicts.

cquike:
Looks like the tests passed after resolving the conflicts.

PR #1109, opened 2021-05-27
MSIter: Compute field attributes on demand.

cquike:
Similar to PR #1103  and #1104  some metadata like field attributes (fieldID, field name) and related attributes (restFrequency and phaseCenter) are now computed on demand only if it is asked for. For clients that don't require retrieving of field information for each iteration there is no need to compute that metadata.
Note that the relevant metadata accessors are phaseCenter(), fieldId(), fieldName(), restFrequency().

### DIFF
--- a/ms/MeasurementSets/MSIter.cc
+++ b/ms/MeasurementSets/MSIter.cc
@@ -148,6 +148,15 @@ void MSIter::construct(
         sortColumnNames[iCol] = element.first;
         sortCompareFunctions[iCol] = element.second;
         ++iCol;
+        if (element.first == "DATA_DESC_ID") {
+            ddInSort_p = true;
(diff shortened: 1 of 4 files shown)


cquike:
A unit test has been added to tMSIter that retrieves the Field related information in a number of different cases:
- when the Field is part of the sorting columns (in this case it is always computed).
- when the Field is part of the sorting columns but not the fastest changing keyword. In this case the feed related metadata is always computed but it might not change for every iteration.
- when the Field is not part of the sorting column. In this case it is also exercised that the query of the Field related information is not done for every row. That's really where the lazy feature will be exercised, since for a number of rows there won't be any computation. There is a loop on the number of iterations for which no information is retrieved, so different cases are tested.

Issue #1111, opened 2021-06-08
Table remains in cache after PR #1084 merge


cquike:
CASA is trying to update to the latest casacore and some tests are failing due to #1084 merge. I managed to create a very small code snippet that reproduces the problem:
```
    {
      casacore::MeasFrame mframe_;
      MeasComet comet("EPHEM0_Sol_58327.6.tab");
      mframe_.set(comet);
    }

    auto& cache = PlainTable::tableCache();
    const Vector<String> &openTables = cache.getTableNames();

    for (auto& table : openTables)
        cout << " ERROR: Table open " << table << endl;
```
This produces the output:
` ERROR: Table open EPHEM0_Sol_58327.6.tab`
whereas before #1084 was merged everything was fine.

The table cache is checked in some CASA tests to verify that no open tables are left and with this problem the tests are not passing.
Additionally valgrind reports memory leaks with the code snippet above. I will attach the valgrind report soon.
The EPHEM0_Sol_58327.6.tab table is not a particular table, I guess any ephemerides tabe would produce the same result but if needed I can provide a copy

cquike:
Please find attached logfile from valgrind executed on the previous code snippet.
[valgrind.log](https://github.com/casacore/casacore/files/6614317/valgrind.log)

aroffringa:
Do I need `EPHEM0_Sol_58327.6.tab` to run this? Where can I find that file?

cquike:
I have attached a tarball with the table here. BTW, you have to substitute EPHEM0_Sol_58327.6.tab in the code snippet with the full path.
[EPHEM0_Sol_58327.6.tab.tar.gz](https://github.com/casacore/casacore/files/6614517/EPHEM0_Sol_58327.6.tab.tar.gz)

cquike:
Thanks @aroffringa  for looking into this. I think the problem is rep->mymcf inside MeasFrame. That creates an object of type MCFrame stored in a smart pointer. However MCFrame itself stores a copy of the MeasFrame, So there is kind of a circular dependency in the ownership.
I have created a patch that I will post as a PR that you can review or simply serve as inspiration :)

cquike:
Hi @aroffringa for your help on fixing this! Unfortunately this is not yet the end of the story. Now the CASA test is still failing with the same reason with a slightly different pattern in its the use of MeasComet:

```
    {
      MEpoch ep;
      MPosition pos;
      MDirection dir;

[...]

PR #1112, opened 2021-06-08
Break circular ownership between MeasFrame and MCFrame.

cquike:
The MeasFrame class owns a MCFrame object which itself owns a copy of its parent MeasFrame. This causes problems with memory leaks and potential table being left open. This PR simply stores a pointer to the parent MeasFrame in the MCFrame.

This should solve #1111. After this change there are no more tables open and valgrind doesn't report more significant problems (there are still leaks related to LogIO, but those are not relevant here).

@aroffringa: of you can review this it would be great. Or maybe you can find a better solution based on the idea of this commit. Thanks!


aroffringa:
I think this solution looks fine, thanks for solving @cquike ! It might be good to add a comment in MCFrame.h before the constructor, that the MeasFrame pointer is stored in the class and must exist for the duration of MCFrame.

measures/Measures/MCFrame.h
```diff
@@ -96,7 +96,7 @@ class MCFrame {
   
   //# Constructors
   // Construct using the MeasFrame parent
-  MCFrame(MeasFrame &inf);
+  MCFrame(MeasFrame *inf);
```
aroffringa, line 99:
```suggestion
  // The MeasFrame pointer inf is stored inside this MCFrame. When MeasFrame is destructed,
  // the MCFrame must also be destructed.
  MCFrame(MeasFrame *inf);
```

aroffringa (approved):

cquike:
Some macOS tests failed, so maybe there is more work to do on this :(

aroffringa:
I rewrote the circular-dependant code somewhat, amongst other to use a weak_ptr. While the weak_ptr adds a bit of complexity, ownership has become more clear now, up to a level that I think I can understand what happens :). If CI succeeds, I think it should be ready to merge after someone reviews it.

measures/Measures/MCFrame.cc
```diff
@@ -487,114 +508,117 @@ void MCFrame::makeEpoch() {
...
     delete static_cast<MDirection::Convert *>(dirConvJ2000);
-    dirConvJ2000 = 0;
+    dirConvJ2000 = nullptr;
   }
-  dirConvJ2000 = new MDirection::Convert(*(myf.direction()),
+  std::unique_ptr<casacore::Measure>& direction = myf.lock()->dirval;
+  dirConvJ2000 = new MDirection::Convert(*direction,
 					 MDirection::Ref(MDirection::J2000,
-							 this->myf));
+							 MeasFrame(this->myf.lock())));
```
cquike, line 579: I think this is the key of the failure in my patch. If I add this then the test that was failing is now working. Great!

aroffringa: I thought what went wrong in your case was this scenario:
- Make a MeasFrame A (creates a Rep, let's say R. R points via a MCFrame to A.)
- Copy MeasFrame A to B. B points also to R, R points via a MCFrame to A.
- Destruct A. R still exists because B points to it. R points to non-existing A.
That scenario I tackled by letting MCFrame point directly to R instead of to a MeasFrame.

aroffringa:
Thanks @cquike . Should we ask @tammojan for a quick review and merge afterwards? 

PS: I think it would be nice to in the future also change all the raw pointers like 'epTDBp' to std::unique_ptrs (and replace the void* ptrs by more apropriate ones), it would simplify the code quite a bit. I resisted doing it for now since I don't want to introduce more changes and possibly new bugs in this PR ;).

cquike:
As discussed in #1111 this PR will be dropped since is still causing problems with the CASA tests. The branch https://github.com/cquike/casacore/tree/measframe_leak will remain to help in the future to fix the problems that #1084 tried to solve originally.

PR #1115, opened 2021-06-24
Add support for X-Y mounts

kettenis:
This makes it possibly to apply the (co)-parallactic angle correction needed for VLBI data processing of data that includes antennas with an X-Y mount. An example of such an antenna is the 26m antenna at Hobart that is part of the Australian Long Baseline Array (LBA).
This code has been tested by applying the parallacting angle correction to data for LBA observation VT23 in both AIPS and CASA and comparing the phases on baselines that include the Hobart antenna.

### DIFF
--- a/ms/MSOper/MSDerivedValues.cc
+++ b/ms/MSOper/MSDerivedValues.cc
@@ -269,7 +269,7 @@ Double MSDerivedValues::parAngle()
   Double pa=0;
 
   if (mount_p(antenna_p)==0 || mount_p(antenna_p)==4 ||
-      mount_p(antenna_p)== 5) {
+      mount_p(antenna_p)==5) {
     // Now we can do the conversions using the machines
     mRADecInAzEl_p     = cRADecToAzEl_p();
     mHADecPoleInAzEl_p = cHADecToAzEl_p();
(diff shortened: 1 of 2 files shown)


tammojan:
Is the swapping of 2 and 3 intentionally part of this pull request?

If so, I'm curious about the implications of this change. Which code checks for these numbers? Only software outside casacore?

kettenis:
Hi Tammo Jan,

> Is the swapping of 2 and 3 intentionally part of this pull request?

Yes. Without that change, importing a FITS-IDI file will result in an MS that has Hobart (which is an antenna with an X-Y mount) in its ANTENNA table with an "ORBITING" mount type (which is intended to be used for antennas in space, i.e. for space VLBI). So this has to be fixed in order to have a working parallactic angle correction for (Australian) LBA data.

> If so, I'm curious about the implications of this change. Which code checks for these numbers? Only software outside casacore?

Only code *inside* casacore checks these numbers. There are a few places where numbers are used in casacore to encode mount types. The code in FitsIDItoMS.cc is used to convert the numbers defined for FITS-IDI files into strings for use in the MS. The code in MSDerivedValues.cc then converts the MS strings back into numbers but this is only a mapping for internal use. It is somewhat confusing that these mappings are no longer the same, but this isn't wrong. We could consider changing that internal mapping to match FITS-IDI and UVFITS though.

Data that was converted using the old (current) casacore (i.e. all released versions of CASA) will of course have mislabeled the mount type in the MS. But there is nothing we can do about that. I'll make sure this is mentioned in the CASA release notes when this change makes it into CASA.

tammojan:
> We could consider changing that internal mapping to match FITS-IDI and UVFITS though.

That would be a good idea, otherwise the code will be too confusing for the occasional reader. I guess around line 194 of MSDerivedValues.cc. If it is not changed, at least a warning should be in the code that explains future generations that it is not a problem that they're swapped.

In an ideal world, an enum would be used so that the actual numbers would be hidden and we could search for `ORBITING` instead of `3`. This would also make the lines like `else if (mount_p(antenna_p)==2)` a bit more readable. But I can imagine that refactoring the code to use an enum is outside the scope of this work.

kettenis:
Pushed a change that adjusts the internal mapping such that it matches the mapping used by FITS-IDI and UVFITS.
Using an enum would indeed be a bit better, but would probably require exposing the enum in the MSDerivedValues.h header and I don't think that is something we really want to do.

PR #1116, opened 2021-07-05
Adios2stman improvements

rtobar:
This set of commits improves on the flexibility of the `Adios2StMan` by allow it to be configured via properties, and exposing its specification to external users. This allows us to remove a number of static members that were previously used to transmit state between different storage manager copies, and to flexibly configure instances of the storage manager from non-C++ environments like python or taql.

The storage manager also confused its "name" with its "type name", so that distinction has been clarified. This was changed in such a way that serialization from/to the `AipsIO` object remains backwards compatible.

cc: @JasonRuonanWang

### DIFF
--- a/tables/DataMan/Adios2StMan.cc
+++ b/tables/DataMan/Adios2StMan.cc
@@ -35,10 +35,11 @@ namespace casacore
 {
 
 // Static objects in impl class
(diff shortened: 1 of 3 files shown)


JasonRuonanWang:
Thank you @rtobar . This looks great.

PR #1117, opened 2021-07-14
Make Dysco part of casacore tables

### LINKED ISSUE #876: Include the DyscoStMan in Casacore
Dysco is a (lossy) compressing storage manager used for LOFAR (and is also used occasionally for MWA and possibly Apertif). I've discussed with @tammojan and @gervandiepen that it would be useful to have Dysco in Casacore, most of all to avoid version issues (e.g., versions of casa can't load Dysco because its casacore version doesn't match the one compiled with Dysco).

Dysco is currently hosted here: https://github.com/aroffringa/dysco.

aroffringa:
A much requested feature, this MR imports the Dysco source into casacore and makes sure that it gets installed along. This avoids binary incompatibility between Casa and Dysco, which so far has prevented users to make use of Casa on modern LOFAR sets (which are always compressed). It includes the source from https://github.com/aroffringa/dysco literally. It can be switched off with the BUILD_DYSCO cmake variable, that is ON by default.

This fixes #876.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,7 @@ option (BUILD_PYTHON3 "Build the python3 bindings" NO)
 
 option (BUILD_DEPRECATED "build and install deprecated classes (such as Map)" NO)
 option (BUILD_FFTPACK_DEPRECATED "build FFTPack" NO)
+option (BUILD_DYSCO "Build the Dysco compression storage manager" YES)
 
 # By default build shared libraries
 option (ENABLE_SHARED "Build shared libraries" YES)
@@ -579,6 +580,7 @@ message (STATUS "FFTW3 library? ........ = ${FFTW3_LIBRARIES}")
 
 message (STATUS "BUILD_FFTPACK_DEPRECATED= ${BUILD_FFTPACK_DEPRECATED}")
 message (STATUS "BUILD_DEPRECATED ...... = ${BUILD_DEPRECATED}")
+message (STATUS "BUILD_DYSCO ........... = ${BUILD_DYSCO}")
 message (STATUS "BUILD_PYTHON .......... = ${BUILD_PYTHON}")
 message (STATUS "BUILD_PYTHON3 ......... = ${BUILD_PYTHON3}")
 
--- a/tables/CMakeLists.txt
+++ b/tables/CMakeLists.txt
@@ -209,6 +209,10 @@ if(MPI_FOUND)
     endif(ADIOS2_FOUND)
 endif(MPI_FOUND)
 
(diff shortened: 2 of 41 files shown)


kettenis:
Hi Andre,
I seem to have built a version of CASA with built-in Dysco support using the import-dysco branch.  I ran into some issues though:

- The CASA build infrastructure that builds the CASA Python modules assumes C++ files use a .cc extension (which is what the existing casacore C++ code files use) but the Dysco files use .cpp as the extension. I think it would make sense to rename the files you're adding.
- I had to add some code to tables/DataMan/DataManager.cc to register the DyscoStMan. But doing this may conflict with how you build Dysco as a separate shared library for normal casacore?

So far I've checked that listobs does work for a LOFAR ILT dataset that I had lying around. I'll see if I can do some further testing, but that may have to wait until Des Small is back from vacation.

aroffringa:
Hi Mark, thanks for the checks. Renaming the files to `.cc` files shouldn't be a problem, I'll do that.

I'm not sure why you needed to register the library. You should see a libdyscostdman.so in your `lib` dir, and casacore should find that lib (given all dirs are correctly set). Or is the lib dir of Casa not automatically in the search path (because Casa doesn't use it otherwise) ?

I think we _could_ register dysco as you did without issue. If it's registered, it would override searching for an external lib and always use the one provided with casacore. That might be difficult when developing dysco as a separate lib -- basically with newer casacore the separate lib is never used unless Dysco is disabled inside casacore. But I guess that's what we want :). If I want to develop Dysco further (there are ideas for improvements) I could develop it inside (a fork of) casacore, or build a casacore without Dysco, that's fine to me (and probably necessary anyway -- even without registration inside casacore).

Does your fix call include the dyscostman.h header and `register_dyscostman()` ? If so can you commit that to this branch?

kettenis:
Hi Andre,

The CASA team changed the way casacore is integrated into the CASA build with CASA 6. They no longer build casacore as shared libraries using cmake, but instead they use the Python setup tools to build a Python module out of the casacore C++ code files. One can question the wisdom in that (I certainly did), but it does seem to allow them to build precompiled Python wheels for various Linux/MacOS versions and different Python versions with relative ease.
I'll do a branch with my change and check if I can still build casacore using cmake with it.

kettenis:
This is the change I had in my casa6 tree: https://github.com/kettenis/casacore/commit/cb08f778a31b133c6c4b22d74d3e809a6b84f6bb

However, that fails to build with cmake because the Dysco code is in a different shared library. So for this to work, the Dysco code should be made part of the casa_tables library like what is done for the ADIOS2 storage manager.

PR #1118, opened 2021-07-26
fix insufficient precision issue in T+dT syntax

tnakazato:
I noticed that timerange selection by "T+dT" syntax omits the sub-seconds value. More specifically, if we write "HH:MM:SS.ss", ".ss" part is omitted in parsing "T+dT" string. I think this is a bug because "T1~T2" syntax handles sub-seconds value properly. This pull request is possible fix for the bug.

### DIFF
--- a/ms/MSSel/MSTimeGram.yy
+++ b/ms/MSSel/MSTimeGram.yy
@@ -174,11 +174,12 @@ rangetimeexpr: yeartimeexpr DASH yeartimeexpr
 
 		   Time time0($1.year,$1.month,$1.day,$1.hour,$1.minute,s0);
 		   Time time1($3.year,$3.month,$3.day,$3.hour,$3.minute,s1);
-		   Double mjd=time1.modifiedJulianDay()*86400.0;
-
-		   time1 = time0 + mjd;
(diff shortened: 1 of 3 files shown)


tammojan:
Well spotted, thanks for the fix! Could you perhaps add a test for this? This could be done by altering `tMSTimeGram.run` (copying the structure from `tMSSelection.run`), and making a reference output `tMSTimeGram.out`. The measurement set `mssel_test_small.ms` should contain enough time slots to come up with a useful test, in particular it has 23 time slots spaced 1.92 seconds apart.

tnakazato:
Hi, @tammojan. I will try adding the test. By the way, I pushed further improvement using MVEpoch.

tnakazato:
Added minimal set of verification tests in tMSTimeGram. The test checks if T+dT" works properly. As a reference, "T1~T2" test, which selects effectively the same time range, is also added. I confirmed that the test passes on the branch while fails with the code before fix.

tnakazato:
Any update on this? Please let me know if you have more suggestion. Thanks!

tammojan:
Thank you, this change looks good to me. Also good to have TimeGram tested now!

tnakazato:
Thanks!

PR #1125, opened 2021-10-05
Replace dysco files with those from the dysco repos after fixing some bugs

aroffringa:
This fixes https://github.com/aroffringa/dysco/issues/21


aroffringa:
@tammojan I think there's little to review; this is just the replacement of all dysco files with those of the dysco repos, which you already reviewed ;). I ran a formatter before, but apparently it hadn't used the correct clang formatting script, so the sources have been reformatted with our 'schaap' standard formatting script, so many files are changed because of this -- but now they are correctly formatted.

tammojan (approved):
In principle it would be good to separate the bug fix from the formatting change. Otherwise fine; it's ok with me to merge as it is now.

PR #1127, opened 2021-10-12
Replace leftofer Mutex/ScopedMutexLock with std::mutex

### LINKED ISSUE #1126: Compilation failure when USE_STACKTRACE is enabled
With USE_STACKTRACE set to ON, Exceptions/Error2.cc fails to compile because it has some old Mutex and ScopedMutexLock variables (now deprecated and removed, ). Looks like these were left after Mutex was removed (#1095 ?)
The errors look like:
```
/scratch/src/casacore/casa/Exceptions/Error2.cc:99:10: error: ‘Mutex’ does not name a type
   static Mutex  lastErrorMutex;
          ^~~~~
/scratch/src/casacore/casa/Exceptions/Error2.cc: In static member function ‘static void casacore::AipsError::getLastInfo(casacore::String&, casacore::String&)’:
/scratch/src/casacore/casa/Exceptions/Error2.cc:103:5: error: ‘ScopedMutexLock’ was not declared in this scope
     ScopedMutexLock lock(lastErrorMutex);
     ^~~~~~~~~~~~~~~
/scratch/src/casacore/casa/Exceptions/Error2.cc:103:5: note: suggested alternative: ‘pthread_mutex_lock’
     ScopedMutexLock lock(lastErrorMutex);
...
```

FedeMPouzols:
Fixes #1126
I also did the same change for a few commented out Mutex instances in other files, to prevent future confusion.

### DIFF
--- a/casa/Exceptions/Error2.cc
+++ b/casa/Exceptions/Error2.cc
@@ -96,11 +96,11 @@ AipsError::~AipsError() noexcept
   // stack trace from last exception
   static String lastStackTrace = "*no-stack-trace*";
   // protects the lastMessage and lastStackTrace statics
-  static Mutex  lastErrorMutex;
+  static std::mutex  lastErrorMutex;
 
   void AipsError::getLastInfo (String & message, String & stackTrace)
   {
-    ScopedMutexLock lock(lastErrorMutex);
+    std::lock_guard<std::mutex> lock(lastErrorMutex);
     message = getLastMessage();
     stackTrace = getLastStackTrace();
   }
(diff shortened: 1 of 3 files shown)


casa/Exceptions/Error2.cc
```diff
@@ -96,11 +96,11 @@ AipsError::~AipsError() noexcept
   // stack trace from last exception
   static String lastStackTrace = "*no-stack-trace*";
   // protects the lastMessage and lastStackTrace statics
-  static Mutex  lastErrorMutex;
+  static std::mutex  lastErrorMutex;
```
aroffringa, line 99: One should probably also add an "#include <mutex>" to Error2.cc to make sure std::mutex is available.

aroffringa:
Thanks for the fix @FedeMPouzols ; seems like a missed these uses of Mutex in my MR.

casa/Exceptions/Error2.cc
FedeMPouzols: Good point. I've opened PR #1128 to add an "include mutex" if we want to ensure it is included in this file.
I was curious where mutex was being included from. It turns out  this file uses mutexes only when USE_STACKTRACE (inside ifdefs) and in that case there is an include of <casacore/casa/System/AipsrcValue.h> which includes mutex.

PR #1128, opened 2021-10-18
Explicit include of mutex in Error2.cc

FedeMPouzols:
This is a follow up on #1126.
Added an explicit include of <mutex> (following discussion from PR #1127). Not strictly needed but probably more clear, and less dependent on what AipsrcValue.h decides to include.


aroffringa:
Looks good, thanks!

PR #1132, opened 2021-11-09
Support writing FITS files with more than 256 antennas

tammojan:
This change is inspired by the routine fantbas from Miriad.

@kettenis Could you review this please? I have no way to properly test this.

### DIFF
--- a/msfits/MSFits/MSFitsOutput.cc
+++ b/msfits/MSFits/MSFitsOutput.cc
@@ -1343,8 +1343,18 @@ std::shared_ptr<FitsOutput> MSFitsOutput::_writeMain(Int& refPixelFreq, Double& 
         *odate2 = dayFraction;
 
         // BASELINE
-        *obaseline = antnumbers(inant1(tbfrownr)) * 256 + antnumbers(inant2(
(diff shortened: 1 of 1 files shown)


kettenis:
Hi @tammojan,
I think this is the wrong approach. For all practical purposes UV-FITS is standardized in AIPS memo 117:

`ftp://ftp.aoc.nrao.edu/pub/software/aips/TEXT/PUBL/AIPSMEM117.PDF`

Eric Greisen added support for more than 256 antennas in AIPS and in the process revised the memo. Instead of trying to encode the baseline antenna numbers and subarray number in a single 32-bit variable, new SUBARRAY, ANTENNA1 and ANTENNA2 random parameters are defined.

Given the interoperability requirements of CASA with AIPS, I don't think we should implement the approach taken by Miriad in casacore. Also, I think the AIPS approach is better since it avoids ambiguities.

kettenis:
P.S. I'll reply to the mail exchange as well.

tammojan:
Thanks Mark, I thought this would be a quick fix but apparently not. I look forward to your mail.

PR #1134, opened 2021-11-09
Add -std=legacy compiler option for scimath_f

mpokorny:
Compiling `fftpak.f` with newer versions of gcc (>= 10?) gfortran results in compile time errors. This change allows builds to complete.  I realize that fftpack is deprecated, but some versions of CASA still require it. A more targeted option to try is `-fallow-argument-mismatch`. Testing fortran compiler options in CMake seems not to be possible with the minimum CMake version for casacore, and I'm not sure how to deal with that possibility. I'm also not sure how much effort is warranted here --- I wouldn't mind if the PR were not merged.

### DIFF
--- a/scimath_f/CMakeLists.txt
+++ b/scimath_f/CMakeLists.txt
@@ -37,7 +37,7 @@ if (BUILD_FFTPACK_DEPRECATED)
 endif ()
 
 add_library (casa_scimath_f ${buildfiles})
-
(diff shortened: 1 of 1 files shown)


aroffringa:
@tammojan This option is "strongly discouraged" by gcc. I wouldn't start doing this to be honestly, and revert this change.

PR #1137, opened 2021-11-24
Fixed clang warnings

### LINKED ISSUE #1136: Remove clang warnings
clang (Apple clang version 13.0.0 (clang-1300.0.29.3)) gives several warnings, sometimes about deprecated functionality.
These warnings should be removed.

gervandiepen:
Apple clang version 13.0.0 (clang-1300.0.29.3) gave several warnings which have been fixed.
Most warnings were about using std::move (if e.g. a Vector is returned as Array) or about self-assignment in test programs.

Close #1136

### DIFF
--- a/tables/Tables/NullTable.cc
+++ b/tables/Tables/NullTable.cc
@@ -252,9 +252,11 @@ Bool NullTable::adjustRownrs (rownr_t, Vector<rownr_t>&,
 }
 
 BaseTable* NullTable::doSort (PtrBlock<BaseColumn*>&,
-			      const Block<CountedPtr<BaseCompare> >&,
-			      const Block<Int>&,
-			      int)
+                             const Block<CountedPtr<BaseCompare> >&,
+                             const Block<Int>&,
+                             int,
+                             std::shared_ptr<Vector<rownr_t>>,
+                             std::shared_ptr<Vector<size_t>>)
 {
   throw makeError ("doSort");
 }
--- a/casa/BasicSL/STLMath.h
+++ b/casa/BasicSL/STLMath.h
@@ -99,7 +99,7 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
   {
     std::vector<T> result(left.size());
     std::transform (left.begin(), left.end(), result.begin(),
-                    std::bind(std::divides<T>(), std::placeholders::_1, right));
+                    [right](T x) { return x / right; });
     return result;
   }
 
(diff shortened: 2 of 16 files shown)


tables/Tables/NullTable.cc
```diff
@@ -252,9 +252,11 @@ Bool NullTable::adjustRownrs (rownr_t, Vector<rownr_t>&,
...
 
 BaseTable* NullTable::doSort (PtrBlock<BaseColumn*>&,
-			      const Block<CountedPtr<BaseCompare> >&,
-			      const Block<Int>&,
-			      int)
+                             const Block<CountedPtr<BaseCompare> >&,
+                             const Block<Int>&,
+                             int,
+                             std::shared_ptr<Vector<rownr_t>>,
+                             std::shared_ptr<Vector<size_t>>)
```
aroffringa, line 259: For the future (not for now) it might be a good idea to start using the `override` keyword. It might find some issues. It's possible to enable warnings on not using it with `-Wsuggest-override`.

aroffringa (changes requested):
Looks good! I added some small comments.

gervandiepen:
Thanks for the comments. I've changed STLMath accordingly. I've added override to NullTable.h. It looks as if you have to add it to all functions, otherwise warnings are given for the virtual functions without override.
LattStatsSpecialize is maintained by NRAO, so I leave up to them to remove functions from it.

aroffringa:
Thanks for the changes, looks good to me!

dmehring:
>  LattStatsSpecialize is maintained by NRAO, so I leave up to them to remove functions from it.

I was unaware that this code is now the responsibility of NRAO to maintain, but being the person at NRAO who probably has looked at the lattices code the most over the years, I'd be happy to take a look at this. Would someone please tell me exactly what the issue is?
Thanks

gervandiepen:
LatticeMath is merely used by NRAO. I think that Dave wrote LatticeStats.
The 'problem' is in the LatticeStatsSpecialize::accumulate which has several unused arguments. It is the question if this function is used anywhere. If so, can't these unused arguments not be removed.

dmehring:
I believe LatticeStats pre-dates me, although it is quite possible I've modified it over the years. In any case, CASA does not seem to depend upon LatticeStatsSpecialize::accumulate() at this point, and it was noted in the .h file that these methods were deprecated, so, because no one seems to require them at this point, I've removed them, cf https://github.com/casacore/casacore/pull/1145

PR #1138, opened 2021-11-25
Fix copy of null MS

### LINKED ISSUE #1135: Copying an empty MeasurementSet gives an exception
The following lines will make casacore throw a TableError:
```
const casacore::MeasurementSet ms;
const casacore::MeasurementSet copy(ms);
```
I didn't expect an exception from this simple code. Instead, I expected that both `ms.isNull()` and `copy.isNull()` would be true now.

gervandiepen:
The MS copy constructor and assignment now test for a null MS object.

Close #1135

### DIFF
--- a/ms/MeasurementSets/MeasurementSet.cc
+++ b/ms/MeasurementSets/MeasurementSet.cc
@@ -261,27 +261,30 @@ MeasurementSet::MeasurementSet(const MeasurementSet &other)
   hasBeenDestroyed_p(False)
 {
   doNotLockSubtables_p = other.doNotLockSubtables_p;
-  copySubtables (other); // others will be handled by initRefs
+  if (! other.isNull()) {
+    copySubtables (other); // others will be handled by initRefs
 
-  mainLock_p=TableLock(TableLock::AutoNoReadLocking);
+    mainLock_p=TableLock(TableLock::AutoNoReadLocking);
 
-  // verify that other is valid
+    // verify that other is valid
 
-  if (&other != this) {
+    if (&other != this) {
       addCat();
       if (! validate(this->tableDesc()))
-          throw (AipsError("MS(const MeasurementSet &) - "
-                  "MeasurementSet is not a valid MS"));
-  }
+        throw (AipsError("MS(const MeasurementSet &) - "
+                         "MeasurementSet is not a valid MS"));
+    }
 
-  if (!isNull()){
+    if (!isNull()){
       initRefs();
+    }
   }
 }
 
 MeasurementSet::~MeasurementSet()
 {
 // check to make sure that this MS is still valid
+  if (! isNull()) {
     if (!hasBeenDestroyed_p  &&  !validate()) {
 	// the table is otherwise OK, so ensure that it is written if necessary
 	this->flush();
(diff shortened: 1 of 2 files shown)


ms/MeasurementSets/MeasurementSet.cc
```diff
@@ -261,27 +261,30 @@ MeasurementSet::MeasurementSet(const MeasurementSet &other)
...
       initRefs();
+    }
   }
 }
 
 MeasurementSet::~MeasurementSet()
 {
 // check to make sure that this MS is still valid
+  if (! isNull()) {
     if (!hasBeenDestroyed_p  &&  !validate()) {
```
mnijhuis-tos, line 288: Optional: You could combine the two if statements into one statement.

ms/MeasurementSets/MeasurementSet.cc
```diff
@@ -261,27 +261,30 @@ MeasurementSet::MeasurementSet(const MeasurementSet &other)
   hasBeenDestroyed_p(False)
```
mnijhuis-tos, line 261: The default constructor, which yields a Null MS, initializes `hasBeenDestroyed_p` to True. Can you initialize `hasBeenDestroyed_p` to `other.isNull()` here? Then you also don't need to modify the destructor.

ms/MeasurementSets/MeasurementSet.cc
```diff
@@ -308,11 +312,15 @@ MeasurementSet::operator=(const MeasurementSet &other)
 	mrsDebugLevel_p = other.mrsDebugLevel_p;
 	memoryResidentSubtables_p = other.memoryResidentSubtables_p;
 
-	copySubtables (other);
+        if (! (isNull()  ||  other.isNull())) {
```
mnijhuis-tos, line 315: Why do you need the `isNull()` check? It looks like checking `if (!other.isNull())` should be enough here.

ms/MeasurementSets/MeasurementSet.cc
```diff
@@ -308,11 +312,15 @@ MeasurementSet::operator=(const MeasurementSet &other)
...
+          copySubtables (other);
+        }
```
mnijhuis-tos, line 317: GitHub shows different indentation levels in this function, which is confusing.

mnijhuis-tos (changes requested):

ms/MeasurementSets/MeasurementSet.cc
mnijhuis-tos: I see you now set `hasBeenDestroyed_p` to `other.hasBeenDestroyed_p` if `other` is not null. I think you can directly initialize `hasBeenDestroyed_p` to `other.hasBeenDestroyed_p` here instead, which is a littile simpler.

mnijhuis-tos (changes requested):

gervandiepen:
I think it's better to assign hasBeenDestroyed_p later because an exception might be thrown in the constructor in which case the destructor should not destroy too much.

mnijhuis-tos:
> I think it's better to assign hasBeenDestroyed_p later because an exception might be thrown in the constructor in which case the destructor should not destroy too much.

If an exception occurs in the constructor, the destructor is never called, since destructors are only called for completely initialized objects.

mnijhuis-tos (approved):

PR #1152, opened 2022-01-31
Added support for row tiling

### LINKED ISSUE #1140: Make multi-dim tiles possible in MeasurementSet Tables
Currently the visibility data can be stored in a tiled way, but the rownr can only be used as a single dimension. Hence, it is not possible to tile on the baseline or time axis. It should be made possible that in the TiledColumnStMan (and possibly TiledShapeStMan) the rownr can be mapped to multiple axes. It requires that in some way the size of these axes have to be defined, possibly in the constructor properties.

gervandiepen:
Support for tiling on row number has been added. It makes it possible for a MeasurementSet to use tiles with a limited number of baselines in it. It makes iteration over baselines much faster.
A test program has been added that can test the performance of various way of storing a DATA column. A test on a 3 GB DATA column (using 820 baselines) with 4 or 8 baselines per 1 MB tile  shows a dramatic performance improvement in iterating over baseline compared to storing the rows linearly.

Close #1140

### DIFF
--- a/tables/DataMan/TiledColumnStMan.cc
+++ b/tables/DataMan/TiledColumnStMan.cc
@@ -55,6 +55,15 @@ TiledColumnStMan::TiledColumnStMan ()
 
 TiledColumnStMan::TiledColumnStMan (const String& hypercolumnName,
 				    const IPosition& tileShape,
+				    const IPosition& rowShape,
+				    uInt64 maximumCacheSize)
+: TiledStMan  (hypercolumnName, maximumCacheSize),
+  tileShape_p (tileShape),
+  rowShape_p  (rowShape)
+{}
+
+TiledColumnStMan::TiledColumnStMan (const String& hypercolumnName,
+				    const IPosition& tileShape,
 				    uInt64 maximumCacheSize)
 : TiledStMan  (hypercolumnName, maximumCacheSize),
   tileShape_p (tileShape)
@@ -66,6 +75,9 @@ TiledColumnStMan::TiledColumnStMan (const String& hypercolumnName,
 {
     if (spec.isDefined ("DEFAULTTILESHAPE")) {
         tileShape_p = IPosition (spec.toArrayInt ("DEFAULTTILESHAPE"));
+    }
+    if (spec.isDefined ("ROWSHAPE")) {
+        rowShape_p = IPosition (spec.toArrayInt ("ROWSHAPE"));
     }
     if (spec.isDefined ("MAXIMUMCACHESIZE")) {
         setPersMaxCacheSize (spec.asInt64 ("MAXIMUMCACHESIZE"));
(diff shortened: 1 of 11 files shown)


tables/DataMan/TiledColumnStMan.cc
```diff
@@ -95,28 +108,38 @@ String TiledColumnStMan::dataManagerType() const
     return "TiledColumnStMan";
 }
 
-Bool TiledColumnStMan::canAccessColumn () const
+Bool TiledColumnStMan::canAccessColumn() const
 {
-    return True;
+    // This is only possible if an integral nr of rows fit the hypercube.
+    return nrrow_p % rowMap_p[rowMap_p.size() - 1] == 0;
```
tammojan, line 114: Segfaults with empty rowmap?

tables/DataMan/TiledColumnStMan.h
```diff
@@ -74,6 +74,23 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
...
+//       row shape has to be specified which defines how the rows are
+//       divided over the tiles. For example, in a MeasurementSet the data
+//       are usually 4-dim (say 4 pols, 64 freqs, 90 baselines, ntimes) where
+//       each data array contains 4 pol and 64 freqs. By giving a tile shape
+//       [4,64,400] the baselines and times are flattened into the row numbers
+//       making access in baseline order slowish. But one can tell there are
+//       90 baselines by defining the row shape as [90,0]. Setting the
+//       tile shape to e.g. [4,64,10,40] has the effect that a single
+//       tile will contain only 10 baselines instead of 90. It will speed
+//       up reading in baseline order because lesser tiles have to be read.
```
tammojan, line 90:
```suggestion
//       up reading in baseline order because fewer tiles have to be read.
```

tables/DataMan/TiledColumnStMan.h
```diff
@@ -157,6 +174,8 @@ class TiledColumnStMan : public TiledStMan
     // with the given name. The columns used should have the FixedShape
     // attribute set.
     // The hypercolumn name is also the name of the storage manager.
+    // The cube shaope can be given to know how to divide the rows over the
```
tammojan, line 177:
```suggestion
    // The cube shape can be given to know how to divide the rows over the
```

tables/DataMan/TiledColumnStMan.cc
gervandiepen: Added comment and DebugAssert

tables/DataMan/TiledColumnStMan.h
gervandiepen: Fixed

tables/DataMan/TiledColumnStMan.h
gervandiepen: Fixed

gervandiepen:
The test program tTiledColumnStManPerf can be used to test the performance. This program has been extended slightly to set the maximum cache size.

Issue #1143, opened 2021-11-25
Implement Copy COLUMN in TaQL


gervandiepen:
Note 199 describes the syntax of the COPY COLUMN command in TaQL. However, it has not been implemented yet.

PR #1153, opened 2022-02-01
Created TableUtil containing former static Table functions

gervandiepen:
Class Table contained several static functions which have been moved to the new TableUtil. A few more functions have been added to deal with creating  a new table and handling subtables.
Several casacore classes have been adapted for this change.

This is part of issue #1143.


tables/Tables/TableUtil.cc
```diff
@@ -0,0 +1,321 @@
...
+        // Check if main table exists.
+        if (! Table::isReadable (names[0])) {
+          msg = "main table " + names[0] + " does not exist";
+        } else {
+          tab = Table(names[0]);
+          // Get name of last subtable.
+          if (names.size() > 1) {
+            lastPart = names[names.size()-1];
+            // Check if all subtables exist, maybe except last one.
+            uInt nrname = mustExist ? names.size() : names.size()-1;
```
mnijhuis-tos, line 302: I'm somewhat confused what the return value of `tryFindSubTable` is in different cases: The value of `tab` may vary:
- If `mustExist` is true, `tab` is always the last table, with name `lastPart`.
- If `mustExist` is false, `tab` is the last-but-one table, and `lastPart` is the name of a subtable of `tab` that may or may not exist.

Maybe it's simpler if `tab` is always the last-but-one table: The semantics of the return value are then always the same. At the end of the function, you can then add `if (mustExist && !tab.keywordSet().isDefined(lastPart)) { msg = ...; }`

Also, please clearly document the return value in TableUtil.h.

tables/Tables/test/tTableUtil.cc
```diff
@@ -0,0 +1,276 @@
...
+//# Correspondence concerning AIPS++ should be addressed as follows:
+//#        Internet email: aips2-request@nrao.edu.
+//#        Postal address: AIPS++ Project Office
+//#                        National Radio Astronomy Observatory
+//#                        520 Edgemont Road
+//#                        Charlottesville, VA 22903-2475 USA
+//#
+//# $Id$
+
+//# Includes
```
mnijhuis-tos, line 28: Again, please remove.

tables/Tables/test/tTableUtil.cc
```diff
@@ -0,0 +1,276 @@
...
+// This program tests some table utility functions.
+
+
+// Create a table and various subtables.
+void createTables()
+{
+  // Create a main table.
+  TableDesc td("", "", TableDesc::Scratch);
+  td.addColumn (ScalarColumnDesc<Int>("RowNr"));
+  Table tab(TableUtil::createTable("tTableUtil_tmp/maindata", td, Table::New));
```
mnijhuis-tos, line 50: Use a `const std::string` or `const String` that contains the magic constant "tTableUtil_tmp/maindata". You can then use it throughout this file.

tables/Tables/test/tTableUtil.cc
```diff
@@ -0,0 +1,276 @@
...
+    testDelete();
+    testReplace();
+    testErrors();
+    testLayoutInfo("tTableUtil_tmp/maindata");
+    testLayoutInfo("tTableUtil_tmp/maindata::SubTab2");
+  } catch (const std::exception& x) {
+    cout << "Caught an exception : " << x.what() << endl;
+    return 1;
+  } 
+  cout << "OK" << endl;
```
mnijhuis-tos, line 273: I rather see something more explicit, since anything could have written "OK".
```suggestion
  cout << "tTableUtil: All tests passed." << endl;
```

tables/Tables/test/tTableUtil.cc
```diff
@@ -0,0 +1,276 @@
...
+#include <casacore/tables/Tables/TableUtil.h>
+#include <casacore/tables/Tables/SetupNewTab.h>
+#include <casacore/tables/Tables/TableRecord.h>
+#include <casacore/tables/Tables/TableDesc.h>
+#include <casacore/tables/Tables/ScaColDesc.h>
+#include <casacore/casa/Utilities/Assert.h>
+#include <casacore/casa/OS/Directory.h>
+#include <casacore/casa/iostream.h>
+
+#include <casacore/casa/namespace.h>
```
mnijhuis-tos, line 39: [Google forbids importing an entire namespace.](https://google.github.io/styleguide/cppguide.html#Namespaces) Instead, only use what you need, for example:
```
using casacore::Table;
using casacore::TableDesc;
using casacore::TableUtil;
```

tables/Tables/Table.cc
```diff
@@ -362,6 +363,7 @@ Table& Table::operator= (const Table& that)
     return *this;
 }
 
+/*
```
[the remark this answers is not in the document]
gervandiepen: This was an oversight. Removed it.

tables/Tables/TableUtil.cc
```diff
@@ -0,0 +1,321 @@
...
+                            const String& subtableName,
+                            Bool checkSubTables)
+    {
+      // Get the full table name of the subtable.
+      // Make sure the Table object is deleted, otherwise isOpened is always true.
+      String fullName;
+      {
+        Table table = parent.keywordSet().asTable (subtableName);
+        fullName = table.tableName();
+      }
```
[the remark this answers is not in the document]
gervandiepen: Is it guaranteed that the temporary Table object is deleted before the next line?

tables/Tables/TableUtil.cc
```diff
@@ -0,0 +1,321 @@
...
+               ("TableUtil::getLayout: unknown table kind " + tp));
+      }
+      ios.close();
+      return nrow;
+    }
+
+    TableInfo tableInfo (const String& tableName)
+      {
+        return BaseTable::tableInfo (getFullName(tableName));
+      }
```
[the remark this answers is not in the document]
gervandiepen: I've replaced the tabs by spaces and re-indented all code.
An automatic tool would generate a massive change. Furthermore, different casacore developers usually like their own style and I think it would be hard to find a compromise.

tables/Tables/TableUtil.cc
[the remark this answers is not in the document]
mnijhuis-tos: Yes. From the C++ standard:

12.2 Temporary objects, clause 3: "Temporary objects are destroyed as the last step in evaluating the full-expression (1.9) that (lexically) contains the point where they were created."

1.9 Program execution, clause 12: "A full-expression is an expression that is not a subexpression of another expression."

tables/Tables/Table.cc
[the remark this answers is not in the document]
gervandiepen: It was an oversight that the outcommented code was still present. Removed it.

tables/Tables/TableUtil.cc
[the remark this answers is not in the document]
gervandiepen: Is it guaranteed that the temporary Table object is removed before the next line is executed or has the compiler the freedom to delete it only at the end of the function?

tables/Tables/TableUtil.cc
```diff
@@ -0,0 +1,321 @@
...
+        // No subtable given, so create a main table.
+        SetupNewTable newtab(tableName, desc,
+                             tableName.empty() ? Table::Scratch : tabOpt,
+                             storageOption);
+        newtab.bindCreate (dmInfo);
+        // Create the table.
+        return Table(newtab, tabType, lockOptions, nrrow, initialize, endian, tsmOpt);
+      }
+      return createSubTable (t.first, t.second, desc, tabOpt, storageOption,
+                             dmInfo, lockOptions, nrrow, initialize, endian, tsmOpt);
```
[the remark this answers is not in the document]
gervandiepen: In the past compilers used to complain about 'no return' if put in an else branch. Probably they are smarter now.
But what is the advantage?

tables/Tables/TableUtil.cc
```diff
@@ -0,0 +1,321 @@
...
+                     Table::TableOption option,
+                     const TSMOption& tsmOption)
+    {
+      // See if the table can be opened as such.
+      if (Table::isReadable(tableName)) {
+        return Table(tableName, lockOptions, option, tsmOption);
+      }
+      // Try to find and open the last subtable by splitting at ::
+      // It throws an exception if not possible.
+      std::pair<Table,String> t = tryFindSubTable (tableName, True);
```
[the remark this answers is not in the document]
gervandiepen: Fixed by tryFindSubTable changes.

tables/Tables/TableUtil.cc
```diff
@@ -0,0 +1,321 @@
...
+//# Correspondence concerning AIPS++ should be addressed as follows:
+//#        Internet email: aips2-request@nrao.edu.
+//#        Postal address: AIPS++ Project Office
+//#                        National Radio Astronomy Observatory
+//#                        520 Edgemont Road
+//#                        Charlottesville, VA 22903-2475 USA
+//#
+//# $Id$
+
+//# Includes
```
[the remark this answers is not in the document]
gervandiepen: $Id$ is part of standard casacore header, although it is old cvs/svn functionality.

tables/Tables/TableUtil.cc
```diff
@@ -0,0 +1,321 @@
...
+      } else if (names.size() > 1  ||  mustExist) {
+        // No need to check if new table and no subtable given.
+        // Check if main table exists.
+        if (! Table::isReadable (names[0])) {
+          msg = "main table " + names[0] + " does not exist";
+        } else {
+          tab = Table(names[0]);
+          // Get name of last subtable.
+          if (names.size() > 1) {
+            lastPart = names[names.size()-1];
```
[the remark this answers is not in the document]
gervandiepen: Changed

tables/Tables/TableUtil.cc
gervandiepen: Changed name to findParentTable and removed Bool argument.
openTable now does the last open, which also solves problem that LockOption was not used.

tables/Tables/test/tTableUtil.cc
gervandiepen: Done

tables/Tables/test/tTableUtil.cc
gervandiepen: Casacore forbids imports of class, function or namespace in header files, so anybody can safely uses them. But imports are allowed it in .cc files because that scope is limited.
I've changed the include into a normal using statement.

tables/Tables/TableUtil.cc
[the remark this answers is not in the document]
gervandiepen: Undid the change, because it changed the behaviour in case names.size()==1.

tables/Tables/TableUtil.cc
```diff
@@ -0,0 +1,321 @@
...
+      return tabName;
+    }
+    
+    // Return Table and name of last part.
+    std::pair<Table,String> tryFindSubTable (const String& fullName, Bool mustExist)
+    {
+      Table tab;
+      String lastPart, msg;
+      // Split the name on :: to get the main and subtable names.
+      Vector<String> names = stringToVector(fullName, std::regex("::"));
```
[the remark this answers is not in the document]
gervandiepen: changed

tables/Tables/TableUtil.cc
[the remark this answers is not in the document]
mnijhuis-tos: Using `else` makes the code more explicit and a therefore a little more readable. I think compilers shouldn't complain about missing returns anymore in such cases.

However, I said this improvement is 'Optional', so feel free to ignore it.

tables/Tables/Table.h
```diff
@@ -1003,33 +1003,33 @@ friend class TableExprNodeRep;
     // Some deprecated functions for backward compatibility, now in TableUtil.h.
     // Use old way of indicating deprecate to avoid -Wc++14-extensions warnings.
     // <group>
-     // [[deprecated ("Now use TableUtil::openTable")]]
+     //# [[deprecated ("Now use TableUtil::openTable")]]
```
mnijhuis-tos, line 1006: Why did you add the # ? I don't mind, but I couldn't find any documentation what # does here.

tables/Tables/TableUtil.cc
```diff
@@ -57,9 +57,16 @@ namespace casacore {
...
+      std::pair<Table,String> t(findParentTable(tableName));
+      if (t.first.isNull()) {
+        throw TableError ("Table " + tableName + " does not exist");
+      }
+      // Find the last subtable in the parent.
+      if (! t.first.keywordSet().isDefined(t.second)) {
+        throw TableError ("Table name " + tableName + " is invalid (" +
+                          "subtable " + t.second + " is unknown)");
+      }
+      return t.first.keywordSet().asTable (t.second);
```
mnijhuis-tos, line 69: You still don't use `lockOptions`, `option` and `tsmOption` when opening a subtable.

tables/Tables/TableUtil.cc
```diff
@@ -186,8 +190,8 @@ namespace casacore {
       }
       // See if the name represents a table.
       if (! Table::isReadable(tableName)) {
-        // See if the name contains subtable names using ::
-        std::pair<Table,String> t = tryFindSubTable (tableName, False);
+        // See< if the name contains subtable names using ::
```
mnijhuis-tos, line 193: Remove '<'

tables/Tables/TableUtil.cc
```diff
@@ -186,8 +190,8 @@ namespace casacore {
...
+        std::pair<Table,String> t = findParentTable (tableName);
         if (! (t.first.isNull()  ||  t.second.empty())) {
```
mnijhuis-tos, line 195: Only the first part of this check should be enough:
```
if (!t.first.isNull()) {
```

tables/Tables/TableUtil.cc
```diff
@@ -137,28 +144,28 @@ namespace casacore {
                          Bool checkSubTables, Bool splitColons)
     {
       if (splitColons) {
-        std::pair<Table,String> t = tryFindSubTable (tableName, True);
+        std::pair<Table,String> t = findParentTable (tableName);
         if (! (t.first.isNull()  ||  t.second.empty())) {
```
mnijhuis-tos, line 148: Only checking `!t.first.isNull()` should be enough.

tables/Tables/TableUtil.cc
```diff
@@ -260,15 +264,15 @@ namespace casacore {
...
+      return BaseTable::tableInfo (getFullName(tableName));
+    }
 
     String getFullName (const String& tableName)
     {
       // See if a subtable is given using ::.
       String tabName;
-      std::pair<Table,String> t = tryFindSubTable (tableName, False);
+      std::pair<Table,String> t = findParentTable (tableName);
       if (t.first.isNull()  ||  t.second.empty()) {
```
mnijhuis-tos, line 276: `t.first.isNull()` by itself already says if a subtable is given.

tables/Tables/TableUtil.cc
```diff
@@ -288,26 +292,23 @@ namespace casacore {
...
-                tab = Table();
-                break;
-              }
-              tab = tab.keywordSet().asTable (names[i]);
+          lastPart = names[names.size()-1];
+          // Check if all subtables exist, except last one.
+          for (uInt i=1; i<names.size()-1; ++i) {
+            if (! tab.keywordSet().isDefined(names[i])) {
+              msg = "subtable " + names[i] + " is unknown";
+              tab = Table();
```
mnijhuis-tos, line 308: Optional: Resetting `tab` isn't strictly necessary, since you're throwing an exception below. If it is necessary, I would reset `tab` just before throwing the exception, so it's always reset in all error cases.

tables/Tables/test/tTableUtil.cc
mnijhuis-tos: For .cc files, Google also forbids importing an entire namespace. I don't know to what extend we want to follow Google style for Casacore, though.

tables/Tables/TableUtil.cc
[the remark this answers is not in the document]
mnijhuis-tos: I only meant the `//# Includes`

tables/Tables/TableUtil.cc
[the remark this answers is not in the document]
mnijhuis-tos: GitHub doesn't show any change...

mnijhuis-tos (changes requested):

tables/Tables/Table.h
gervandiepen: //# lines are not included in the doxygen documentation.
casacore used to have its own documentation processor, but switched to doxygen quite some time ago. A preprocessing script turns comment lines with casacore tags into doxygen lines.

tables/Tables/TableUtil.cc
gervandiepen: Your're right. Changed at all places.

tables/Tables/TableUtil.cc
gervandiepen: Added them to findParentTable.

tables/Tables/TableUtil.cc
gervandiepen: Done

mnijhuis-tos (approved):
Since there's only one minor unresolved conversation left, I'll approve.

PR #1158, opened 2022-02-09
Check open func map before using defaults in ImageOpener

pford:
This PR changes ImageOpener::openImage to check the function map before using the default open functions in casacore.

ImageOpener::openImage uses defined open functions for certain image types (including HDF5) before checking the function map, so there is no way to set a custom function for these types.  CARTA HDF5 images fail with casacore::HDF5Image; we have created a custom carta::CartaHdf5Image which we would like to open instead, using registerOpenImageFunction with type HDF5.

### DIFF
--- a/images/Images/ImageOpener.cc
+++ b/images/Images/ImageOpener.cc
@@ -253,6 +253,10 @@ LatticeBase* ImageOpener::openImage (const String& fileName,
      return 0;
    }
    ImageOpener::ImageTypes type = ImageOpener::imageType(fileName);
+   // override default open functions
+   if (theirOpenFuncMap.find(type) != theirOpenFuncMap.end()) {
+     return theirOpenFuncMap[type] (fileName, spec);
+   }
    // Do not require the registration of a PagedImage or HDF5Image openFunction.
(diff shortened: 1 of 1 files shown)


gervandiepen:
Does this mean that type HDF5 is always opened with CartaHDF5?
Maybe it's a good idea to put a try/catch around the new line 258 to follow the old code in case the open fails.

pford:
Good point, that way either type of HDF5 image will be opened.  This requires that the registered open function throws an AipsError rather than returning a nullptr.

gervandiepen:
Why so? I thought that the CartaHDF5 open function will be called for each HDF5 image file, but that only the HDF5 images matching the Carta format can be opened successfully. So CartaHDF5::open can throw an exception for the other cases which can be caught.

pford:
Yes, I meant throw an exception if not a CARTA HDF5 image file.

gervandiepen:
I think catching AipsError is too limited because a std::exception might be thrown in such a 'private' open function.
In fact you can do:
  catch (...) {
  }
to catch all exceptions.

PR #1159, opened 2022-02-10
Use 4-byte rownr for TiledDataStMan in older tables

### LINKED ISSUE #1156: Reading old measurement sets with casacore-3.4
Working on ASKAPsoft conversion into casacore-3.4 I accidentally came across an issue that sometimes an old measurement set could be read by casacore-3.2.1 and not by casacore-3.4. It definitely doesn't happen for all measurement sets, and from what I've seen so far, is more of an exception than the rule. I ran into a problem with the small test measurement set which we used for various unit tests (it was generated with NewMSSimulator-based code but very long time ago, more than 10 years ago, so it is possible that some aspect of this test MS is not quite right), but a few real ASKAP datasets I tried worked fine. However, it would be nice to investigate what exactly causes the problem just in case reading the measurement set could be made more robust. Below is the description how to reproduce the issue using just casacore:

1. Download test measurement set from [here](https://gitlab.com/ASKAPSDP/accessors/-/blob/master/testdataset.tar.bz2) -  or ask myself, if you have issues (I was unable to attach it to this issue for some reason, although it is only 2.3 Mb) and untar it.
2. Try showtableinfo in=testdataset.ms

it will work with casacore-3.2.1, but will raise an exception like this with casacore-3.4.0:

showtableinfo: Version 2016Feb17GvD
FilebufIO::read - incorrect number of bytes (0 out of 1114374144) read for file /...../testdataset.ms/table.f2

The actual place where this exception occurs is in the casacore::Table constructor. I also noticed that I cannot open this measurement set with casabrowser in casa 5.8.0, but it works in casa 5.6.2 which may be the same thing.

gervandiepen:
It appeared that the change to rownr_t was not done correctly for the seldomly used storage TiledDataStMan. The rownr was always read as 8 bytes, instead of 4 bytes for the older files. Now its version is increased to make the distinction; luckily the version in the internal header could be used to distinguish old and new. The row map has been changed to 8-byte rownrs too.
A test has been added using the old MS sent by Maxim.

Close #1156


tammojan (approved):
This looks good to me.

PR #1160, opened 2022-02-10
Allow offsets to JPL coordinates

tammojan:
This adds the possibility to point towards solar system offsets with an offset. For example, the MDirection `sun_offset(Quantity(3, "deg"), Quantity(2, "deg"), MDirection::SUN)` will point away from the sun center 3 degrees in the right ascension direction, 2 degrees in the declination direction.

Previously, the quantities given to any object in a `MDirection::SUN` or other solar system frame were ignored. With this change, we assume that previously for getting solar system coordinates, either the coordinates were omitted or set to zero.


aroffringa (approved):
Looks good to me! I'll overlook Google style issues since I don't think we apply that yet to casacore ;)

PR #1166, opened 2022-02-22
Use 64-bit rownrs in Dysco

### LINKED ISSUE #1139: Let Dysco support 64-bit rownrs
Currently Dysco is still using the old 32-bit rownr Tables interface. It should be changed to the new 64-bit interface.

gervandiepen:
The Dysco storage manager has been changed to use 64-bit row numbers.
Furthermore, it has been assured that the array storage is contiguous by using getStorage.

Close #1139

### DIFF
--- a/tables/Dysco/threadeddyscocolumn.cc
+++ b/tables/Dysco/threadeddyscocolumn.cc
@@ -105,21 +105,20 @@ void ThreadedDyscoColumn<DataType>::loadBlock(size_t blockIndex) {
 
 template <typename DataType>
 void ThreadedDyscoColumn<DataType>::getValues(
-    casacore::uInt rowNr, casacore::Array<DataType> *dataPtr) {
+    casacore::rownr_t rowNr, casacore::Array<DataType> *dataArr) {
+  // Make sure array storage is contiguous.
+  casacore::Bool deleteIt;
+  DataType* dataPtr = dataArr->getStorage (deleteIt);
   if (!areOffsetsInitialized()) {
     // Trying to read before first block was written -- return zero
     // TODO if a few rows were written of the first block, those are
     // incorrectly returned. This is a rare case but can be fixed.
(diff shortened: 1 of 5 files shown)


tables/Dysco/threadeddyscocolumn.cc
```diff
@@ -105,21 +105,20 @@ void ThreadedDyscoColumn<DataType>::loadBlock(size_t blockIndex) {
 
 template <typename DataType>
 void ThreadedDyscoColumn<DataType>::getValues(
-    casacore::uInt rowNr, casacore::Array<DataType> *dataPtr) {
+    casacore::rownr_t rowNr, casacore::Array<DataType> *dataArr) {
+  // Make sure array storage is contiguous.
+  casacore::Bool deleteIt;
+  DataType* dataPtr = dataArr->getStorage (deleteIt);
```
aroffringa, line 111: Sometime in the future, it would be better to do this using a RAII structure. There are all kind of exceptions possible in the code below, which would lead to leaks. Not really directly a concern now, so no changes required, but maybe for a future MR.

tables/Dysco/threadeddyscocolumn.cc
```diff
@@ -105,21 +105,20 @@ void ThreadedDyscoColumn<DataType>::loadBlock(size_t blockIndex) {
 
 template <typename DataType>
 void ThreadedDyscoColumn<DataType>::getValues(
-    casacore::uInt rowNr, casacore::Array<DataType> *dataPtr) {
+    casacore::rownr_t rowNr, casacore::Array<DataType> *dataArr) {
+  // Make sure array storage is contiguous.
+  casacore::Bool deleteIt;
```
aroffringa, line 110: Would it be possible to start using just `bool` for new code?

aroffringa (approved):
Looks fine to me!

PR #1168, opened 2022-02-24
Avoid possible TaQLNode memory leaks

gervandiepen:
TaQLNode and associated classes were using raw pointers at a few places. It has been replaced by unique_ptr. Furthermore, restore now returns a TaQLNode object instead of a pointer.

### DIFF
--- a/tables/TaQL/TaQLNodeDer.cc
+++ b/tables/TaQL/TaQLNodeDer.cc
@@ -1667,7 +1667,7 @@ void TaQLRecFldNodeRep::show (std::ostream& os) const
     os << itsFromName;
   } else if (itsValues.isValid()) {
     if (itsValues.nodeType() == TaQLNode_Multi  &&
-        dynamic_cast<const TaQLMultiNodeRep*>(itsValues.getRep())->itsNodes.empty()) {
+        dynamic_cast<const TaQLMultiNodeRep&>(*(itsValues.getRep())).itsNodes.empty()) {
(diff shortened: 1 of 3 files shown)


tables/TaQL/TaQLNodeDer.cc
```diff
@@ -1667,7 +1667,7 @@ void TaQLRecFldNodeRep::show (std::ostream& os) const
     os << itsFromName;
   } else if (itsValues.isValid()) {
     if (itsValues.nodeType() == TaQLNode_Multi  &&
-        ((const TaQLMultiNodeRep*)(itsValues.getRep()))->itsNodes.empty()) {
+        dynamic_cast<const TaQLMultiNodeRep*>(itsValues.getRep())->itsNodes.empty()) {
```
mordante, line 1670:
```
dynamic_cast<const TaQLMultiNodeRep&>(*itsValues.getRep()).itsNodes.empty()) {
```
This way when the cast fails it thows an exception instead of having undefined behaviour due to deferring a `nullptr`.

mordante (commented):
LGTM, one minor suggestion.

tables/TaQL/TaQLNodeDer.cc
gervandiepen: When failing, the dynamic cast of a pointer returns a null pointer and does not throw. It will result in a core dump. A reference will throw though, so I'll use that.

PR #1169, opened 2022-02-24
Implement COPY COLUMN, LIKE and DROP COLUMN

gervandiepen:
This is the final stage of the new commands added to TaQL. These are:
- COPY COLUMN
- LIKE in the CREATE TABLE with a possible DROP COLUMN clause
- LIKE in the column specification
- Full support of :: in all commands to denote subtables; also for the creation of subtables.


tammojan:
For reference: I reproduced the `Syntax error: Bad fd number` on an ubuntu18.04/clang build. It's in the docker image `casacore_1804_branch_issue1143-final` on das5, make sure to uncomment lines 45 and 46 in /code/tables/TaQL/test/tTableGramCretab.run

tables/TaQL/TableParse.cc
```diff
@@ -67,8 +68,9 @@
 #include <casacore/casa/IO/AipsIO.h>
 #include <casacore/casa/OS/Timer.h>
 #include <casacore/casa/ostream.h>
+#include <algorithm>
 
-#include <casacore/casa/Containers/BlockIO.h>
+//#include <casacore/casa/Containers/BlockIO.h>
```
mordante, line 73: Please remove.

tables/TaQL/TableParse.cc
```diff
@@ -79,19 +81,26 @@ TableParse::TableParse()
...
-  : shorthand_p (shorthand),
+TableParse::TableParse (const Table& table, Int tabnr, const String& name,
+                          const String& shorthand)
+  : tabnr_p     (tabnr),
+    name_p      (name),
+    shorthand_p (shorthand),
     table_p     (table)
 {}
 
 TableParse::TableParse (const TableParse& that)
```
mordante, line 92: Why not use the default generated compiler function?

tables/TaQL/TableParse.cc
```diff
@@ -79,19 +81,26 @@ TableParse::TableParse()
...
-  : shorthand_p (that.shorthand_p),
+  : tabnr_p     (that.tabnr_p),
+    name_p      (that.name_p),
+    shorthand_p (that.shorthand_p),
     table_p     (that.table_p)
 {}
 
 TableParse& TableParse::operator= (const TableParse& that)
```
mordante, line 99: Why not use the default generated compiler function?

tables/TaQL/TableParse.cc
```diff
@@ -242,155 +252,162 @@ TableParseSelect::~TableParseSelect()
...
-      if (!foundSH  &&  alwaysOpen) {
-        table = TableUtil::openTable(name);
+    }
+  }
+  // Okay; we have found the first table.
+  AlwaysAssert (!table.isNull(), AipsError);
+  // Now process all parts in all subtable names, where the first name or
+  // first part might need to be skipped because already processed.
+  const TableRecord* keywords = &(table.keywordSet());
+  for (uInt k=stSub; k<subs.size(); ++k) {
```
mordante, line 359: Is there a reason to use `Int` and `uInt` in new code?

tables/TaQL/TableParse.cc
```diff
@@ -242,155 +252,162 @@ TableParseSelect::~TableParseSelect()
...
-                                 const Table& ftab,
-                                 const String& shorthand,
-                                 Bool addToFromList,
-                                 const vector<const Table*> tempTables,
-                                 const vector<TableParseSelect*>& stack)
+Table TableParseSelect::addTable (Int tabnr, const String& name,
+                                  const Table& ftab,
+                                  const String& shorthand,
+                                  Bool addToFromList,
+                                  const vector<const Table*>& tempTables,
```
mordante, line 259: Is this a `std::vector`? If so I prefer to name it `std::vector` instead of `vector`. Knowing it's a `std::vector` makes the code easier to understand.

tables/TaQL/TableParse.h
```diff
@@ -537,23 +568,22 @@ class TableParseSelect
...
-                 const Table& table,
-                 const String& shorthand,
-                 Bool addToFromList,
-                 const vector<const Table*> tempTables,
-                 const vector<TableParseSelect*>& stack);
+  Table addTable (Int tabnr, const String& name,
+                  const Table& table,
+                  const String& shorthand,
+                  Bool addToFromList,
+                  const vector<const Table*>& tempTables,
```
mordante, line 575: Nice change making it a reference.

tables/TaQL/TableParse.h
```diff
@@ -956,14 +998,26 @@ class TableParseSelect
 
 //# Implement the inline functions.
 inline Bool TableParse::test (const String& str) const
-  { return (shorthand_p == str  ?  True : False); }
+  { return (str.empty()  ||  shorthand_p == str); }
```
mordante, line 1001: This change in semantics is intended?

mordante (commented):
I only have some remarks. I think it would be good if somebody more familiar with the code does a more indepth review.

tables/TaQL/TableParse.cc
gervandiepen: Removed

tables/TaQL/TableParse.cc
gervandiepen: Removed

tables/TaQL/TableParse.cc
gervandiepen: Removed

tables/TaQL/TableParse.cc
gervandiepen: Not really. Those are the casacore types which I'm pretty used to.

tables/TaQL/TableParse.h
gervandiepen: Good to hear.

gervandiepen:
I'm afraid nobody is sufficiently familiar with this code to perform an indepth review. To do so, somebody has to spend some days (or weeks?) to learn all aspects of TaQL and the code. I guess such time is not available.

tammojan (approved):
I've looked through the tests, and from a user perspective I love the changes! It will be very useful to do e.g. `alter table bla.MS copy column FLAG_BACKUP=FLAG`.
Also `create table empty.MS like existing.MS` can be very useful.

PR #1170, opened 2022-02-28
Revert "Add -std=legacy compiler option for scimath_f"

tammojan:
This reverts commit c32cf89d81e697b64b89d8e9248d63419c1e8da1 as requested by @aroffringa in #1134.

### DIFF
--- a/scimath_f/CMakeLists.txt
+++ b/scimath_f/CMakeLists.txt
@@ -37,7 +37,7 @@ if (BUILD_FFTPACK_DEPRECATED)
 endif ()
 
 add_library (casa_scimath_f ${buildfiles})
-target_compile_options (casa_scimath_f PRIVATE -std=legacy)
+
 target_link_libraries (casa_scimath_f casa_tables ${LAPACK_LIBRARIES} ${BLAS_LIBRARIES} ${CASACORE_ARCH_LIBS})
 
 install (


aroffringa (approved):

cquike:
I would like to reconsider this change. The CASA code still depends on FFTPack. There are simply not enough resources to test whether a change to FFTW would be acceptable or not, and therefore CASA still needs to unfortunately compile casacore with FFTPack support.

I understand that the option -fallow-argument-mismatch is strongly discouraged (https://gcc.gnu.org/onlinedocs/gfortran/Fortran-Dialect-Options.html) but I seem not to see a lot of options here. The reality is that users who need to compile casacore/casa in a modern compiler do need to provide that option. So in practical terms, even if it is not included in the casacore repository, code is compiled with the flag.

Of course other option is to fix the FFTPack code, if someone can contribute a patch for it...

aroffringa:
I think that keeping FFTPack around will in the long run require more resources than dropping support and testing that now. To me, what you're basically saying is that there are not sufficient resources to maintain CASA. I can't really help with that I'm afraid, but I think it shouldn't hold back evolution of the casacore library, which is also widely used outside of Casa... But on a wider note, maybe we (people involved in Casacore) should at some point come together and discuss the approach to casacore development and its future.

cquike:
CASA codebase is very large and some parts of it (like the FFTPack dependency) are not considered high priority. I agree that this builds technical debt, but that doesn't deem CASA an unmaintained project. That's not what I meant.
Indeed, having a discussion on the future of casacore development would be a good thing to do.

tammojan:
I think (as suggested by @gervandiepen just now) that an option `-fallow-argument-mismatch` is warranted if the option `-DBUILD_FFTPACK_DEPRECATED` is set also. Upgrading to CMake 3.3 for this (as it has [CheckFortranCompilerFlag](https://cmake.org/cmake/help/latest/module/CheckFortranCompilerFlag.html)) would be fine by me. @cquike If you agree I could make a pull request with this idea.

cquike:
Thank you for helping help on this @tammojan . Yes, I fully agree with that approach, so please go ahead with the PR.

PR #1172, opened 2022-03-01
Remove custom type system

aroffringa:
See #1100 .


casa/Containers/RecordInterface.cc
```diff
@@ -703,16 +702,9 @@ RecordNotice::RecordNotice (NoticeType changeType, uInt fieldNumber)
...
-    // This function returns the "Notice" type, retrieved
-    // from the "type registry".
-    return Register(this);
-}
-
-int RecordNotice::operator== (const Notice& that) const
-{
-    if (type() != that.type()) {
+    if (typeid(*this) != typeid(that)) {
 	return 0;
```
gervandiepen, line 708: Maybe more logical to redturn False

measures/Measures/MeasIERS.cc
```diff
@@ -41,6 +41,8 @@
 #include <casacore/casa/BasicSL/String.h>
 #include <casacore/casa/System/AppState.h>
 
+#include <atomic>
```
gervandiepen, line 44: Why is this include necessary?

measures/Measures/MeasTable.cc
```diff
@@ -53,6 +53,8 @@
 #include <casacore/casa/System/AipsrcValue.h>
 #include <casacore/casa/Utilities/Assert.h>
 
+#include <atomic>
```
gervandiepen, line 56: Why is this include necessary?

ms/MSOper/MSSummary.cc
```diff
@@ -821,7 +821,7 @@ void MSSummary::listAntenna (LogIO& os, Bool verbose) const
         os << "z";
         os << endl;
         vector<MPosition> antPos = _msmd->getAntennaPositions();
-        Bool posIsITRF = antPos[0].type() != MPosition::ITRF;
+        Bool posIsITRF = antPos[0].getRefPtr()->getType() != MPosition::ITRF;
```
gervandiepen, line 824: Why not antPos[0].getRef().getType() as done in MSMetaData.cc above?

ms/MSSel/MSSpWindowIndex.cc
```diff
@@ -142,7 +142,7 @@ Vector<Int> MSSpWindowIndex::matchFreq(const Vector<MFrequency>& chanFreq,
     if (freqMatch(row)) {
       for (uInt chan=0; chan<nChan; chan++) {
 	freqMatch(row) = (freqMatch(row) &&
-			  chanFreq(chan).type() == rowChanFreq(chan).type() &&
+			  chanFreq(chan).getRefPtr()->getType() == rowChanFreq(chan).getRefPtr()->getType() &&
```
gervandiepen, line 145: Same remark as above; also for several files hereafter.

gervandiepen:
In #1100 you comment that getRefPtr()->getType() returns an enum, but isn't it returning a uInt (from MRBase or MeasRef)?

measures/Measures/MeasIERS.cc
aroffringa: A std::atomic is used in line 100.

measures/Measures/MeasTable.cc
aroffringa: Also because of use of std::atomic

ms/MSOper/MSSummary.cc
aroffringa: It's not clear to me when to use which one. I've changed it now.

aroffringa:
> In #1100 you comment that getRefPtr()->getType() returns an enum, but isn't it returning a uInt (from MRBase or MeasRef)?

It's slightly confusing, it does return enum values like ITRF, but you're right that technically it returns an uInt. Not sure why.

gervandiepen (approved):

PR #1173, opened 2022-03-01
Improve documentation

gervandiepen:
Various doxygen warnings are fixed to improve the documentation.
Furthermore, some user messages are extended to be more clear.

### DIFF
--- a/tables/Dysco/threadeddyscocolumn.h
+++ b/tables/Dysco/threadeddyscocolumn.h
@@ -50,7 +50,7 @@ class ThreadedDyscoColumn : public DyscoStManColumn {
   virtual void setShapeColumn(const casacore::IPosition &shape) override;
 
   /** Get the dimensions of the values in a particular row.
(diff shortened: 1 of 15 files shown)


tables/Dysco/threadeddyscocolumn.h
```diff
@@ -50,7 +50,7 @@ class ThreadedDyscoColumn : public DyscoStManColumn {
   virtual void setShapeColumn(const casacore::IPosition &shape) override;
 
   /** Get the dimensions of the values in a particular row.
-   * @param rownr The row to get the shape for. */
+   * The rownr parameter is not used as the shape is the same for all rows. */
```
tammojan, line 53: If I were a doxygen purist, I'd argue that the `@param` markup should be kept here, something like
```
* @param rownr This parameter is ignored as the shape is the same for all rows.
```
But feel free to ignore this.

tammojan (approved):
Nice changes, thanks!

PR #1175, opened 2022-03-18
CAS-13749 retrieve values from ALMA specific adhoc SPECTRAL_WINDOW::SDM_CORR_BIT column

dmehring:


ms/MSOper/MSMetaData.h
```diff
@@ -707,6 +709,8 @@ class MSMetaData {
         QVD effbw;
         // RESOLUTION
         QVD resolution;
+        // CORR_BIT, CAS-13749
```
tammojan, line 712: Could you document briefly in the code what corr_bit is or does?

tammojan (commented):
If `SPECTRAL_WINDOW::SDM_CORR_BIT` is ALMA specific, I'm not sure the casacore-wide MS is the right place to retrieve this data. Could you elaborate a bit on why you implement it here?

ms/MSOper/MSMetaData.h
dmehring: Hi @tammojan 
Because 
1. it is measurement set metadata and so the MSMetaData class is the obvious place to implement it
2. it fits naturally into the spectral window specific struct SpwProperties that MSMetaData already defines and has been expanded in the past for new fields like this
3. there is already telescope-specific code in MSMetaData, eg SpwProperties.bbcno, so the precedent was long ago set, and there are many other telescope-specific bits of code spread throughout the casacore codebase that pre-date MSMetaData, 
4. There is negligible overhead if this particular column doesn't exist in the MS, so the code does not negatively impact performance for measurement sets created from other telescopes.
5. if I had created a whole MS column infrastructure around this column in casacore/ms/MeasurementSet/, I could understand objections since this column is not part of the MS spec. That is why I did not and limited the supporting implementation to just a handful of lines of code in the class in which it belongs. 
6. implementing it here takes advantage of supporting code that already exists, thus taking advantage of code reuse instead of implementing the same supporting infrastructure elsewhere.

tammojan: Thanks @dmehring. I agree the overhead is small here, and performance is not impacted. Could you, next to `CAS-13749`, add one line in the code comment that gives a hint on what `CORR_BIT` is, and that it is ALMA specific?

dmehring: Code comments added per request

dmehring: @tammojan can this be merged now?
Thanks

tammojan: Yes, sorry, it slipped my attention. Thanks for the reminder.

PR #1176, opened 2022-03-23
clean up things that were missed in PR #1172

dmehring:


casa/Utilities/Sequence2.cc
```diff
@@ -30,7 +30,7 @@
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
 #if defined(USE_THREADS)
-std::atomic<uInt> uIntSequence::next(1); // start at 1 to stay in sync with RegSequence
+std::atomic<uInt> uIntSequence::next(1); // start at 1 to stay in sync with RegSequence, FIXME fix comment, RegSequnce no longer exists
```
tammojan, line 33: @aroffringa or @gervandiepen Could you comment on what this comment should be? I have no idea why this `next` should start at 1.

tammojan (changes requested):
Thanks! Interesting how this got through the tests. Apparently this header is not included in any test.
Let's wait a little bit with merging until Ger or André can comment on the `next=1` comment.

casa/Utilities/Sequence2.cc
dmehring: Hi @tammojan . If we are still waiting on the resolution to the comment in Sequence2.cc, then could we please at least merge the other file at this point? I personally don't care about the comment, so leaving Sequence2.cc as is on master HEAD is fine with me. I need the code changes in the .h file so our CASA code will actually compile.
Thanks

tammojan: I'm actually fine with leaving this fixme here.

PR #1177, opened 2022-03-29
Move back to maia.usno.navy.mil for measures data

tammojan:
Work done by Arno Schoenmakers


dpetry:
@tammojan , this is only temporary, right? Once the IERS services are up again, this will be reverted, correct?

tammojan:
I don't know why IERS is down (and why the maia server is up!), if IERS comes back we could switch back. It would be good to keep the change that allows https in that case.

dpetry:
IERS is down because of maintenance. MAIA was down because of a website upgrade which is completed now.
Ideally, our system should try to use MAIA if IERS is down.

dpetry:
If the fall-back feature could be implemented now, that would be good because now we have a moment where we can test it. I suspect IERS is going to be up again soon.

dpetry:
However, looking at the code, I don't see a simple way of implementing a fall-back. Do you?

tammojan:
No, I think implementing a fall-back in this code will be hard.

tammojan:
Thanks for the review + merge @dpetry! Let's move back to iers.org once the maia server gives problems again..

PR #1178, opened 2022-03-30
WCExtension sort extension/stretch axes only if nonzero size

pford:
WCExtension uses the construct &x[0], where x can be a std::vector of length 0. The behavior in that case is undefined, and the _GLIBCXX_ASSERTIONS compile flag causes an Abort and core dump at runtime.  CARTA creates a carta-casacore RPM containing CASA imageanalysis and casacore, and we think rpmbuild sets this compile flag.

WCExtension::doToLCRegion sets the stretch and extension axes in std::vector, which may be length 0 if these axes are not required.  But then the empty vector is sorted with GenSortIndirect, causing a core dump when the compile flag is used:

GenSortIndirect<Int,uInt>::sort (strinx, &(tmpstr[0]), ndstr);

/usr/include/c++/8/bits/stl_vector.h:932: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = int; _Alloc = std::allocator<int>; std::vector<_Tp, _Alloc>::reference = int&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__builtin_expect(__n < this->size(), true)' failed.
Aborted (core dumped)

The fix is simply to check if the number of axes (ndext and ndstr) > 0 before sorting them.

Also fixed inconsistent format from tabs in the code.

### DIFF
--- a/images/Regions/WCExtension.cc
+++ b/images/Regions/WCExtension.cc
@@ -144,18 +144,18 @@ LCRegion* WCExtension::doToLCRegion (const CoordinateSystem& cSys,
     // In our axesDesc the first axes are used for the region.
     for (uInt i=0; i<ndreg; i++) {
         regPixMap(i) = pixelAxesMap(i);
-	regOutOrd(i) = outOrder(i);
+        regOutOrd(i) = outOrder(i);
     }
     // The rest of the pixel/outOrder are for the extend box.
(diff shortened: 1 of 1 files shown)


tammojan:
Thanks!

PR #1180, opened 2022-04-04
Remove dependency on kernsuite image

mnijhuis-tos:
Also combine docker RUN commands and use all cores for make.

### DIFF
--- a/docker/ubuntu1804_clang.docker
+++ b/docker/ubuntu1804_clang.docker
@@ -1,6 +1,7 @@
-FROM kernsuite/base:6
-RUN docker-apt-upgrade
-RUN docker-apt-install \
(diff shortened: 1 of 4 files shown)


mnijhuis-tos:
With this change, casacore no longer depends on the kernsuite image. It was only used for some handy ubuntu packaging shortcuts and for the casacore-data package. The Linux builds now manually download WSRT_Measures.ztar from Astron, similarly to the OSX build. Please let me know if we should continue depending on kernsuite or not.

tammojan (approved):
I agree that the dependency of CI on kernsuite just for casacore-data and the convenience wrappers is a bit of an overkill. This will also make it easier to create a CI job for ubuntu 22.

PR #1181, opened 2022-04-05
Array: Fix static member not being defined.

RemiLacroix-IDRIS:
Compilation was broken with Intel compilers as "uninitialized" was declared but never defined.

### DIFF
--- a/casa/Arrays/Array.tcc
+++ b/casa/Arrays/Array.tcc
@@ -1191,6 +1191,10 @@ void Array<T, Alloc>::tovector(std::vector<T, U> &out) const {
   const_cast<Array<T, Alloc>*>(this)->freeStorage(stor, deleteIt);
 }
 
+// Define static member
+template<typename T, typename Alloc>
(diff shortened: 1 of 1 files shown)


RemiLacroix-IDRIS:
I have checked that this is done correctly everywhere else in the code, for example:
https://github.com/casacore/casacore/blob/318fbcd0abf0abb077e1ca34c3e658a6e943fdf0/casa/Utilities/Template.tcc#L46-L49

The GNU compilers do not seem to complain about this either because they are less strict or because in this specific case, the variable is only used as a tag and might be optimized out completely.

RemiLacroix-IDRIS:
@tammojan: Sorry to bother you but do you know who should I ping for review?

aroffringa (approved):

tammojan:
Thanks @RemiLacroix-IDRIS for the improvement, and thanks @aroffringa for the review. I'll merge when the pipeline succeeds.

RemiLacroix-IDRIS:
You're welcome, thanks to @aroffringa for reviewing.

PR #1183, opened 2022-04-08
Expose beam areas to API

dmehring:


dmehring:
I'm going to add a couple of related methods, so I'll submit them all at once.

Issue #896, opened 2019-04-29
Remove obsolete features and replace by C++11 where possible


gervandiepen:

gervandiepen:
Issue #118 has been fixed by PR #890.
However, there are a few remaining issues:
- Mutex could be replaced by std::mutex.
- Notify and Link can be removed once the Record classes do not support notification anymore. AFAIK notification has never been used (apart from test programs), so can be removed.
- Bool, True and False could be replaced by bool, true and false. However, there is no real need to do so as they are #defined as such.
- Similarly Double and Float could be replaced by double and float; Int and uInt by int32_t and uint32_t.
But again, no real need to do so.
- As discussed in #108 replacing String by std::string is difficult since String has quite some additional functionality.
- As discussed in PR #881 replacing Regex by std:regex is already done. However, it cannot be merged as long as gcc-4.8 is used anymore.

PR #1185, opened 2022-04-13
Remove notice and link

aroffringa:
According to @gervandiepen (in https://github.com/casacore/casacore/issues/896) this functionality is unused.
This PR changes the behaviour of RecordFieldPtr. Some tests in
Record and RecordField explicitly check for this behaviour, and therefore
fail currently, but all other tests still succeed. It should be checked if this is ok
before merging.

This is #1096, but on the casacore repos instead of my own.


gervandiepen:
I've looked briefly at the changes. Next week I'll do a more detailed review.
One question. Has it been checked with the CASA group that BUILD_DEPRECATED (and the old source files) can be removed?

aroffringa:
It hasn't -- who to ask? Maybe @cquike ?

gervandiepen:
Probably he is a good candidate as he is contributing most from the CASA group.

tammojan:
@cquike If you don't object, I'll merge this one in a few days.

cquike:
Hi @tammojan : yes, please go ahead. The CASA team is indeed not aware of any potential problem that this change might cause.
Thanks for checking!

gervandiepen:
No further review comments.

gervandiepen (approved):

PR #1187, opened 2022-04-14
Create vector of correct size

rtobar:
The code clearly assigns 4 values into the vector, but it only requests
space for 2. This was warned by g++ 11.2.


tammojan:
Thanks! This bug has been in there for 13 years (according to `git praise` it was introduced on 2009-04-09 23:12:39).

PR #1189, opened 2022-04-22
Make conversion from std::vector to casacore::Vector explicit

aroffringa:
Before this MR, it is possible to implicitly convert a std::vector to a casacore::Vector, for example:

std::vector<int> foo();

casacore::Vector<int> v = foo();

would be fine. Since this can be expensive, this should not be done implicitly. There are several bugs in DP3 because of this, and I also found one casacore typo-bug (in RFReaderWriter.cc) by making this conversion explicit.

It might require a few fixes in software that uses casacore, particularly casa (but the upside is that it might also highlight some bugs :) ).

The related DP3 MR is here: https://git.astron.nl/RD/DP3/-/merge_requests/661

### DIFF
--- a/ms/MSSel/MSArrayParse.h
+++ b/ms/MSSel/MSArrayParse.h
@@ -99,7 +99,7 @@ public:
   const TableExprNode *selectRangeGTAndLT(const Int& n0, const Int& n1);
   const TableExprNode *selectRangeGEAndLE(const Int& n0, const Int& n1);
   const TableExprNode *selectArrayIds(const Vector<Int>& arrayids);
-  inline const TableExprNode *selectArrayIds() {return selectArrayIds(parsedIDList_p);}
+  inline const TableExprNode *selectArrayIds() {return selectArrayIds(Vector<int>(parsedIDList_p));}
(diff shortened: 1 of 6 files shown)


ms/MSSel/MSArrayParse.h
```diff
@@ -99,7 +99,7 @@ class MSArrayParse : public MSParse
   const TableExprNode *selectRangeGTAndLT(const Int& n0, const Int& n1);
   const TableExprNode *selectRangeGEAndLE(const Int& n0, const Int& n1);
   const TableExprNode *selectArrayIds(const Vector<Int>& arrayids);
-  inline const TableExprNode *selectArrayIds() {return selectArrayIds(parsedIDList_p);}
+  inline const TableExprNode *selectArrayIds() {return selectArrayIds(Vector<int>(parsedIDList_p));}
```
gervandiepen, line 102: Should be Vector<Int> (with capital I)
Same in next 2 files.

aroffringa: Thanks, corrected.

gervandiepen (approved):

PR #1194, opened 2022-05-17
Fix missing initialization of MSIter (rebased)

tammojan:
See #1069


cquike:
Thank you @tammojan for rebasing my old #PR. Should have done that myself...

aroffringa:
In the other ticket, Ger van diepen mentioned "In general this change looks fine, but shouldn't the class member more_p be initialized as well?" -- I think that is not solved/answered yet.

mnijhuis-tos (approved):
These changes are a good improvement, however, looking at the origin() function, I suppose newDataDescId_p and more_p should also be true by default.

Further improvement (optional): Create a private constructor that initializes all members to their default values, and call this constructor from the existing constructors. This improvement removes the duplicate initialization code.

Issue #1171, opened 2022-03-01
TempImage throws std::bad_alloc under casacore-3.4 when handling degenerate dimensions


VoronkovMA:
I found this issue converting ASKAPsoft into casacore-3.4. Apparently, if an array with additional degenerate dimensions (w.r.t. the dimensions of the TempImage object) is passed to the TempImage::put method, the program  fails by either throwing a std::bad_alloc exception or because it got killed by the operating system. It is likely that it tries to allocate all available memory in a loop before the failure occurs. Interestingly, the same code works under casacore-3.2.1. 

Although passing an array with not exactly the same, but conforming, shape is a sloppy code (and the fix is trivial via Array::reform), it may be worth adding assert statements or otherwise check if the relevant code can be made more reliable, especially because there is a change in behaviour between casacore-3.4 and casacore-3.2.1. And moreover, the resulting bug is difficult to trace.

I attached some test code which can be used to reproduce the issue. It requires just casacore.   

[test.tar.gz](https://github.com/casacore/casacore/files/8158686/test.tar.gz)

tammojan:
Thanks for the clear issue report and reproducer. This change of behavior could have been introduced with the arrays modernization that @aroffringa did some time ago. Perhaps he can have a look?

```
#include <casacore/images/Images/TempImage.h>
#include <casacore/casa/Arrays/Array.h>
#include <casacore/casa/Arrays/Vector.h>
#include <casacore/casa/Arrays/Matrix.h>
#include <casacore/coordinates/Coordinates/CoordinateSystem.h>
#include <casacore/coordinates/Coordinates/LinearCoordinate.h>


#include <iostream>

casacore::CoordinateSystem getDummyCoordinateSystem()
{
   // this is just to create a dummy coordinate system required in the constructor
   casacore::Vector<casacore::String> names(2);
   names[0]="x"; names[1]="y";
   casacore::Vector<double> increment(2 ,1.);

   casacore::Matrix<double> xform(2,2,0.);
   xform.diagonal() = 1.;
   const casacore::LinearCoordinate linear(names, casacore::Vector<casacore::String>(2,"pixel"),
           casacore::Vector<double>(2,0.),increment, xform, casacore::Vector<double>(2,0.));

   casacore::CoordinateSystem coords;
   coords.addCoordinate(linear);
   return coords;
}


int main() {
   try {
      const casacore::IPosition shape(2,1024,1024);

      // this works fine
      //const casacore::IPosition arrayShape(shape);

[...]

PR #1196, opened 2022-05-30
Added assert to detect issue #1171

VoronkovMA:
A trivial change "fixing" issue #1171 - added assert to ensure the lattice has more dimensions than the source before adding degenerate axes. As discussed on the ticket, it is possible to support ldim < sdim case as well via reform method. However, mismatched dimensions in this particular operation is a sloppy code. So perhaps, it is better to detect this condition with an assertion rather than sweep under the carpet. 

I can easily implement the support of the ldim < sdim case, if others think it is worth doing.


aroffringa:
Looks like a good addition to me!

aroffringa (approved):

aroffringa:
Maybe wait until CI is finished next time, @tammojan ;)

PR #1198, opened 2022-06-11
Use overloading to detect strerror variant

### LINKED ISSUE #1197: Detection of strerror variant is not cross platform
Hey everyone! I'm working on packaging casacore into the Julia programming language's binary artifact system and am running into a few problems. By default, Julia's [BinaryBuilder](https://github.com/JuliaPackaging/BinaryBuilder.jl) builds for a pretty wide selection of combinations of gcc/libgfortran/c stdlib/etc. - so finding cross compilation bugs becomes [very easy](https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=20137&view=results).

One of these bugs that's I've run into is due to the detection of the stderr variant

https://github.com/casacore/casacore/blob/0d871e1fca1f96abd8bf1326fb45c253192d01c2/casa/Exceptions/Error2.cc#L221-L229

According to a [few sources](https://www.zverovich.net/2015/03/13/reliable-detection-of-strerror-variants.html), this reliance on the man page's preprocessor condition of `(_POSIX_C_SOURCE >= 200112L || _XOPEN_SOURCE >= 600) && ! _GNU_SOURCE` only works on glibc and some platforms.

This blog posts indicates a few solutions, both a CMake one and a function overloading one - I'm not sure which one we would prefer. In any case, this seems like an easy fix and would allow casacore to run on a multitude of systems.

kiranshila:
Fixes #1197

### DIFF
--- a/casa/Exceptions/Error2.cc
+++ b/casa/Exceptions/Error2.cc
@@ -213,19 +213,19 @@ SystemCallError::SystemCallError (int error, const String &msg,
 {}
 SystemCallError::~SystemCallError() noexcept
 {}
+String strerror_overload(int err)
+{
+  return "errno " + String::toString(err);
+}
+String strerror_overload(char *err)
+{
+  return err;
+}
 String SystemCallError::errorMessage(int error)
 {
   // Use strerror_r for thread-safety.
   char buffer[128];
-  // There are two incompatible versions of versions of strerror_r()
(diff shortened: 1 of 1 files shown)


tammojan:
Thanks for fixing this!

This looks good to me, and since it's only a few lines I'd prefer it to a CMake-based mechanism. If no one else objects, I'll merge in a few days.

kiranshila:
> This looks good to me, and since it's only a few lines I'd prefer it to a CMake-based mechanism.

That was my thought as well, overloading certainly seems elegant while everything in CMake is anything but :smiley:

PR #1199, opened 2022-06-13
Use shared_ptr instead of own reference counting of BaseTable

gervandiepen:
The Table object used its own (old) way of counting BaseTable references. It has been replaced by shared_ptr<BaseTable>.
However, it is somewhat tricky because not all BaseTable references in DataManager objects can be counted (otherwise mutual referencing occurs). Therefore the actual pointer in a Table object is still a raw pointer to BaseTable. A weak_ptr (in the BaseTable object itself) it used to created shared_ptr as needed.

The code has been run through valgrind without any error or leak.

### DIFF
--- a/tables/Tables/BaseTabIter.h
+++ b/tables/Tables/BaseTabIter.h
@@ -96,7 +96,8 @@ public:
     // iteration boundary. This improves performance in general but will
     // break existing applications that change the comparison objects
     // (cmpObjs) between iterations.
-    BaseTableIterator (BaseTable*, const Block<String>& columnNames,
+    BaseTableIterator (const std::shared_ptr<BaseTable>&,
+                       const Block<String>& columnNames,
                        const Block<CountedPtr<BaseCompare> >& cmpObjs,
                        const Block<Int>& orders,
                        int option,
@@ -111,7 +112,7 @@ public:
     virtual void reset();
 
     // Return the next group.
-    virtual BaseTable* next();
+    virtual std::shared_ptr<BaseTable> next();
 
     virtual void copyState(const BaseTableIterator &);
 
@@ -119,10 +120,11 @@ public:
     // comparison function) to terminate the most recent call to next()
     // Enables clients to sense iteration boundary properties
     // and organize associated iterations
-    inline const String& keyChangeAtLastNext() const { return keyChangeAtLastNext_p; };
+    inline const String& keyChangeAtLastNext() const
+      { return keyChangeAtLastNext_p; }
 
 protected:
-    BaseTable*             sortTab_p;     //# Table sorted in iteration order
+    std::shared_ptr<BaseTable> sortTab_p; //# Table sorted in iteration order
     rownr_t                lastRow_p;     //# last row used from reftab
     uInt                   nrkeys_p;      //# nr of columns in group
     String                 keyChangeAtLastNext_p;  //# name of column that terminated most recent next()
@@ -132,7 +134,7 @@ protected:
     // Copy constructor (to be used by clone)
     BaseTableIterator (const BaseTableIterator&);
 
-    BaseTable* noCachedIterBoundariesNext();
+    std::shared_ptr<BaseTable> noCachedIterBoundariesNext();
 
 private:
     // Assignment is not needed, because the assignment operator in
@@ -147,7 +149,8 @@ private:
     std::shared_ptr<Vector<size_t>> sortIterKeyIdxChange_p;
     Vector<rownr_t>::iterator sortIterBoundariesIt_p;
     Vector<size_t>::iterator  sortIterKeyIdxChangeIt_p;
-    RefTable* aRefTable_p;
+    std::shared_ptr<BaseTable> aBaseTable_p;
+    RefTable* aRefTable_p;      //# Same as aBaseTable_p
 };
 
 
--- a/tables/Tables/BaseTabIter.h
+++ b/tables/Tables/BaseTabIter.h
@@ -149,8 +149,8 @@ private:
     std::shared_ptr<Vector<size_t>> sortIterKeyIdxChange_p;
     Vector<rownr_t>::iterator sortIterBoundariesIt_p;
     Vector<size_t>::iterator  sortIterKeyIdxChangeIt_p;
-    std::shared_ptr<BaseTable> aBaseTable_p;
-    RefTable* aRefTable_p;      //# Same as aBaseTable_p
+    RefTable* aRefTable_p;      //# RefTable returned in each iteration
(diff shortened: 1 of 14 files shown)


tables/Tables/BaseTable.cc
```diff
@@ -745,32 +726,42 @@ BaseTable* BaseTable::doSort (PtrBlock<BaseColumn*>& sortCol,
...
-    return rtp;
+  std::shared_ptr<BaseTable> rtp (new RefTable(this, rowOrder, initialNrrow));
+  rtp->setWeakPtr (rtp);
+  return rtp;
 }
 
+RefTable* BaseTable::asRefTable()
+{
+  RefTable* rtp = dynamic_cast<RefTable*>(this);
+  AlwaysAssert (rtp, AipsError);
```
aroffringa, line 756: Why an AlwaysAssert? Generally programming errors are checked using a debug assert that is compiled out when doing a production compilation, whereas possible users (e.g. error in parset) are more strictly checked.

(I'm also not sure this function is too desirable. It might be better to dynamic_cast where necessary instead of adding an extra indirection for the programmer, that now has to look up what asRefTable does.)

tables/Tables/BaseTable.cc
```diff
@@ -779,16 +770,16 @@ BaseTable* BaseTable::select (rownr_t maxRow, rownr_t offset)
         maxRow = nrow() - offset;
     }
     if (offset == 0  &&  maxRow == nrow()) {
-        return this;
+      return getSharedPtr();    // pointer to itself
```
aroffringa, line 773: This comment can better be placed above the definition of `getSharedPtr();` and not repeated here.

tables/Tables/BaseTable.cc
```diff
@@ -980,46 +948,34 @@ BaseTable* BaseTable::tabXor (BaseTable* that)
...
-    if (allsw1) {
-	delete [] inx1;
-    }
-    if (allsw2) {
-	delete [] inx2;
-    }
-    return rtp;
+    //# Get both rownr arrays which are sorted if not in row order.
+    Vector<rownr_t> r1 = this->logicRows();
+    Vector<rownr_t> r2 = that->logicRows();
```
aroffringa, line 953: Nice!

tables/Tables/Table.cc
```diff
@@ -527,19 +508,19 @@ BaseTable* Table::makeBaseTable (const String& name, const String& type,
     ios >> format;
     ios >> tp;
     if (tp == "PlainTable") {
-	baseTabPtr = new PlainTable (ios, version, name, type, nrrow,
-				     tableOption, lockOptions, tsmOpt,
-                                     addToCache, locknr);
+      baseTabPtr.reset (new PlainTable (ios, version, name, type, nrrow,
+                                        tableOption, lockOptions, tsmOpt,
+                                        addToCache, locknr));
```
aroffringa, line 513: It is generally advised to always use `ptr = std::make_shared()` instead of `ptr.reset(new ...)`, because the second one costs one more allocation. This would then become `baseTabPtr = std::make_shared<PlainTable>(ios, version, ...)`. This also holds for the other cases below.

tables/Tables/BaseTabIter.h
```diff
@@ -147,7 +149,8 @@ class BaseTableIterator
     std::shared_ptr<Vector<size_t>> sortIterKeyIdxChange_p;
     Vector<rownr_t>::iterator sortIterBoundariesIt_p;
     Vector<size_t>::iterator  sortIterKeyIdxChangeIt_p;
-    RefTable* aRefTable_p;
+    std::shared_ptr<BaseTable> aBaseTable_p;
+    RefTable* aRefTable_p;      //# Same as aBaseTable_p
```
aroffringa, line 153: See my later/earlier comment about enable_shared_from_this, would that be possible here maybe too?

aroffringa (changes requested):
Great changes... that makes things more clear and easier to understand where ownership is. I have a few non-essential suggestions. It might be worth considering whether std::enable_shared_from_this can be used to further simplify things. If not I'm happy with as is.

tables/Tables/ConcatTable.h
```diff
@@ -303,8 +303,8 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
     Block<BaseColumn*> getRefColumns (const String& columnName);
 
     // Create a (temporary) Table object from it.
-    Table asTable()
-      { return Table (this, False); }
+    //Table asTable()
+    ///  { return Table (this, False); }
```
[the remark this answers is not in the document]
gervandiepen: Right.

tables/Tables/BaseTable.cc
gervandiepen: Removed asRefTable and used DebugAssert

tables/Tables/Table.h
```diff
@@ -1035,45 +1035,50 @@ friend class TableExprNodeRep;
...
-    //# Counter of last call to hasDataChanged.
+    // Shared pointer to count the references to the BaseTable object.
+    // The shared pointer can be null, so it is not counted which is necessary for
+    // the Table object in the DataManager. Otherwise mutual referencing would occur.
+    // Note that the BaseTable object contains a weak_ptr to itself which is the
+    // basis for all shared pointer counting.
+    std::shared_ptr<BaseTable> countedTabPtr_p;
+    // Pointer to BaseTable object which is always filled and always used.
+    // The shared_ptr above is only for reference counting.
+    BaseTable* baseTabPtr_p;
```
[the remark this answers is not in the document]
gervandiepen: Now using std::enable_shared_from_this instead opf own wek_ptr in BaseTable.h
However, the separate baseTabPtr_p is needed, because not all BaseTable instances can be counted due to mutual referencing as a Table object is used in DataManager.
So baseTabPtr_p is used in all functions, while countedTabPtr_p is only used for counting the references and for the final automatic deletion of BaseTable.

tables/Tables/BaseTable.cc
gervandiepen: Replaced getSharedPtr by shared_from_this

tables/Tables/Table.cc
gervandiepen: std::make_shared can only be used for concrete classes, not for abstract ones. Table is using std::shared_ptr<BaseTable>, thus for an abstract class. I realize that a std::shared_ptr<PlainTable> can be casted to a std::shared_ptr<BaseTable>, but for the few times a PlainTable object is allocated, this sounds like overkill to me.

gervandiepen:
Is the review complete and can the PR be merged?

aroffringa (approved):
Looks good to me!

PR #1201, opened 2022-06-21
  Fixed a bug where a 0 length vactor was being sent to min or max.

kgolap:
Fixed a bug where a 0 length vector may get addressed (in the min …
…max calls e.g min(chanIn)) in function matchFrequencyRange.  this happens (CAS-13843) when in the first pass the freqrange appear to be in a given spw min and max frequencies but on second pass the range falls in gaps between channels.  Fixed it by making the second pass check (numMatched can be 0) before trying to decide to address the vectors used.

Also removed a commented out piece of code in that function

### DIFF
--- a/ms/MSSel/MSSpwIndex.cc
+++ b/ms/MSSel/MSSpwIndex.cc
@@ -38,6 +38,7 @@
 #include <casacore/casa/Logging/LogIO.h>
 #include <algorithm>
 #include <vector>
+#include <iomanip>
 
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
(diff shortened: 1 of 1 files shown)


tammojan:
Looks good to me, thanks. I'll wait a little bit to give others the chance to comment.

mnijhuis-tos (changes requested):
The fix (adding a check for numMatched>0) looks good. Please verify that the second condition or the OR is not needed.

mnijhuis-tos (approved):
Thank you for the changes! LGTM!

Issue #82, opened 2015-04-03
Export cmake config on install


tammojan:
On install of casacore, cmake knows best what its configuration is (e.g. was it built with HDF5 support). This can be written out in a cmake config file, stored in ~/.cmake/packages/casacore (or in the registry on Windows). See http://www.cmake.org/Wiki/CMake/Tutorials/How_to_create_a_ProjectConfig.cmake_file and http://www.cmake.org/cmake/help/git-master/manual/cmake-packages.7.html#creating-packages

gijzelaerr:
Did you have a look at the `CMakeCache.txt` file? Sounds like what you describe.

gijzelaerr:
Do we still want this? I find it a bit ugly and there is no default location on the filesystem to put this.

rtobar:
+1 for this, the lack of this has led to many `FindCasacore.cmake` files scattered around projects. Regarding the destination in the filesystem, there are plenty of options, but I tend to go for something under `<prefix>/share/`.

Happy to contribute with a first version of this in a couple of weeks when I'm back to work.

tammojan:
@rtobar It would be great if you could contribute to this. Given the age of the request I think it can wait for a few weeks.

PR #1202, opened 2022-06-23
Add support for installation of pkg-config files

cquike:
Using pkg-config files (https://en.wikipedia.org/wiki/Pkg-config) is a convenient way for clients of casacore to get the compiler and linker flags needed to develop an application that uses casa. This pull request adds support for the creation and installation of a casacore.pc file that can be used for such purpose.

The need for this comes from casa, which traditionally has been keeping internal knowledge and how to compile against casacore. The idea here is to remove such knowledge from casa itself and leave it to the casacore installation to provide such information in a standard way.


cquike:
Hi @tammojan, thank you for looking into this. Indeed, #82 would be a nice addition too, they are not mutually exclusive.

tammojan (approved):

PR #1214, opened 2022-07-07
Update python docker images

tammojan:
These are the docker images that I just used for deploying python-casacore 3.5.

### DIFF
--- a/docker/py310_wheel.docker
+++ b/docker/py310_wheel.docker
@@ -44,8 +44,6 @@ RUN make -j${THREADS}
 WORKDIR /build
 RUN tar jxvf /tmp/boost.tar.bz2
 WORKDIR /build/boost_${BOOST_}
-# patch for python 3.10 (multiple digits)
-#sed -i 's/(\[0-9\])/([0-9][0-9]?)/' tools/boost_install/BoostConfig.cmake
 RUN ./bootstrap.sh --prefix=/opt/boost \
     --with-libraries=python \
     --with-python=/opt/python/${TARGET}/bin/python \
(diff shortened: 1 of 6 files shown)


docker/py310_wheel.docker
```diff
@@ -0,0 +1,79 @@
...
+WORKDIR /build/sofa/20150209_a/f77/src
+RUN make -j${THREADS}
+
+
+# setup boost
+WORKDIR /build
+RUN tar jxvf /tmp/boost.tar.bz2
+WORKDIR /build/boost_${BOOST_}
+# patch for python 3.10 (multiple digits)
+#sed -i 's/(\[0-9\])/([0-9][0-9]?)/' tools/boost_install/BoostConfig.cmake
```
mnijhuis-tos, line 48: This patch line is commented out. Do we still need it? Or does using BOOST 1.79 instead of 1.75 already address this issue?

docker/py310_wheel.docker
```diff
@@ -0,0 +1,79 @@
...
+    link=static,shared install
+
+# casacore wants numpy
+RUN /opt/python/${TARGET}/bin/pip install numpy==1.22.4
+
+# set up casacore
+ADD . /code
+RUN useradd -ms /bin/bash casacore
+RUN chown casacore.casacore /code
+RUN mkdir /code/build
```
mnijhuis-tos, line 65: The existing py3x_wheel.docker files have `USER casacore` above this line. Why did you remove it for python3.10?

mnijhuis-tos (changes requested):

docker/py310_wheel.docker
tammojan: I think indeed a boost upgrade did the trick. Building the wheels now, let's see.

docker/py310_wheel.docker
tammojan: I removed it because sometimes it's annoying to not be root. But for consistency I put the line back, thanks.

mnijhuis-tos (approved):

PR #1219, opened 2022-09-14
Add wcslib includes to the clients of casacore.

cquike:
The WCSLIB headers are included in public casacore headers, hence the flags to discover the WCSLIB library must also be propagated.

This is a small correction patch on previous PR #1202 .


tammojan (approved):

cquike:
Thanks for merging!

PR #1220, opened 2022-09-21
Remove Notice leftovers

FedeMPouzols:
When trying to compile CASA with an updated casacore, we noticed that there is a leftover include of Notice.h after the removal done in #1185, in a Utilities.h that is presumably only used from CASA.
This PR removes that include and fixes CASA compilation. It also removes a comment that refers to Notice.


aroffringa (approved):
Thanks, sorry for missing this include -- looks good to me!

PR #1221, opened 2022-09-23
Add AROUND/IN to TaQL and improve TableExprNodeSetElem

gervandiepen:
TaQL has been extended with AROUND/IN and <:> to specify an interval as mid-width. In this way it's easier to specify intervals for a MeasurementSet since its intervals are defined that way. If the width is zero, the interval is infinite which follows the MS definition.
It is not possible to follow the MS definition that a negative width means that a value is valid till the next row with the same keys, because it is not defined in the MS tables which columns form the keys.

To easily facilitate the new syntax, class TableExprNodeSetElem has been split into several classes, one for each element type. It also made it easier to optimize TaQL's 'expr IN set' command, which is now done using class TableExprNodeSetOpt and its derivations. Because templates are used in there, it was necessary to change the hasInt, etc. functions to the generic function name 'contains'.


tables/TaQL/ExprNodeSetOpt.cc
```diff
@@ -0,0 +1,450 @@
...
+  (const TableExprNodeSet& set,
+   const std::vector<T>& start, const std::vector<T>& end, 
+   const std::vector<Bool>& leftC, const std::vector<Bool>& rightC)
+   {
+    // See if all intervals have the same left and right closedness.
+    // If so, a better version can be used that does not need to test on it
+    // for each compare.
+    Bool same = True;
+    for (size_t i=1; i<leftC.size(); ++i) {
+      if (leftC[i] != leftC[0]  ||  rightC[i] != rightC[0]) {
```
mordante, line 285: There's no constraint that `leftC.size() == rightC.size()`

tables/TaQL/ExprNodeSetOpt.cc
```diff
@@ -0,0 +1,450 @@
...
+        } else {
+          optSet = new TableExprNodeSetOptContSetOO<T> (set, start, end);
+        }
+      }
+    } else {
+      optSet = new TableExprNodeSetOptContSet<T> (set, start, end,
+                                                  leftC, rightC);
+    }
+    TENShPtr ptr(optSet);
+    ///optSet->setRows (rowNrs);
```
mordante, line 311: Is this needed or not?

tables/TaQL/ExprNodeSetOpt.cc
```diff
@@ -0,0 +1,450 @@
...
+  {
+    TableExprNodeSetOptContSet<T>::show (os, indent);
+    os << " as TableExprNodeSetOptContSetCC" << endl;
+  }
+  
+  template<typename T>
+  Int64 TableExprNodeSetOptContSetCC<T>::find (T value) const
+  {
+    // Use lambda function to use <= instead of <.
+    auto iter = std::upper_bound (this->itsEnds.begin(), this->itsEnds.end(), value,
```
mordante, line 335: This does not explain why `<=` is the proper operation.
It also doesn't explain why `std::less_equal` can't be used.

tables/TaQL/ExprNodeSetOpt.cc
```diff
@@ -0,0 +1,450 @@
...
+
+  template<typename T>
+  void TableExprNodeSetOptContSetCC<T>::show (ostream& os, uInt indent) const
+  {
+    TableExprNodeSetOptContSet<T>::show (os, indent);
+    os << " as TableExprNodeSetOptContSetCC" << endl;
+  }
+  
+  template<typename T>
+  Int64 TableExprNodeSetOptContSetCC<T>::find (T value) const
```
mordante, line 332: These `TableExprNodeSetOptContSetXY<T>::Find` functions have subtle differences. I think it would be good to add a bit more comment in the code to point out the differences.

I even wonder whether it would be one generic function, that takes a comparison operator for the left and right side. Something like

```
template<class T, class ComparisonRight, class  ComparisonLeft>
 Int64 TableExprNodeSetOptContSetFind (T value,  ComparisonRight comparison_right, ComparisonLeft compison_left) {
  auto iter = std::upper_bound (this->itsEnds.begin(), this->itsEnds.end(), value, compison_left);
  
    if (iter != this->itsEnds.end()) {
      size_t index = std::distance (this->itsEnds.begin(), iter);
      if (comparison_right(value, this->itsStarts[index])) {
        return index;
      }
    }
    return -1;
 }
```
Obviously `itsStart` and `itsEnd` should be an argument too.

tables/TaQL/ExprNodeSetOpt.cc
```diff
@@ -0,0 +1,450 @@
...
+      if (value > this->itsStarts[index]) {
+        return index;
+      }
+    }
+    return -1;
+  }
+
+
+  // Instantiate for Double and String.
+  template class TableExprNodeSetOptUSet<Int64>;
```
mordante, line 437: This differs from the comment.
Can you explain why only these types are instantiated?

tables/TaQL/ExprNodeSetOpt.h
```diff
@@ -0,0 +1,358 @@
...
+  // </prerequisite>
+
+  // <synopsis>
+  // This class is a further optimized version of TableExprNodeSetOptContSet
+  // for continuous intervals all using a closed start and closed end.
+  // It reduces the number of comparisons required.
+  // </synopsis>
+
+  template <typename T>
+  class TableExprNodeSetOptContSetCC: public TableExprNodeSetOptContSet<T>
```
mordante, line 241: Why inherit from `TableExprNodeSetOptContSet` instead of `TableExprNodeSetOptContSetBase`.
It saves writing the `size()` function, but it stores two unused Bool vectors.

However that generic part could be move to the Base and make then `itsStarts` and `itsEnds` protected members.

tables/TaQL/test/tExprNodeSetOpt.cc
```diff
@@ -0,0 +1,361 @@
...
+
+// <summary>
+// Test program for the ExprNodeSetOpt classes.
+// </summary>
+
+
+
+void doDoubleContSet()
+{
+  std::vector<double> st ({1,3,5,7,9,11,13,15,17,19,21});
```
mordante, line 43: Please test with vectors with different sizes too.

tables/TaQL/test/tExprNodeSetOpt.cc
```diff
@@ -0,0 +1,361 @@
...
+    AlwaysAssertExit (set.contains (id, 13.));
+    AlwaysAssertExit (set.contains (id, 13.5));
+    AlwaysAssertExit (set.contains (id, 14.));
+    AlwaysAssertExit (! set.contains (id, 14.5));
+    AlwaysAssertExit (set.contains (id, 21.));
+    AlwaysAssertExit (set.contains (id, 21.5));
+    AlwaysAssertExit (set.contains (id, 22.));
+    AlwaysAssertExit (! set.contains (id, 22.5));
+    AlwaysAssertExit (set.find(22.5) == -1);
+    AlwaysAssertExit (set.find(13.5) == 6);
```
mordante, line 64: Please test the border conditions for find too.

tables/TaQL/test/tExprNodeSetOpt.cc
```diff
@@ -0,0 +1,361 @@
...
+    AlwaysAssertExit (! set.contains (id, 14.5));
+    AlwaysAssertExit (! set.contains (id, 21.));
+    AlwaysAssertExit (set.contains (id, 21.5));
+    AlwaysAssertExit (! set.contains (id, 22.));
+    AlwaysAssertExit (! set.contains (id, 22.5));
+  }
+  {
+    // Test intervals with mix of open and closed
+    std::vector<bool> leftC ({false,true,true,true,true,true,false,true,true,true,true});
+    std::vector<bool> rightC({false,true,true,true,true,true,true ,true,true,true,false});
```
mordante, line 123: I really would like to have test with less elements. The input data is quite far from this part of the input and with all true. false mixes here it's hard to verify the results in the test are correct.

tables/TaQL/test/tExprNodeSetOpt.cc
```diff
@@ -0,0 +1,361 @@
...
+                      false,false,true,true,false,true,true,false,false});
+    MArray<bool> res = set.contains (id, MArray<double>(vec));
+    AlwaysAssertExit (allEQ(res.array(), exp));
+  }
+}
+
+void doStringContSet()
+{
+  std::vector<String> st ({"a1"});
+  std::vector<String> end({"a5"});
```
mordante, line 155: This test is a lot easier to review than the one above.

tables/TaQL/test/tExprNodeSetOpt.cc
```diff
@@ -0,0 +1,361 @@
...
+  TableExprId id(0);
+  {
+    // Test closed-closed intervals
+    TableExprNodeSetOptContSetCC<String> set (TableExprNodeSet(), st, end);
+    set.show(cout, 2);
+    AlwaysAssertExit (! set.contains (id, "a0"));
+    AlwaysAssertExit (set.contains (id, "a1"));
+    AlwaysAssertExit (set.contains (id, "a3"));
+    AlwaysAssertExit (set.contains (id, "a5"));
+    AlwaysAssertExit (! set.contains (id, "a6"));
```
mordante, line 165: Should find be tested here too?

tables/TaQL/test/tExprNodeSetOpt.cc
```diff
@@ -0,0 +1,361 @@
...
+    AlwaysAssertExit (set.contains(id,54924.51) == trSet->contains(id,54924.51));
+    AlwaysAssertExit (set.contains(id,54926.) == trSet->contains(id,54926.));
+    AlwaysAssertExit (set.contains(id,54928.49) == trSet->contains(id,54928.49));
+    AlwaysAssertExit (set.contains(id,54928.5) == trSet->contains(id,54928.5));
+  }
+}
+
+void doDoubleMidWidth()
+{
+}
```
mordante, line 335: Here I miss a test.

mordante (commented):
For the next time it would be great the have smaller reviews, for example
1. rename hasFoo -> contains
2. moving files contents unchanged to new files
3. add the new functions

I don't see documentation updates, is there a PR to do that.

casa/Utilities/GenSort.tcc
```diff
@@ -504,7 +504,8 @@ template<class T, class INX>
 INX GenSortIndirect<T,INX>::sort (Vector<INX>& indexVector, const Block<T>& data,
                                   INX nr, Sort::Order ord, int opt)
 {
-    return sort (indexVector, data.storage(), min(nr, data.nelements()),
+    return sort (indexVector, data.storage(),
+                 std::min(nr, static_cast<INX>(data.nelements())),
```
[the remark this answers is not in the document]
gervandiepen: Added a comment to the class description in the header file.

tables/TaQL.h
```diff
@@ -38,34 +38,75 @@
...
+  //  to check (semantically) and execute commands such as SELECT, UPDATE, etc..
+  //  Note that subqueries are executed before the entire parse tree has
+  //  been walked through, thus before possible later semantic errors are
+  //  detected.
+  //  <br>Expressions in the parse tree are converted to expression trees
+  //  using the various TableExprNode classes. Expression trees are evaluated
+  //  for each row in a table. Note that expressions can be used in many
+  //  parts of a TaQL command.
+  //  Functions in a command are handled by TableParseFunc.
+  // </ol>
```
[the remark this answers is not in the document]
gervandiepen: Thanks

tables/TaQL/ExprNodeSetOpt.h
```diff
@@ -0,0 +1,358 @@
...
+
+  template <typename T>
+  class TableExprNodeSetOptUSet: public TableExprNodeSetOptBase
+  {
+  public:
+    // Construct an empty set.
+    TableExprNodeSetOptUSet (const TableExprNodeRep& orig, const Array<T>&);
+
+    // Show the node.
+    virtual void show (ostream& os, uInt indent) const override;
```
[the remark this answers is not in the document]
gervandiepen: Removed virtual in ExprNodeSet* header files.

tables/TaQL/ExprNodeSetOpt.h
```diff
@@ -0,0 +1,358 @@
...
+  };
+
+
+  // <summary>
+  // An optimized representation of a discrete selection set.
+  // </summary>
+
+  // <use visibility=local>
+
+  // <reviewed reviewer="UNKNOWN" date="before2004/08/25" tests="">
```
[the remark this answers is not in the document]
gervandiepen: Rectified the lines

tables/TaQL/ExprLogicNode.cc
```diff
@@ -286,41 +288,53 @@ TableExprNodeINInt::TableExprNodeINInt (const TableExprNodeRep& node,
...
     Array<Int64> arr(values.array());
     if (values.hasMask()) {
       // Remove masked elements.
       arr.reference (values.flatten());
     }
-    if (! arr.empty()) {
-      itsIndexSet.clear();
-      itsIndexSet.insert(arr.begin(), arr.end());
-    }
+    // Use an unordered_set for fast lookup.
```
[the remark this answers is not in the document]
gervandiepen: It is using an unordered_map now, so changed the comment.
Faster than what for uInt64 keys? An unordered_map is certainly much faster than using class TableExprNodeSet itself. I've tried a Block<Bool> for integer values with a small range, but that was hardly any faster than unordered_map. The map was needed for large ranges, so I decided to only use maps.
It is using Int64 keys, since the integer values are signed.

tables/TaQL/ExprNodeSetOpt.h
```diff
@@ -0,0 +1,358 @@
...
+    TableExprNodeSetOptUSet (const TableExprNodeRep& orig, const Array<T>&);
+
+    // Show the node.
+    virtual void show (ostream& os, uInt indent) const override;
+
+    // Where does a value occur in the set? -1 is no match.
+    virtual Int64 find (T value) const override;
+
+  private:
+    std::unordered_map<T,Int64> itsMap;
```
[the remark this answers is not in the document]
gervandiepen: It's a map now.
For TableExprNodeSet a set would suffice because it only needs to know if the value exists. But I'm also going to use it for joins, which need to know which value matches.
I'll update the comments in the header file.

tables/TaQL/ExprLogicNode.cc
```diff
@@ -286,41 +288,53 @@ TableExprNodeINInt::TableExprNodeINInt (const TableExprNodeRep& node,
                                         Bool)
 : TableExprNodeBinary (NTBool, node, OtIN)
 {}
-void TableExprNodeINInt::convertConstChild()
+TableExprNodeINInt::~TableExprNodeINInt()
```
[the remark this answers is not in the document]
gervandiepen: Use = default everywhere in ExprLogicNode.h

tables/TaQL/ExprNodeSetOpt.h
```diff
@@ -0,0 +1,358 @@
...
+  {
+  public:
+    // Construct an empty set.
+    TableExprNodeSetOptUSet (const TableExprNodeRep& orig, const Array<T>&);
+
+    // Show the node.
+    virtual void show (ostream& os, uInt indent) const override;
+
+    // Where does a value occur in the set? -1 is no match.
+    virtual Int64 find (T value) const override;
```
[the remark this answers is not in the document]
gervandiepen: Added a comment.
I've always learned that passing by value is better than by reference for e.g. int and double. Is that still true?

tables/TaQL/ExprNodeSet.cc
```diff
@@ -1288,137 +319,62 @@ MArray<MVTime> TableExprNodeSet::getArrayDate (const TableExprId& id)
...
-}
-
-Bool TableExprNodeSet::hasBool (const TableExprId& id, Bool value)
+Bool TableExprNodeSet::contains (const TableExprId& id, Bool value)
 {
     Bool result = False;
     size_t n = itsElems.size();
     for (size_t i=0; i<n; i++) {
-        castItsElem(i)->matchBool (&result, &value, 1, id);
+        itsElems[i]->matchBool (&result, &value, 1, id);
```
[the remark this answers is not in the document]
gervandiepen: Good suggestion. Added test on result.

tables/TaQL/ExprNodeSetElem.cc
```diff
@@ -0,0 +1,1053 @@
...
+    fillExprType (itsStart.get());
+    fillExprType (itsEnd.get());
+    fillExprType (itsIncr.get());
+  }
+
+  TENShPtr TableExprNodeSetElemBase::evalExpr (const TENShPtr& expr,
+                                               const TableExprId& id) const
+  {
+    TENShPtr res;
+    if (expr) {
```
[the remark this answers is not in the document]
gervandiepen: done

tables/TaQL/ExprNodeSetElem.cc
```diff
@@ -0,0 +1,1053 @@
...
+  }
+
+  TableExprNodeSetElemSingle::~TableExprNodeSetElemSingle()
+  {}
+
+  TENSEBShPtr TableExprNodeSetElemSingle::evaluate
+  (const TableExprId& id) const
+  {
+    return TENSEBShPtr(new TableExprNodeSetElemSingle
+                       (*this, evalExpr(itsStart, id)));
```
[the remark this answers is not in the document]
gervandiepen: I fear it won't work because make_shared makes a shared_ptr<TableExprNodeSetElemSingle>.

tables/TaQL/ExprNodeSetElem.cc
```diff
@@ -0,0 +1,1053 @@
...
+  }
+
+  Bool TableExprNodeSetElemSingle::isDiscrete() const
+    { return True; }
+
+  Bool TableExprNodeSetElemSingle::isSingle() const
+    { return True; }
+
+  void TableExprNodeSetElemSingle::fillVector (Vector<Bool>& vec, Int64& cnt,
```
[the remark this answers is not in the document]
gervandiepen: It's a virtual function also operating on arrays where many more elements are filled.

tables/TaQL/ExprNodeSetOpt.cc
gervandiepen: Added

gervandiepen:
> The TaQL note (199) has been updated to describe the new AROUND syntax.

tables/TaQL/ExprNodeSetElem.cc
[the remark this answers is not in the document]
gervandiepen: make_shared makes a shared_ptr to TableExrNodeSetElemSIngle, not to the base class. You need a cast thereafter to get the right shared pointer.

tables/TaQL/ExprNodeSetElem.cc
```diff
@@ -0,0 +1,1053 @@
...
+
+
+namespace casacore { //# NAMESPACE CASACORE - BEGIN
+
+  TableExprNodeSetElemBase::TableExprNodeSetElemBase (NodeDataType dt)
+    : TableExprNodeRep (dt, VTSetElem, OtUndef, Table())
+  {}
+
+  TableExprNodeSetElemBase::~TableExprNodeSetElemBase()
+  {}
```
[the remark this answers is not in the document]
gervandiepen: Made all destructors in this file default.

tables/TaQL/ExprNodeSetOpt.cc
gervandiepen: Removed the line as well as all statements using rowNrs. It was not needed anymore.

tables/TaQL/ExprNodeSetOpt.cc
gervandiepen: Now using std::less_equal. The interval is closed on the right side, thus needs <=.

tables/TaQL/ExprNodeSetOpt.cc
gervandiepen: Explained.

tables/TaQL/ExprNodeSetOpt.cc
gervandiepen: Nice idea. Only the show function is a bit hard.
I'll leave it as now.

tables/TaQL/ExprNodeSetOpt.h
gervandiepen: itsStarts and itsEnds cannot be moved to ContSetBase because they are templated, while ContSetBase is not.
The leftC and rightC vectors are empty for those classes, thus not a real problem I think.

tables/TaQL/ExprNodeSetOpt.cc
gervandiepen: I've reorganized the inheritance tree (and renamed classes) to address this and the previous point.
This class is now templated on Comparison.

tables/TaQL/test/tExprNodeSetOpt.cc
gervandiepen: Overhauled the test program. Using templated function to test find and contains.
Expected result is now a Vector making it more clear.
This also addresses the next points.

tables/TaQL/test/tExprNodeSetOpt.cc
gervandiepen: Was not needed, so removed it.

gervandiepen:
Merged the master into this branch because there were merge conflicts.

mordante (approved):
LGTM!

PR #1225, opened 2022-09-29
Cas 13921 Needed for CASA build for libsynthesis to link against liblapack

dmehring:


dmehring:
There was a question in PR 1223 on why this is needed. In our new build system, CASA's libsynthesis fails to link against liblapack without it. The liblapack dependency occurs in the casacore function that libsynthesis calls. My understanding is that we might not need this if we used homebrew for installing dependencies, but we use MacPorts. @FedeMPouzols and/or @cquike are better qualified to describe the details than I am.
I've placed the new code in an if (CASA_BUILD) guard as requested in PR 1223.

dmehring:
After discussion with @FedeMPouzols, we would like to hold off on merging this PR for the moment. The `if (CASA_BUILD)` guard is outdated. We would like to wait for stabilization of our new system before deciding how best to handle these new `PRIVATE_LIBS` options. We currently do have a workaround via a cmake command line directive, so we will use that in the meantime.

vsuorant:
CAS-13921 is already closed so I think we can close this one too.

PR #1227, opened 2022-10-12
Cas 11726 update to recognize older VLA archive files with ant positions rotated nonstandardly

dmehring:


dmehring:
I realize that the intersection of sets of those able to approve casacore pull requests and those who are expert VLA users who would understand this change may be the null set, so I've added some pretty detailed comments regarding this change. We (NRAO/CASA) obviously need this fix for our users who want to import old VLA archive files. This has gone through our validation process, so one of our stakeholders who needs this fix has tested and approved the change. So, it is safe for someone to merge here; even if there are concerns about the code, you can see that it only affects VLA users, so I'll be the one hearing about it if there are problems; casacore will not be responsible for the content of this change.
Thanks

tammojan (approved):
Thanks for the detailed comments @dmehring. Indeed this is very VLA specific, so safe to merg.

PR #1229, opened 2022-10-18
Skip van Vleck corrections based on new FITS-IDI keyword

kettenis:
As described in the VLBA Sensitivity Upgrade Memo 52, new versions of the DiFX correlator will apply the van Vleck when requested to do so.  This will be signalled by a new keyword in the FITS header of FITS-IDI output.  See

  https://library.nrao.edu/public/memos/vlba/up/VLBASU_52.pdf

for more details.

Recognize this new keyword and skip the van Vleck correction if it is set to a non-zero value.

### DIFF
--- a/msfits/MSFits/FitsIDItoMS.cc
+++ b/msfits/MSFits/FitsIDItoMS.cc
@@ -169,7 +169,9 @@ Vector<Double> FITSIDItoMS1::effChBw;
 //	
 // Constructor
 //	
-FITSIDItoMS1::FITSIDItoMS1(FitsInput& fitsin, const String& correlat, const Int& obsType, const Bool& initFirstMain)
+FITSIDItoMS1::FITSIDItoMS1(FitsInput& fitsin, const String& correlat,
+			   const Int& obsType, const Bool& initFirstMain,
+			   const Float& vanVleck)
   : BinaryTableExtension(fitsin),
     itsNrMSKs(10),
     itsMSKC(itsNrMSKs," "),
@@ -179,6 +181,7 @@ FITSIDItoMS1::FITSIDItoMS1(FitsInput& fitsin, const String& correlat, const Int&
     ///infile_p(fitsin),
     itsObsType(obsType),
     itsCorrelat(correlat),
+    itsVanVleck(vanVleck),
     msc_p(0)
 {
 
@@ -2247,6 +2250,11 @@ void FITSIDItoMS1::fillMSMainTable(const String& MSFileName, Int& nField, Int& n
 	  Rm = 1.0 / (A * H);
 	  alfa = 1.0;
 	  gamma = 1.0;
+	}
+
+	if (itsVanVleck != 0.0) {
+	  alfa = 1.0;
+	  rho = NULL;
 	}
 
 	bfactc = (gamma*gamma) / (A * Rm * alfa * H);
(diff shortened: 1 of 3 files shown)


msfits/MSFits/FitsIDItoMS.h
```diff
@@ -129,7 +129,9 @@ class FITSIDItoMS1 : public BinaryTableExtension
   // software correlator used by the EVN).
   //
 
-  FITSIDItoMS1(FitsInput& in, const String& correlat, const Int& obsType=0, const Bool& initFirstMain=True);
+  FITSIDItoMS1(FitsInput& in, const String& correlat,
+	       const Float& vanVleck=0.0, const Int& obsType=0,
```
tammojan, line 133: When adding a new argument with a default value, I think it's better to put it at the end.

msfits/MSFits/FitsIDItoMS.cc
```diff
@@ -2249,6 +2252,11 @@ void FITSIDItoMS1::fillMSMainTable(const String& MSFileName, Int& nField, Int& n
 	  gamma = 1.0;
 	}
 
+	if (itsVanVleck != 0.0) {
```
tammojan, line 2255: Isn't exact comparison with a floating point number a bit scary?

msfits/MSFits/MSFitsIDI.cc
```diff
@@ -292,19 +292,24 @@ void MSFitsIDI::readFITSFile(Bool& atEnd)
 
   // Correlator
   String correlat;
+  Float vanVleck = 0;
```
tammojan, line 295: Style: I'd prefer `0.`;

tammojan (commented):
Looks good to me, but I'd appreciate it if @dpetry could have a look as well.

msfits/MSFits/FitsIDItoMS.h
kettenis: I can see your point. Conceptually the new vanVleck parameter is related to the existing correlat parameter, so that is why I added it there.  But I can change it.

This actually relates to what I consider to be a bit of an issue with how the code is structure. I think that for all practical purpose the FITSIDItoMS1 class should be considered an "internal" API. Users should be using the MSFitsIDI class (and this is what importfitsidi does). But there is nothing in the documentation that spells this out.

msfits/MSFits/MSFitsIDI.cc
kettenis: I'll fix that.

msfits/MSFits/FitsIDItoMS.cc
kettenis: Comparison against 0.0 is perfectly safe as zero can be represented exactly. And since the memo referenced in this PR documents non-zero values as doing something different, this is exactly what is needed.

The habit if representing what should clearly be integer parameters as floats in FITS-related things is weird but that's not something we can change.

msfits/MSFits/FitsIDItoMS.h
tammojan: Given that there is no documentation (either about the parameters, or about the fact that this should be an internal API), could you move the parameter to the end to avoid bugs?

msfits/MSFits/MSFitsIDI.cc
kettenis: Done

tammojan (approved):
Looks good to me; let me know if you want to wait for @dpetry's approval.

kettenis:
Let's move on with this. I have the blessing from the NRAO/VLBA folks to move ahead with this.

PR #1236, opened 2022-11-01
Add pre-compiled header support

rtobar:
Using pre-compiled headers speeds up the build depending on what goes into the headers, the build type, the compiler itself, etc. For some numbers see https://github.com/casacore/casacore/pull/1236/commits/809c898b4d80d489f3be4857704f9e413f448cc9.

The PCHs created here consist on the top-level headers for each of the modules, and apply both to the sources going into the shared libraries and to app and test executables, which covers already a good starting base. In the casa library in particular this caused a couple of issues due to function overload problems, which are fixed in this PR. This exercise also found some duplicated code that I removed.

More work can probably be done to speed things up further, but let's see how this plays out in CI first.


rtobar:
Good to see that CI didn't break, so this first pass wasn't too bad. However the gains are not that noticeable, so I'll keep banging at this to add PCH support for test builds too. I'll un-draft this PR once I'm happy with those builds too.

rtobar:
OK, I'm happy with this now. I extended the PCH support so that all apps and tests also benefit from it. Build times even for CI are now clearly better (the slowest job was 44m, now it's 34m).

rtobar:
ping @tammojan, do you think this would be an invited addition?

rtobar:
ping again

casa/Quanta/MVTime.h
```diff
@@ -460,9 +460,19 @@ ostream &operator>>(ostream &is, MVTime &meas);
 ostream &operator<<(ostream &os, const MVTime::Format &form);
 // </group>
 
-// is equal operator, uses operator Double which returns days
+// equality and comparisson operators, use operator Double which returns days
```
tammojan, line 463:
```suggestion
// equality and comparison operators, use operator Double which returns days
```

cmake/PCHSupport.cmake
```diff
@@ -0,0 +1,27 @@
...
+function(init_pch_support shlib headers)
+  if (NOT USE_PCH)
+    return()
+  endif()
+  target_precompile_headers(${shlib} PRIVATE ${headers})
+  set(_pch_pie_target ${shlib}_pch_pie)
+  set(_pch_pie_target ${_pch_pie_target} PARENT_SCOPE)
+  file(TOUCH ${CMAKE_CURRENT_BINARY_DIR}/empty.cpp)
+  add_library(${_pch_pie_target} STATIC ${CMAKE_CURRENT_BINARY_DIR}/empty.cpp)
+  set_target_properties(${_pch_pie_target} PROPERTIES PROSITION_INDEPENDENT_CODE OFF)
```
tammojan, line 18:
```suggestion
  set_target_properties(${_pch_pie_target} PROPERTIES POSITION_INDEPENDENT_CODE OFF)
```
Silly that CMake doesn't warn about this typo.

tammojan (changes requested):
Thanks for the ping! Who wouldn't want shorter compile times (even though on gcc they're not as impressive as on clang). Also, the complexity added to the build system is not too large.
Apart from two typos this looks good to me.

casa/Quanta/MVTime.h
rtobar: Changed, I always get confused with this one...

cmake/PCHSupport.cmake
rtobar: Yeap, that should result in some warning of some type...

Since the target was a static library the property was already OFF by default, but I guess it doesn't hurt to make it explicit anyway, so I kept it with the typo corrected.

rtobar:
Thanks @tammojan for the review. I've applied your suggestions, and also took the opportunity to rebase on top of the latest `master` to have a less confusing git history.

tammojan (approved):

PR #1240, opened 2022-11-10
Set AIPS_DEBUG when building in debug mode

### LINKED ISSUE #1239: DebugAssert is not working when compiling in debug mode
It appears that AIPS_DEBUG is not set in CMakeLists.txt when compiling in debug mode.

gervandiepen:
CMakeLists.txt now sets AIPS_DEBUG when needed.
Added test program tAssert.cc to test the ASSERT macros.

Close #1239


tammojan (approved):
Nice addition of tests. Not sure when the definition is `AIPS_DEBUG` got lost, it must have worked at some time...

PR #1244, opened 2022-11-14
First stage of join clause in TaQL

gervandiepen:
This is the first part of the changes for support of the JOIN clause in TaQL.
This part contains the changes in the scanner/parser and the code to handle the result from the parser and store it in a tree. Handling the result tree is done in a separate PR.

The syntax of the join looks like:

select t1.col1, t2.col2, ... from table table1 t1 JOIN table2 t2 ON condition JOIN table3 t3 ON condition ... WHERE ...

Theta joins (implicit joins in the SELECT clause) are not supported. They are discouraged in SQL.

### DIFF
--- a/tables/TaQL/TaQLNodeHandler.cc
+++ b/tables/TaQL/TaQLNodeHandler.cc
@@ -388,8 +388,7 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
   TaQLNodeResult TaQLNodeHandler::visitJoinNode (const TaQLJoinNodeRep&)
   {
-    throw TableInvExpr ("join is not supported yet");
-    return TaQLNodeResult();
+    throw TableInvExpr("Joins are not implemented yet");
   }
 
   TaQLNodeResult TaQLNodeHandler::visitGroupNode (const TaQLGroupNodeRep& node)
@@ -529,7 +528,7 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
     // Furthermore, handle GIVING first, because projection needs to know
     // the resulting table name.
     visitNode     (node.itsGiving);
-    visitNode     (node.itsJoin);
+    handleJoins   (node.itsJoins);
     handleWhere   (node.itsWhere);
     visitNode     (node.itsGroupby);
     visitNode     (node.itsColumns);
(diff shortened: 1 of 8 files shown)


tables/TaQL/TaQLNodeHandler.cc
```diff
@@ -943,6 +942,17 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
     }
   }
 
+  void TaQLNodeHandler::handleJoins (const TaQLMultiNode& node)
+  {
+    if (! node.isValid()) {
+      return;     // no joins
+    }
+    const std::vector<TaQLNode>& nodes = node.getMultiRep()->itsNodes;
+    for (uInt i=0; i<nodes.size(); ++i) {
```
tammojan, line 951: I guess a range based loop could be used here, but I would immediately agree if you prefer the current way for consistency within the file.

tables/TaQL/TableGram.yy
```diff
@@ -408,17 +413,17 @@ withpart:  {   /* no WITH part */
 
 /* The SELECT command; note that many parts are optional which is handled
    in the rule of that part. The FROM part being optional is handled here
-   because later a join might be added. */
-selcomm:   withpart SELECT selcol FROM tables whexpr groupby having order limitoff given dminfo {
+   because a join might be used. */
+selcomm:   withpart SELECT selcol FROM tables joins whexpr groupby having order limitoff given dminfo {
                $$ = new TaQLQueryNode(
-                    new TaQLSelectNodeRep (*$3, *$1, *$5, 0, *$6, *$7, *$8,
-					   *$9, *$10, *$11, *$12));
```
tammojan, line 415: Sorry, I don't understand this code yet. I see that the fourth argument of `TaQLSelectNodeRep` is a `joins`. In the old code, that was `0`, probably meaning that joins were not supported. In the new code, the order of the arguments shift, and the `$6` that was the fifth argument in the old code, is now the fourth argument. Is the meaning of `$6` changed between the old code and the new code because extra stuff is parsed?

tammojan (approved):
Looks good, but could you help me understand a bit of code?

tables/TaQL/TaQLNodeHandler.cc
gervandiepen: You're right. It is used at many other places in the code.
I might change it one day.

PR #1245, opened 2022-11-15
Remove all  CVS lines from the source files

### LINKED ISSUE #1243: Remove obsolete SVN/CSV magic $Id$ comments
These entries aren't used with git so their information is no longer useful. This was discussed in #1164

gervandiepen:
About all source files contain a $Id line, which is removed. Also the empty comment lines above it are removed.
The standard header has been added to a few source files. where it was missing.

close #1243


mordante (approved):
Thanks, I did some spot checking but didn't review all changes.
LGTM.

PR #1247, opened 2022-11-28
Second stage of JOIN clause in TaQL

gervandiepen:
The TaQL code has been prepared for the execution of the JOIN clause.
- Until now the tables in a FROM clause had to have the same size, but that had to be changed for the join. Hence, the TableExprNodeRep objects do not contain a Table object anymore. Instead, the column nodes have a TableExprInfo object containing the Table object, but also its alias and if it is a join table.
- A new function flattenTree has been added to traverse the entire expression tree and turn it into a vector of nodes. This function is now also used instead of the old tree traversal function (such as getAggrNodes and getColumnNodes). New functions such as isAggregate are implemented and used in ExprNodeUtil.h/cc to get the old behaviour.
- More modern C++ has been used in affected classes (virtual->override, default destructor, dynamic_cast, shared_ptr).
- A test program executing erroneous commands has been added and some error messages have been improved.

### DIFF
--- a/ms/MSOper/MSMetaData.cc
+++ b/ms/MSOper/MSMetaData.cc
@@ -3519,7 +3519,7 @@ Quantity MSMetaData::getEffectiveTotalExposureTime() {
     Double totalExposure = 0;
     String taql = "select FLAG, DATA_DESC_ID, EXPOSURE, TIME from "
         + _ms->tableName() + " where ANTENNA1 != ANTENNA2";
-    Table result(tableCommand(taql));
+    Table result(tableCommand(taql).table());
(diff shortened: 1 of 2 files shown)


tammojan:
[This](https://github.com/casacore/casacore/actions/runs/3738843705/jobs/6345366467#step:8:3008) looks like an actual test failure to me:
```
/Users/runner/work/casacore/casacore/ms/MSOper/MSMetaData.cc:3522:11: error: no matching constructor for initialization of 'casacore::Table'
    Table result(tableCommand(taql));
```

gervandiepen:
I'm already fixing it.

PR #1250, opened 2022-11-30
Add support for compiling with ADIOS but without MPI

rtobar:
This PR contains the necessary changes both in the build system and in the code to enable compilation of the `Adios2StMan` without MPI support. While historically `Adios2StMan` has required MPI to work, this is not a constraint that always applies. Two main uses cases apply:
 * Reading data written with `Adios2StMan` doesn't require MPI, regardless of whether the data was written in parallel or serially
 * Even in a single process it might still be possible gain from using `Adios2StMan` and the different underlying ADIOS technologies, like data compression and different storage technologies.

Most of the changes are to the build system and adding `#ifdef`s around. The only main change that is user facing is the fact that the construct list has changed (see https://github.com/casacore/casacore/commit/962e051372db8116030684b9ad7d63f06eac012a for details). However this storage manager isn't widely used yet, so there's no expectation that this will wreak havoc around the world.

(cc: @VoronkovMA, @pelahi)


gervandiepen (approved):
LGTM

rtobar:
Ping, this has been accepted, is there anything else needed to have it merged? Thanks!

tammojan:
@rtobar Could you please resolve the merge conflict?

rtobar:
@tammojan conflict resolved, also rebased on top of latest `master`.

PR #1251, opened 2022-12-12
Handle VLBA/DiFX data without zero-padding correctly 

kettenis:
Correction of the autocorrelations in the lag domain can only be
done properly if the correlator has zero-padded the data.  Handle
the case where data has not been zero-padded and assume this is
the default as DiFX doesn't zero-pad.

The VLBA correlator indicates whether data has been zero-padded by
setting the ZERO_PAD keyword in the MODEL_COMPS table to a non-zero
value.  Use this to override the default of assuming no zero-padding
has been done.

### DIFF
--- a/msfits/MSFits/FitsIDItoMS.cc
+++ b/msfits/MSFits/FitsIDItoMS.cc
@@ -171,7 +171,7 @@ Vector<Double> FITSIDItoMS1::effChBw;
 //	
 FITSIDItoMS1::FITSIDItoMS1(FitsInput& fitsin, const String& correlat,
 			   const Int& obsType, const Bool& initFirstMain,
-			   const Float& vanVleck)
+			   const Float& vanVleck, const Int& zeroPad)
   : BinaryTableExtension(fitsin),
     itsNrMSKs(10),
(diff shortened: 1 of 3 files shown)


dpetry:
@kettenis , hi, looks good but somewhere there should be a comment saying what the possible values of zeropad are and what their meaning is since you used an int and not a bool.

kettenis:
Thanks @dpetry, the ZERO_PAD keyword is defined in AIPS memo 114, which documents it as "Zero padding factor", with further text stating that
`The oversampling and zero padding "factors" are given by the OVERSAMP and ZERO PAD keywords; a value of 0 for these keywords implies no oversampling and no zero padding.`
The code really only cares about the value being 0 to detect the no zero padding case. I think adding the comment at the point where the keyword is read makes the most sense.

dpetry (changes requested):
OK, as I said, a little comment somewhere saying that zeropad can be 0 (meaning no padding applied) and non-zero (e.g. typically 1?, meaning zero padding has been applied) should be added.

kettenis:
Yes, just pushed a revision that does this.

PR #1252, opened 2022-12-13
Add Python 3.11 support

gmloose:
Added support for building Python 3.11 docker image. Dropped support for Python 3.6, which is EOL.


aroffringa (approved):

gmloose:
@aroffringa, could you please merge it. I don't have write access.

Issue #1253, opened 2022-12-19
cascacore::Path::absoluteName() strips off leading dot (.)


gmloose:
The function `cascacore::Path::absoluteName` strips off a leading `.` from a file or directory name, which is only correct if the following character is a `/` (i.e. stripping off a leading `./`). In other cases this yields an absolute path that is not equivalent to the original relative path. 

The culprit is in the private function `makeAbsoluteName()`. The comment in that function says: `Remove a possible leading . or ./`, which it indeed does. Question is if this is expected behavior. I'm inclined to say it is not. 

Here's a small program that demonstrates this behavior.
```
#include <casacore/casa/OS/Path.h>
#include <string>
#include <iostream>

int main()
{
  casacore::Path outPath(".bashrc");
  std::string result = outPath.absoluteName();
  std::cout << "Absolute path to file "
            << outPath.originalName() << ": "
            << outPath.absoluteName() << std::endl;
}
```
If I run this program from my home directory, I get the following output:
```
Absolute path to file .bashrc: /home/marcel/bashrc
```

PR #1254, opened 2022-12-19
Fix issue #1253

gmloose:
Fixex issue #1253. 

`Path::absolutName()` no longer strips off the leading dot of so-called dot-files -- files or directories whose name start with a `.`.


casa/OS/Path.cc
```diff
@@ -395,23 +395,12 @@ String Path::makeAbsoluteName (const String& inString) const
...
-                case '.' :
-                   from = 0;
-                   break;
-               }
-	    } 
-	    workString = workString.from (from);
-	}
+    if (workString == ".") {
+        workString = "";
+    } else if (workString.before(2) == "./") {
```
mordante, line 402:
```suggestion
    } else if (workString.startsWith("./")) {
```
For readability I would change this, especially since `std::string::starts_with` exists too.

casa/OS/Path.cc
```diff
@@ -395,23 +395,12 @@ String Path::makeAbsoluteName (const String& inString) const
 	return inString;
     }
     // Otherwise we have a relative pathname.
-    // Remove a possible leading . or ./
+    // Remove a possible leading ./
```
mordante, line 398: I would reword this since it partly describes the code but not completely.
No real suggestion, but removing this line would be fine by me. The line above would be enough for me.

mordante (commented):
Thanks for the fixes, some comments, but the general approach looks good.

casa/OS/Path.cc
gmloose: Changed as suggested.

mordante:
LGTM, please give @gervandiepen some time to have a second look too.

mordante (approved):

PR #1261, opened 2022-12-27
Issue 1242

dmehring:


gervandiepen:
I dislike putting different changes in a single PR.
Especially, IMHO the types change is too much and should have been discussed. I prefer doing the entire casacore in one PR. Furthermore, the Int->int change is too rough. E.g., ImageInterface is turned into Imageinterface. Also changing Int to int is a bad idea. It should be changed to int32_t because Int is guaranteed to be 32 bits, which int is not. So the types change should be reverted.

PR #1263, opened 2023-01-11
Restore ExprNode::table()

mnijhuis-tos:
After removing ExprNode::tabe(), DP3 does no longer compile. We can easily restore backwards compatibility by adding a new table() function that uses the new getTableInfo() function, which replaced the table() function.


aroffringa (approved):
Looks good to me, but let's also wait what Ger says.

gervandiepen (approved):

PR #1269, opened 2023-01-18
Remove allocator Template parameter from Arrays

aroffringa:
This simplifies the Array code quite a lot, and probably (not tested) results in a compilation speed up. It also improves compiler error output when a parameter is in error.

Finally, although leaving the template allocator empty results in the default state-less std::allocator, which the compiler should be able to optimize away, in practice this wasn't true. The Array constructor in particular became quite a bit faster by removing the allocator.

### DIFF
--- a/casa/Arrays/Array.h
+++ b/casa/Arrays/Array.h
@@ -44,7 +44,7 @@ namespace casacore { //#Begin casa namespace
 
 // <summary> A templated N-D %Array class with zero origin. </summary>
 
-// Array<T, Alloc> is a templated, N-dimensional, %Array class. The origin is zero,
+// Array<T> is a templated, N-dimensional, %Array class. The origin is zero,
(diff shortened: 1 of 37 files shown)


tammojan:
In case you missed it, CI says there's still an allocator template argument used in casa/IO/ArrayIO.h:101:
```
LogIO &operator<<(LogIO &os, const Array<T, Alloc> &a);
```

aroffringa:
Maybe this is something for @gervandiepen to review?

gervandiepen:
I'll do it next week.

gervandiepen:
Looks like a good change.
It is not possible to allocate in shared memory anymore, but that is no problem because it can be done by allocating first and use the pointer to make the Array object. With the old Alloc parameter a shared memory Array would be resized in shared memory which would be very bad. That is not possible anymore.

gervandiepen (approved):

PR #1280, opened 2023-01-31
Replace PtrHolder by std::unique_ptr

gervandiepen:
All instances of (S)PtrHolder have been replaced by std::unique_ptr. Both scalar and array values are addressed.
In a few files it also required changing casts, which have been replaced by static_cast.

### DIFF
--- a/casa/BasicMath/Random.cc
+++ b/casa/BasicMath/Random.cc
@@ -28,8 +28,8 @@
 #include <casacore/casa/Exceptions/Error.h>
 #include <casacore/casa/Utilities/Assert.h>
 #include <casacore/casa/BasicSL/String.h>
-#include <casacore/casa/Utilities/PtrHolder.h>
 #include <casacore/casa/Arrays/Vector.h>
+#include <memory>
 
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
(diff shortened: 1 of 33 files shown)


mnijhuis-tos (approved):
Nice improvement! Some small remarks:
- It's too bad we can't use `std::make_unique` from C++14.
- Many lines still have a space between the closing `>` with nested templates. For example, `std::unique_ptr<Function<T> >`. The spaces are no longer needed with C++11. Since you're changes those lines anyway, you could consider removing the space as well.
- After checking out the branch, I verified there are no more uses of `(S)PtrHolder`, besides the tPtrHolder test.

gervandiepen:
Removed space from ' > >'
make_unique can only be used for concrete classes, thus not for (abstract) base classes.

mnijhuis-tos (approved):
In QuantumHolder.cc (which github does not show by default, because of the large number of changes) there are still many occurrences of `> >` which could be replaced by `>>`.
However, even without this further improvement, this PR is a good improvement.

mnijhuis-tos (changes requested):
Removing the space between '> >' is nice, however, it seems those spaces magically moved to `reset()` calls... There is now a space between `reset` and `(new ...`, which was not there before. Other `reset` calls in this PR also do not have that space.

gervandiepen:
I find it better readable with a space after reset. There is no casacore guideline wrt spaces.

mnijhuis-tos (approved):
Ok, I won't block this PR because of style discussions.

PR #1282, opened 2023-01-31
Fixed math in fftshift within FFTServer.hcc

### LINKED ISSUE #1281: fftshift has incorrect math
Functions within CASA that regrid data include 'fftshift' as an interpolation method, which points to the fftshift function within scimath/Mathematics/FFTServer.hcc. These functions yield demonstrably incorrect results when comparing 'fftshift' to other methods such as 'linear' or 'cubic'. For example, attached is an image showing a simple Gaussian in the TOPO frame that has been regridded to LSRK (requiring a frequency shift + interpolation). 'linear' yields a properly shifted Gaussian, while 'fftshift' yields a nonsensical result (linear in blue, fftshift in red).
Screenshot from 2023-01-24 14-04-02

In cases where the data has multiple timestamps, the first timestamp is consistent between the two methods, while further timestamps yield increasingly confusing results.
Screenshot from 2023-01-24 16-32-07

Fourier transforming the results from fftshift demonstrates that one half of the FT'd signal is corrupted in some way, with an improper phase rotation as a function of timestamp, leaking power into the imaginaries.
Screenshot from 2023-01-24 16-49-56

I have traced this back to incorrect math within the fftshift function, where the phase shift gradient is calculated as a function of frequency and the relative shift needed for that timestamp. A pythonic implementation of this sort of shift would generally use something like numpy.fft.fftfreq: https://numpy.org/doc/stable/reference/generated/numpy.fft.fftfreq.html. Instead, the casacore implementation uses a monotonically increasing function across the entire band via a for loop. The difference between these can qualitatively be seen here (fftfreq in blue, casacore version in orange):
Screenshot from 2023-01-25 15-30-24

This is why the casacore version yields strange phase rotations on half the band. Phase gradients shown here for the casacore implementation - note that the left hand side has random intercepts, while the right-hand side has proper behavior.
Screenshot from 2023-01-31 11-30-52

Modifying the casacore math from `exp(Double(i)*exponent)` to `exp(-Double((i+vsize/2)%vsize-(vsize/2))*exponent)` changes the behavior to that of fftfreq and fixes the problem. A version of CASA built with this modification yields the following result (linear blue, fftshift red; you can't see linear b/c they are so close and overlaid):
Screenshot from 2023-01-31 11-10-36

I have tested this with both odd and even numbers of channels, as well as off-center signals.

ryanaloomis:
Fixes #1281

### DIFF
--- a/scimath/Mathematics/FFTServer.hcc
+++ b/scimath/Mathematics/FFTServer.hcc
@@ -511,7 +511,7 @@ fftshift(Array<S> & cValues, const uInt& whichAxis,
     Array<S> cv = ait.array(); // reference
     fft0(cv, toFrequency);
     for(uInt i=0; i<vsize; i++){
-      cv(IPosition(1,i)) *= (S)exp(Double(i)*exponent);
+      cv(IPosition(1,i)) *= (S)exp(-Double((i+vsize/2)%vsize-(vsize/2))*exponent);
     }
     fft0(cv, !toFrequency);
     ait.next();
@@ -574,7 +574,7 @@ fftshift(Array<S> & outValues, Array<Bool> & outFlags,
     // apply shift
     fft0(cv, toFrequency);
     for(Int i=0; i<vsize; i++){
-      cv(IPosition(1,i)) *= (S)exp(Double(i)*exponent);
+      cv(IPosition(1,i)) *= (S)exp(-Double((i+vsize/2)%vsize-(vsize/2))*exponent);
     }
     fft0(cv, !toFrequency);
 
@@ -675,7 +675,7 @@ fftshift(Array<T> & outValues, Array<Bool> & outFlags,
     Array<S> cResult;
     fft0(cResult, rv);
     for(uInt i=0; i<cResult.size(); i++){
-      cResult(IPosition(1,i)) *= (S)exp(Double(i)*exponent);
+      cResult(IPosition(1,i)) *= (S)exp(-Double((i+vsize/2)%vsize-(vsize/2))*exponent);
     }
     fft0(rv, cResult);
 


tammojan:
Thanks for the thorough write-up. Could you have a look at the failing test? It [errors](https://github.com/casacore/casacore/actions/runs/4056459001/jobs/6981843010#step:9:523) with:
```
(scimath/Mathematics/test/tFFTServer.cc : 1800) Failed AlwaysAssert (abs(diff.real())<2E-5) && (abs(diff.imag())<2E-5)
```
It could be that the ground truth used in the test is wrong, but we should be very sure before we change this.

ryanaloomis:
Taking a closer look at the test, it actually explains why this issue wasn't caught earlier - the tests only examine integer value shifts (or sequential partial integer shifts that add up to a full integer), which will not demonstrate the phase issue that I diagnosed, as it will effectively have wrapped around an integer multiple of phase rotations. Adding a partial integer shift test would be prudent.

Regarding why the test fails with my version - it appears that there is a sign difference between what the tests expect and what my version yields. I.e. my version shifts left when it expects right, and vice-versa. To be honest, I noticed this when I implemented it, but flipped the sign until things looked right in the CASA output... Perhaps there is a sign flip somewhere in CASA that I compensated for here? I'm convinced that what I implemented here + the current CASA code gives the right final result, but two wrongs -> right is probably not the best thing to rely on...

I'll do some digging to try to track down where the sign flip in CASA is, or if there's another explanation.

ryanaloomis:
After further investigation, I now believe that there is actually a small bug in CASA as well as casacore.

My pull request here which 'fixes' the issue in CASA has a negative sign in it to force the correct direction of the shift in CASA. The problem as you pointed out is that this makes an objectively incorrect fftshift algorithm in casacore (where leftward shifts go toward the right, and vice-versa). This is what was caught by the automated tests.

Investigating the CASA code more carefully, I believe the offending line in CASA is in https://open-bitbucket.nrao.edu/projects/CASA/repos/casa6/browse/casatools/src/code/mstransform/TVI/RegriddingTVI.cc, line 520:
`			fftShift_p = - absoluteShift / bandwidth;`

It is possible that I'm misinterpreting the intention of this block of code, but I believe that this is where the flip is happening, and the correct code would instead be:
`			fftShift_p = absoluteShift / bandwidth;`

After discussion with the CASA team, I have updated my branch and I believe the automated tests should be rerunning now on the pull request. I anticipate that they will pass now... if not, back to the drawing board.

ryanaloomis:
Hi - is there any additional information I can provide? I think that the automated tests won't run without approval.

vsuorant:
Hi, 

What is the state of this ticket? Does this still need merging and if so, what needs to happen to get it merged?

dpetry:
@vsuorant , @tammojan , if needed I could look into this after 5 May.
But you, @tammojan , might know better anyway why this got stuck.

vsuorant:
Hi,

What is the status of this?

tammojan:
I've asked a colleague at JIVE to ask at this as well. I cannot see the outcomes of casa tests.

tammojan:
Reviewed by JIVE's Des Small:

The Fourier shift theorem is usually only invoked for integer shifts — some sources claim it is only defined in that case. But it can be extended to non-integer values, and since it doesn't hurt anyone that doesn't use that functionality it seems reasonable to support that.

I find a stackoverflow page[[1](https://stackoverflow.com/questions/21830878/shift-the-elements-of-a-vector-by-non-integer-shift-in-matlab), [2](https://stackoverflow.com/a/76971365/4140955)] that says "the Fourier shift theorem only works for integer values unless you are careful to make your frequency ramp continuous across DC"; the code exhibited there is functionally the same as the patch from Ryan Loomis.

I admit I have not tried to understand exactly why this constraint eliminates unwanted ringing, but I would recommend accepting the patch on the grounds that

(a) It doesn't hurt anyone who doesn't use it (except the CASA code that apparently needs to be tweaked in sync)
(b) There is at least some prior art supporting this solution.

dpetry:
@tammojan , that sounds reasonable. Thanks. 
@vsuorant , when this is merged, Patrick or Josh need to follow up on the required change in
https://open-bitbucket.nrao.edu/projects/CASA/repos/casa6/browse/casatools/src/code/mstransform/TVI/RegriddingTVI.cc
line 520  (remove the minus). Can you communicate that to them or shall I?

dpetry (approved):
Thanks, Ryan, then all is well.

dpetry:
@tammojan , could you have a look at the two checks which haven't completed, yet?
Do these need to be re-started?

ryanaloomis:
@dpetry @tammojan - I have figured out what is going wrong, but first I should apologize that I didn't take the time earlier to learn how to run these tests myself. I am not a C++ developer, and had simply ported over math from some python code I had written while diagnosing the bug. I wasn't sure how to actually properly compile casacore + run its test suite, so I poorly substituted with some validation tests via a build of CASA overtop of my modified casacore, which didn't probe the same full parameter space as the casacore test suite.

I've rectified this and spent the time now to learn how to compile casacore + run its tests, which helped me debug that my earlier code didn't respect the fact that the iterator variable 'i' in the for loops is an unsigned int, rather than an int. For smaller fractional shifts this is not an issue, but for larger shifts this causes the math in my previous commit to overflow (prior to the double cast) and return garbage results, which I wasn't catching with my tests within CASA. Adding explicit Int() casts within that expression fixes the problem and results in 1) passing casacore tests, 2) correct behavior with fractional shifts, and 3) correct behavior with larger shifts. I will add another commit shortly with these changes.

ryanaloomis:
Changes committed. And just to be explicit - @dpetry, nothing should need to be changed with the test values being compared against. The error was on my end for larger shifts, and the test values are still correct even with this new phase ramp. (Nominally it would be good to add fractional shift tests, which would have caught the original phase ramp issue, but that seems outside of scope of this pull request.)

PR #1283, opened 2023-02-01
Fix String.h hash namespace

vsuorant:
Casa build fails with the attached error when using the latest Casacore version: 
In file included from casacore/casacore/casa/Exceptions/Error.h:31:0,
                 from binding/source/conversions_python.cc:6:
casacore/casacore/casa/BasicSL/String.h:1039:13: error: specialization of ‘template<class _Tp> struct std::hash’ in different namespace [-fpermissive]
 struct std::hash<casacore::String>

This patch fixes the compilation error.


tammojan (approved):
Oops. Thanks!

PR #1287, opened 2023-02-09
Fix leftovers of old whatType() calls in LatticeHistograms

FedeMPouzols:
This updates some old leftovers of the old whatType() functions that were still around in the LatticeHistograms template.

In this PR: https://github.com/casacore/casacore/pull/1185/files#diff-511c3c4d4bebbedf81577943a6ec6f82cd76907d3f7ba19b244d637b1533a8fa the old whatType() functions were replaced by template specializations. There were some old style dummy+whatType calls left in LatticeHistogram.tcc that break compilation of one test (ImageAnalysis/test/dImageHistograms.cc) added now in the new CASA build system.


aroffringa (approved):
Thanks -- Looks good to me

PR #1290, opened 2023-02-15
CAS 13973 add methods to retrieve alma specific spw data

dmehring:


ms/MSOper/MSMetaData.cc
```diff
@@ -4942,11 +4968,59 @@ MSMetaData::ColumnStats MSMetaData::getIntervalStatistics() const {
     return stats;
 }
 
+std::shared_ptr<vector<int>> MSMetaData::_almaReceiverBands(uint nspw) const {
+    // for ALMA specific receiver band (RB) info
```
dpetry, line 4972: Maybe add a few more lines of comments on how the the receiver bands are expected to be defined in the ASDM_RECEIVER table and where this table comes from (i.e. that we are expecting importasdm to have copied it over with the "asis" feature verbatim from the ASDM). After all, ASDM_RECEIVER is not a standard MS subtable.

ms/MSOper/MSMetaData.cc
```diff
@@ -2128,6 +2130,30 @@ vector<String> MSMetaData::getSpwNames() const {
...
+    );
+    vector<int> out;
+    for (const auto spw: props) {
+        out.push_back(spw.rb);
+    }
+    return out;
+}
+
+vector<int> MSMetaData::getSpwSubwindows() const {
+    std::set<uInt> avgSpw, tdmSpw, fdmSpw, wvrSpw, sqldSpw;
```
dpetry, line 2146: Please document somewhere the concept of SPW subwindows. This is not a standard MS feature.

dpetry (changes requested):
Would be good to add a few lines of comments explaining where the ASDM_RECEIVER table and the concept of SPW subwindows comes from since both of these are not part of the standard MS.
Concerning testing: how does this code behave when operating on an MS which does not have the ASDM_RECEIVER table, e.g. simulated ALMA data or data from other observatories?

renaudmiel:
I fully agree with @dpetry 's comments

dmehring:
I've added requested documentation. The code behaves as it did previously if the ASDM_RECEIVER table doesn't exist; the relevant block of code is only executed if this table exists. See line 4978

dpetry (approved):

PR #1300, opened 2023-02-23
Fix build warning in tables/DataMan

gervandiepen:
Also the code has been modernized a bit.


rtobar:
I don't know much about this corner of the code, but don't the changes now fully, silently ignore the `tsmOption().bufferSize()` option? I don't know where this is set, or whether it's user-facing or not, but something to consider.

gervandiepen:
This option has never been used. It is even not possible to use the
tsmoption 'buffer' because there is some flaw somewhere in the code I have
to look at.

rtobar (approved):
I don't really know how much my opinion weights here, but it LGTM then! :)

tammojan (approved):

PR #1301, opened 2023-02-24
Fix build warning in LCSLicer

gervandiepen:
The compiler warned about using MimicSource as a float number, because it is not exact.
The code has been changed to be more relaxed in the test on MimicSource. It appeared that the test program did not test on it, so MimicSource tests have been added.

### DIFF
--- a/lattices/LRegions/LCSlicer.cc
+++ b/lattices/LRegions/LCSlicer.cc
@@ -48,11 +48,11 @@ LCSlicer::LCSlicer (const Vector<Float>& blc, const Vector<Float>& trc,
 		    RegionType::AbsRelType absRel)
 : itsBlc (blc.copy()),
   itsTrc (trc.copy()),
-  itsInc (blc.nelements())
+  itsInc (blc.size())
 {
     itsInc = 1;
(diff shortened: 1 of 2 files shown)


lattices/LRegions/LCSlicer.cc
```diff
@@ -327,53 +327,55 @@ void LCSlicer::fill()
...
     // Check if the region is absloute, fractional, unspecified, or strided.
+    // The float representation of MimicSource is inaccurate (648 instead of 646
+    // as the last digits), so be relaxed in testing it.
     itsIsFractional  = False;
     itsIsAbsolute    = True;
     itsIsUnspecified = False;
     itsIsStrided     = False;
     for (uInt i=0; i<nr; i++) {
-        if (itsBlc(i) == Slicer::MimicSource) {
+        if (static_cast<Int64>(itsBlc[i]) < Slicer::MimicSource+10) {
```
tammojan, line 361: I have no idea what `Slicer::MimicSource + 10` means, could you explain? Or maybe someone more familiar with lattices could have a look?

tammojan (approved):

PR #1302, opened 2023-02-24
Fix build warnings and indentation

gervandiepen:
A build warning said that std::move should be used at a few places.
This has been fixed as well as the indentation that was off at a few places.


aroffringa:
Hi @gervandiepen , would it be an idea to start using an automatic formatter like clang-format? It avoids all comments / edits addressing formatting issues and can be checked by the CI. We use it for many of the other repositories and it requires a bit getting used to but, once there, it saves a lot of time.

tammojan (approved):

gervandiepen:
Indeed one standard format would be nice. But I know that some people strongly prefer their own formatting style. My preference for one format is not too strong, so I propose to leave it as is.

PR #1309, opened 2023-05-09
add C-style header to fix GCC 13.1 compile error on uint16_t 

broekema:
It appears GCC 13.1 no longer includes standard C types and requires explicit inclusion of cstdint.


aroffringa (approved):

aroffringa:
Looks good, thanks

PR #1311, opened 2023-05-19
mirlib: fix minor compilation issues

cquike:
This is a small PR that prevents compilation issues within the mirlib directory when using gcc 8.5 and several -W options enabled like -Wmissing-declarations . The changes are:
- Defining _XOPEN_SOURCE to get the strdup declaration (man strdup)
- Defining _DEFAULT_SOURCE to get the sys_nerr declaration (man sys_nerr)
- Using I_ISDIR rather than I_IFDIR (https://www.gnu.org/software/libc/manual/html_node/Testing-File-Type.html)

### DIFF
--- a/mirlib/dio.c
+++ b/mirlib/dio.c
@@ -341,6 +341,7 @@ void dreaddir_c(char *contxt,char *path,int length)
     Strcpy(path,dp->d_name);
     Strcpy(npath,d->path);    Strcat(npath,path);
     (void)stat(npath,&buf);
-    if(S_IFDIR & buf.st_mode)Strcat(path,"/");
+    if(S_ISDIR(buf.st_mode)
+      Strcat(path,"/");
   }
 }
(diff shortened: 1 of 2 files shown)


tammojan (approved):

mirlib/dio.c
```diff
@@ -341,6 +341,7 @@ void dreaddir_c(char *contxt,char *path,int length)
     Strcpy(path,dp->d_name);
     Strcpy(npath,d->path);    Strcat(npath,path);
     (void)stat(npath,&buf);
-    if(S_IFDIR & buf.st_mode)Strcat(path,"/");
+    if(S_ISDIR(buf.st_mode))
+      Strcat(path,"/");
```
gmloose, line 345: Is this coding style allowed in `casacore`? I mean, having an `if` statement on multiple lines without using curly braces?

gmloose (changes requested):

mirlib/dio.c
tammojan: It's consistent with the rest of the file, that's about the code style we're aiming for. But I added curly braces here.

PR #1312, opened 2023-06-27
Correct allocator_is_always_equal type

sjperkins:
I discovered this while compiling casacore on c++17.  I suspect this doesn't work because `allocator_is_always_equal` is declared before `allocator_type`. The simplest fix seem to be using `Alloc` when specifying `allocator_is_always_equal`'s type (and which allocator_type is assigned to lower down).


sjperkins:
Other option might be to just move the definition of `allocator_is_always_equal` below the definition of `allocator_type`.

Issue #1314, opened 2023-07-25
[3.4.0] Failures on i386 when compiled with gcc 13


olebole:
Debian just switched to gcc version 13, and with this (after applying #1309) I see a number of failures on our i386 platform. Interestingly, they do not appear on other 32-bit platforms (ARM, x32):

```
The following tests FAILED:
	 25 - tInput (Failed)
	 67 - tQuantum (Failed)
	 69 - tQVector (Failed)
	109 - tConcatTable2 (Failed)
	190 - tRecordGram (Failed)
	219 - tFunctionHolder (Failed)
	255 - tBiweightStatistics (Failed)
	257 - tClassicalStatistics (Failed)
	258 - tFitToHalfStatistics (Failed)
	259 - tHingesFencesStatistics (Failed)
	385 - tLCSlicer (Failed)
	404 - tFITSImage (Failed)
Errors while running CTest
```

There are not much details on each failure, except for the tFitsImage. There, DATE-OBS is negative, and this seems to cause a segmentation fault:

```
2023-07-24 19:30:43	INFO	ImageFitsConverter::ImageHeaderToFITS 	Finding scaling factors for BITPIX=16 and look for masked or blanked values
2023-07-24 19:30:43	SEVERE	FITSDateUtil::toFITS (file ./fits/FITS/FITSDateUtil.cc, line 82)	unexpected problem converting time to FITS string - the resulting string is off by more than a day - 
2023-07-24 19:30:43	SEVERE	FITSDateUtil::toFITS (file ./fits/FITS/FITSDateUtil.cc, line 82)	The output date string is : -5877631-01-18T00:00:00.000000
2023-07-24 19:30:43	SEVERE	FITSDateUtil::toFITS (file ./fits/FITS/FITSDateUtil.cc, line 82)	please report this bug using bug().
2023-07-24 19:30:43	INFO	ImageFitsConverter::ImageToFITS 	Copying '/<<PKGBUILDDIR>>/obj-i686-linux-gnu/images/Images/test/imagetestimage.fits' to file All pixels fit in memory (8588 pixels).
2023-07-24 19:30:43	SEVERE		HDU error:  Errors in keyword list[HeaderDataUnit::init_data_unit]

FITSIO status = 107: tried to move past end of file
ffopen could not interpret primary array header of file: 
imagetestimage2.fits
2023-07-24 19:30:43	SEVERE		BlockIO:  File imagetestimage2.fits Physical record 0 logical record 0 --
[...]

PR #1315, opened 2023-07-25
Use CMAKE_CXX_STANDARD instead of manually implementing C++ standard selection

AdrianBunk:
This changes behaviour due to enabling compiler specific extensions by setting e.g. `-std=gnu++11` instead of `-std=c++11` when building with gcc. This might be considered desirable (see #1314) or not, in the latter case additionally `set(CMAKE_CXX_EXTENSIONS OFF)` would keep the `-std=c++11`.


aroffringa (approved):
Looks good to me, thanks! Do I understand correctly this also fixes the i386 build on Debian?

AdrianBunk:
> Do I understand correctly this also fixes the i386 build on Debian?

Yes.

aroffringa:
Fixes  #1314.

Issue #1307, opened 2023-04-19
Import FITS-IDI EOP information


kettenis:
Astrometric VLBI requires a very accurate correlator delay model. The calculation of the delays involve a terrestial-to-celestial transformation which depends on a number of parameters that describe the orientaton of the earth: the difference between UT1 and UTC, and the position of the celestial pole. These Earth Orientation Parameters (EOPs) can be predicted, but ultimately need to be measured for maximum accuracy. The most accurate measurements are only available after a few weeks and can differ significantly (for astrometric purposes) from the predictions. These final measurements may not be available at time of correlation, so data may be correlated with predicted or less accurate measurements. As long as these difference with the final values is small, phase corrections can be made after correlation. Code to apply these corrections is being developed for CASA.

To calculate these corrections we need to know the EOPs that were used for correlation. For the VLBA (and potentially other VLBI arrays that use the DiFX correlator) these values are available in the (nonstandard) CALC table in the FITS-IDI files. So we need to import these values and make them available in the MeasurementSet. A time series of EOPs is provided, typically extending two days before and after the observation such that accurate interpolation (in time) of the values is possible. This issue proposes a new MeasurementSet extension table (in the spirit of the GAIN_CURVE and PHASE_CAL tables that have been added somewhat recently). Feedback on the proposed table is welcome.

I will be implementing support for importing EOPs from FITS-IDI files in casacore based on the outcome if that discussion.

## EARTH_ORIENTATION

Name | Format | Units | Measure | Comments
---- | ---- | ---- | ---- | ----
TIME | Double | s | EPOCH | Time for the measurement/prediction
UT1_UTC | Double | s | | UT1 - UTC
PM | Double(2) | rad | | Position of celestial pole 

### TIME

Timestamp of the EOP values; this is expected to be the UTC time although some sources use TAI instead

### UT1_UTC

Difference between UT1 and UTC (UT1 - UTC) in seconds.

### PM

Position of celestial pole; these are provided as X and Y components (in that order)

Rationale: UT1_UTC and PM_x and PM_y are the parameters used in the IAU 2000 and 2006 precession and nutation models. Older models used more parameters. These models are no longer in widespread use in VLBI and we don't plan to implement support for these models in CASA.

## Further comments

It may be relevant to know whether the EOP values are predicted, intermediate or final.  This could be done in the form of a string with reserved keywords like what is done for mount types in the MS 2.0 standard.

kettenis:
[...]

PR #1318, opened 2023-09-26
Add support for Earth Orientation Parameters

kettenis:
See issue #1307

### DIFF
--- a/ms/MSOper/MSConcat.cc
+++ b/ms/MSOper/MSConcat.cc
@@ -358,6 +358,9 @@ IPosition MSConcat::isFixedShape(const TableDesc& td) {
   // PHASE_CAL
   copyPhaseCal(otherMS, newAntIndices);
 
+  // EARTH_ORIENTATION
+  copyEOP(otherMS);
+
   /////////////////////////////////////////////////////
 
   // copying all subtables over to otherMS
@@ -1246,6 +1249,11 @@ IPosition MSConcat::isFixedShape(const TableDesc& td) {
   // PHASE_CAL
   if(!copyPhaseCal(otherMS, newAntIndices)){
     log << LogIO::WARN << "Could not merge PhaseCal subtables " << LogIO::POST ;
+  }
+
+  // EARTH_ORIENTATION
+  if(!copyEOP(otherMS)){
+    log << LogIO::WARN << "Could not merge EOP subtables " << LogIO::POST ;
   }
 
   //////////////////////////////////////////////////////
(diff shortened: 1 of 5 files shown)


dpetry:
Hi Mark, 
so, this PR is adding a whole new optional sub-table to the MS!
As far as I can see, this new subtable is only documented in casacore issue #1307.
I wonder if at least a basic description of the table, i.e. one sentence about its purpose
plus the column descriptions, could also go into the code?
Also pointing out the "key" columns for accessing the table.

Secondly, of course it would be good to have some kind of unit test for 
writing and reading this table.

kettenis:
Hi Dirk,

Yes, I've followed the approach taken with the GAIN_CURVE and PHASE_CAL tables, which also have the table documented in issues. Maybe it it is a good idea to add those descriptions in somewhere in the repository itself. The issues use markdown, so copying the important bits out of the issue into a .md file is trivial. And github can render those properly. What would be an appropriate place for those files?

As for unit tests, there will be tests in CASA of course, and so far we've not added any FITS-IDI tests in casacore.  If we want that, we probably need two or three FITS-IDI files in the data repository that casacore uses for its unit tests.

Cheers,

Mark

ms/MSOper/MSConcat.cc
```diff
@@ -2387,6 +2395,65 @@ Bool MSConcat::copyPhaseCal(const MeasurementSet& otherMS,
...
+Bool MSConcat::copyEOP(const MeasurementSet& otherMS){
+  // uses newSPWIndex_p; to be called after copySpwAndPol
+
+  LogIO os(LogOrigin("MSConcat", "copyEOP"));
+
+  Bool itsEOPNull = (!itsMS.rwKeywordSet().isDefined("EARTH_ORIENTATION") || (itsMS.rwKeywordSet().asTable("EARTH_ORIENTATION").nrow() == 0));
+  Bool otherEOPNull = (!otherMS.keywordSet().isDefined("EARTH_ORIENTATION") || (otherMS.keywordSet().asTable("EARTH_ORIENTATION").nrow() == 0));
+
+  if(itsEOPNull && otherEOPNull){ // neither of the two MSs do have valid EOP tables
+    os << LogIO::NORMAL << "No valid EOP tables present. Result won't have one either." << LogIO::POST;
```
dpetry, line 2407: Should probably be done with LogIO::DEBUG since this is an optional table and we don't need to record that it is not there.

kettenis: That makes sense; used LogIO::DEBUG1 since LogIO::DEBUG doesn't exist and LogIO::DEBUG1 is what is used elsewhere in this file.

dpetry (approved):
OK, I don't want to hold up your development.
Clearly there are two nice-to-haves which you should consider adding at some point:
1) ca. half an A4 page worth of text describing the VLBI-related add-ons to the MS and
    from where they are typically filled, i.e. some kind of FITSIDI->MS mapping.
    Just the table names on both sides would already be helpful.
    Leaving this in some issue text is not good. These things get lost or at least are
    very difficult to find after a few years.
2) some rudimentary unit test within casacore

kettenis:
@tammojan Can this PR be merged?

Regarding documenting the VLBI extension tables, what would the best place to document those? Easiest would be if I could add a markdown file somewhere in the tree that documents them. Or perhaps a separate markdown file for each. The documentation in the issues is already markdown and github knows how to render them.

tammojan:
As discussed offline with @kettenis, I think the 'traditional' way to store casacore notes is on https://github.com/casacore/casacore-notes . Technically a markdown in this repository would also be good, but this addition is so small that I'd rather not change the way of working because of this. So @kettenis will (shall/should/may) add a note to casacore-notes.

tammojan (approved):

tammojan:
In fact, @kettenis you could make a pull request against casacore note 264 (the MSv3 draft).

kettenis:
> In fact, @kettenis you could make a pull request against casacore note 264 (the MSv3 draft).

@tammojan I think it makes sense to have it as a separate note. These are de-facto MSv2 extensions and I'm not sure to what extent the MSv3 draft is relevant now that MSv4 is being discussed... What I could do is have the actual table definitions in a separate .tex file that gets included in both the separate note and the MSv3 draft note.

PR #1319, opened 2023-09-26
Warn about old VLBA data that needs FFT artifact correction

kettenis:


kettenis:
Justin Linford from the VLBA told me is happy with the way this would warn (their) users about data that may lack certain corrections when imported into CASA.  So I'm removing draft status from this PR.

dpetry:
Sorry, didn't see this before. Will look at it in detail tomorrow.

dpetry:
@kettenis , hi, I had a first look. I don't think _all_ warnings should be reduced to normal infos.
I think those messages which are about deprecated or incorrectly used FITS-IDI features and similar, should remain
warnings or at least have some string in them which people can grep for.
But I'll give a complete review tomorrow ..

dpetry (changes requested):
I agree that all the "treating as scalar" warnings can be turned to INFO.
Also the Double should be Float can be INFO.
But the "not yet implemented" and "will ignore it" messages should stay
warnings such that people realise that some of the input is not making it 
into the MS.

kettenis:
@dpetry thanks for the review. Yes maybe I went a little bit too far in demoting the warnings got standard FITS-IDI tables that we don't handle yet. For the BASELINE, BANDPASS and CALIBRATION tables that is not an issue, since those tables are not present in typical archival data. I'm not even sure if code exists to produce these. For the INTERFEROMETER_MODEL your concern is certainly justified. Even though there is no code in CASA to use the correlator model, there is general consensus that this data should be added to the MS. A new table was proposed as part of MSv3 and is still on the table for MSv4.

dpetry (changes requested):
This looks good except for the very last message about the unconventional extension.
IMHO that should remain a warning since people should be told if their input
contains something which cannot be interpreted and is ignored.
(Sorry if I'm delaying you.)

kettenis:
@dpetry
Hi Dirk,

Unfortunately the VLBA data tends to have a few non-standard tables attached to it. Some of those are empty (no rows) or really are of no interest to the modern user (TAPE_STATISTICS). And these tables tend to come at the end so the warning you're talking about tends to come at the end of the output the user sees. The VLBA stakeholders are worried that this means users will miss the warnings.

How about printing that warning only if the number of rows in the table is greater than zero and exclude the TAPE_STATISTICS table?
```
      if (extname != "TAPE_STATISTICS" && nrows() > 0) {
        *itsLog << LogIO::WARN << "Extension " << extname
                << " not part of the FITS-IDI convention. Will ignore it." << LogIO::POST;
      }
```

dpetry:
@kettenis , yes, that's a great idea. Please go ahead with this.

kettenis:
@dpetry Thanks, done!

dpetry (approved):
Not sure what the TAPE_STATISTICS part is about but I trust this is fine.
Approved.

PR #1322, opened 2023-10-12
Fix compilation under gcc 13.2 due to undefined uint16_t

cquike:
This pull requests fixes compilation of casacore under Fedora 38 with gcc 13.2.1
The compilation error shown is like:
```
casacore/tables/Dysco/tests/../bytepacker.h:721:22: error: ‘uint16_t’ does not name a type
casacore/tables/Dysco/tests/../bytepacker.h:5:1: note: ‘uint16_t’ is defined in header ‘<cstdint>’; did you forget to ‘#include <cstdint>’?
```

### DIFF
--- a/tables/Dysco/bytepacker.h
+++ b/tables/Dysco/bytepacker.h
@@ -2,6 +2,7 @@
 #define DYSCO_BYTE_PACKER_H
 
 #include <stdexcept>
+#include <cstdint>
 
 namespace dyscostman {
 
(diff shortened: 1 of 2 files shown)


cquike:
Sorry, I have just seen that there is already a commit 68e43f489abd3d4f1e2fe54a42695396703aa81a in master that fixes that. We have a version of casacore in casa that is older than that commit.
Closing then.

PR #1327, opened 2023-11-07
Update macOS runners to macos-12

rtobar:
macos-10.15 isn't a supported flavour anymore, and thus actions are blocked from running, an eventually fail with a timeout after 24 hours.


rtobar:
Superseded by #1328

Issue #1331, opened 2023-12-01
Forward Declaration in Regex.h causes conflicts with google test.


preshanth:
The following two lines in `casa/Utilities/Regex.h` lead to a build issue for google test code which relies on `std::regex` . 
```
//# Forward declarations.                                                                                 
// struct re_pattern_buffer;                                                                              
// struct re_registers;          
```

So I have commented out the two lines built and made sure the same number of CTests are passing. Is there a specific reason for their existence or can we do without them?

In my fork 4 tests are failing due to python I believe but it is the same tests failing after the code has been commented out.
```
The following tests FAILED:
	142 - tTableUtil (Failed)
	156 - tDataManInfo (Failed)
	354 - tMSTimeGram (Failed)
	459 - tFITSImage (Failed)

```
I am happy to submit a PR for this from my fork.

tammojan:
Sorry it took a while, #1332 is now merged.

preshanth:
Thanks @tammojan  appreciate it.

PR #1332, opened 2023-12-01
Commented out the forward declaration of conflicting structs

preshanth:
Here is the PR with the fixed forward declaration.


preshanth:
Directly following from #1331

tammojan (approved):
Looks good, thanks. I removed the commented out lines altogether; we'll always have the git history.

PR #1333, opened 2023-12-15
Generate casarc file

gmloose:
Generate a casarc file that can be used by the user and copied to a different location. The user could then use the environment variable `CASARCFILES` to point to the desired casarc file.

### DIFF
--- a/measures/CMakeLists.txt
+++ b/measures/CMakeLists.txt
@@ -15,6 +15,12 @@ if (DATA_DIR)
     set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DCASADATA='\"${DATA_DIRX}\"'")
 endif (DATA_DIR)
 
+configure_file(
+    ${CMAKE_CURRENT_SOURCE_DIR}/casacore.rc.in
+    ${CMAKE_CURRENT_BINARY_DIR}/casacore.rc
+    @ONLY
+)
+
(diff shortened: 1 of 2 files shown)


tammojan:
With this pull request, the `@DATA_DIR@` specified to CMake gets both hard-coded in MeasIERS.cc, and in this generated `.casarc` file, which I think is not found by default. Couldn't you generate this .casarc file in the repository where you need this use case (I guess python-casacore)?
If you proceed with this pull request, could you update the docs in [measures/Measures/MeasIERS.h](https://github.com/casacore/casacore/blob/d7d5a127539dea0dd80110c95e4ae9c91a41435e/measures/Measures/MeasIERS.h#L64)?

gmloose:
No, that's not true. I didn't touch any of the logic used in `MeasIERS.cc`. This PR only ensures that a `casacore.rc` file is generated with the correct contents (based on the value of the CMake variable `DATA_DIR`), and is installed in `<prefix>/etc`. This as a courtesy to the user. But I'm happy to ditch this PR if you think it doesn't add much.

aroffringa:
If a data dir is specified, the data dir already gets hardcoded into casacore, doesn't it? It's not necessary to have a casacore.rc file in that case, as casa already finds the dir... so I think this PR may be redundant now. Shall I close it @gmloose ?

PR #1334, opened 2024-01-16
Fix building of docker images

gmloose:
Docker images used to be built on quay.io, but that doesn't seem to work anymore, since images haven't been built for quite a while.

Therefore, it was decided to create a GitHub workflow instead for building these Docker images: i.e. build on GitHub and push to quay.io.


tammojan (approved):
Nice improvement! Please make sure to squash the commits before merging.

PR #1339, opened 2024-03-21
Change aips2-request@nrao.edu to casa-feedback@nrao.edu

vsuorant:
Casa would like to change the contact email from aips2-request@nrao.edu to casa-feedback@nrao.edu. Let me know if there is other email address that would be a better fit.


schiebel (approved):
Only comment changes

PR #1341, opened 2024-03-25
Make export to UVFits work with non-constant receptor angles

### LINKED ISSUE #1340: MSFits: non-constant receptor angles in MS FEED table prevent export to UV FITS
In an ALMA helpdesk ticket, a user reported a case where an MS contained non-constant receptor angles in the FEED table and could therefore not be exported.
These angles are only relevant for calibration. If the date is already calibrated, which is usually the case before export to UV FITS, the angles are irrelevant.
I propose to relax the conditions of the code and make it only produce a warning when non-constant (in time) receptor angles are encountered.

tammojan:
Implemented by Dirk Petry (@petry)

Fixes #1340

### DIFF
--- a/msfits/MSFits/MSFitsOutput.cc
+++ b/msfits/MSFits/MSFitsOutput.cc
@@ -1850,11 +1850,11 @@ Bool MSFitsOutput::_writeAN(std::shared_ptr<FitsOutput> output, const Measuremen
                 // inicates that row applies for all spectral windows
                 if (
                     Int(antnum) == inantid(i)
-                    && (
-                        spwids(i) == -1
-                        || uSpws.find(spwids(i)) != uSpws.end()
-                    )
-                ) {
+                      && (
+                          spwids(i) == -1
+                          || uSpws.find(spwids(i)) != uSpws.end()
+                      )
+                    ) {
                     found = True;
                     Vector<String> poltypes = inpoltype(i);
                     Vector<Quantity> ra;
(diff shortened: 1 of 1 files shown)


dpetry:
@tammojan , looks like the MacOS failure is not related to the actual code change ...

tammojan:
> @tammojan , looks like the MacOS failure is not related to the actual code change ...

Agree. Not the first time we have to babysit the MacOS build. In the meantime, this build failure should not delay fixing your issue. Do you have a volunteer to review this change?

dpetry:
@kettenis , hi, would you mind reviewing this little adjustment? It downgrades the reaction to a potential problem
in the MS-to-UVFITS (not FITS-IDI) translation of the FEED table from an exception to a Warning.
The warning text explains it.

dpetry:
@tammojan , not sure if Mark (Kettenis) is actually entitled to do a review. Is he? If not, could you pick a different reviewer?

msfits/MSFits/MSFitsOutput.cc
```diff
@@ -1871,10 +1871,19 @@ Bool MSFitsOutput::_writeAN(std::shared_ptr<FitsOutput> output, const Measuremen
                         antToRA[antnum] = ra;
                     }
                     else {
-                        _checkReceptorAngles(ra, antToRA[antnum], antnum);
+                        try {
+                            _checkReceptorAngles(ra, antToRA[antnum], antnum);
+                        }
+                        catch (const std::exception& x) {
+                            os << LogOrigin("MSFitsOutput", __func__)
+                               << LogIO::WARN << "Receptor Angles in FEED table are not constant for antenna " << antnum
```
apschoenmakers, line 1879: constant in time?

apschoenmakers (approved):

PR #1346, opened 2024-05-02
Modernize the Constants.h file and switch to c++17

aroffringa:
The Constants.cc fails to compile on my compiler (gcc 13.2), because it complaints that the constants should be declared inside namespace C. Looking at the file it seems to contain a lot of constants that are declared nowadays in the standard C++ headers. I propose to make the duplicates deprecated.

Also, they can be constexpr which will potentially allow more functions to become constexpr. This however requires C++17. Since modern compilers compile in C++17 by default nowadays, this might be a good moment to switch (for discussion). It eliminates a .cc file.


dpetry:
@aroffringa , hi, I would love to do my share and review this but it seems this has wider implications.
Has it been decided that all casacore users switch to C++17?
Besides, there seem to be several failed tests.
Looks like it is too early to merge this.

aroffringa:
Hi @dpetry , thanks for the response. From the ASTRON side, requiring C++17 is not a problem: all our repositories that use casacore are already requiring C++17 (and we're just migrating the first repository to C++20 as a test). I don't know if this is the case from the casa side of course. At ASTRON we had at some point a discussion about when to switch where, and we decided to skip C++14 because the number of compilers that support 14 but not 17 is almost none, so after migrating to 11 (which was a bit more painful) we switched reasonably quickly to 17 (which was really simple). 

We had a recent release of casacore, so most users that use older systems should be able to use that release when they can't build with 17. 

Maybe this MR needs to be split into some parts, e.g. first requiring 17, then replacing a few constants. This might make it easier to test things -- particularly for casa, if someone can do that. Happy to do so if you think that's useful.

dpetry:
@aroffringa , I think @vsuorant should be consulted concerning the CASA side.

vsuorant:
I will try this branch out soon-ish (hopefully only a few days). I believe we are at a point where this shouldn't be a huge issue for CASA. On a related note: We do will likely need a Casacore branch to support our LTS release. I think the branch point has to be before this gets merged.

vsuorant:
I did test building the casa6 repository and our UIs with these changes. All in all the changes required on the Casa side are fairly small, so I think this should be fine to merge. Before you do though, can you please create a Casacore branch called casa-6.6.6 from commit 73f74f580a2e6e04b2a967c97736243acdaf1adc ? release/6.6.6 is the basis for our next Pipeline release. We may still need Casacore patches for it, but we do not want this update included.

aroffringa:
I've made the branch as requested: https://github.com/casacore/casacore/tree/casa-6.6.6

tammojan (commented):
One minor nit, otherwise this looks like a good step into the 2020s!

tammojan (approved):

PR #1355, opened 2024-06-28
Replace insecure sprintf by snprintf

gervandiepen:
clang warned that the insecure sprintf should be replaced by snprintf.
It was easy to do; only fits.h/cc required a bit more work.

### DIFF
--- a/fits/FITS/fits.h
+++ b/fits/FITS/fits.h
@@ -993,9 +993,9 @@ class FitsKeyCardTranslator {
         static void fmtcard(char *, const FitsKeyword &);
     private:
 	int cardno;		// the current card number within record
-	const int FitsCardSize;
-	const int FitsMaxCard;
-	const int FitsRecSize;
+	static const int FitsCardSize = 80;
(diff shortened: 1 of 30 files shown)


fits/FITS/fits.h
```diff
@@ -993,9 +993,9 @@ class FitsKeyCardTranslator {
         static void fmtcard(char *, const FitsKeyword &);
     private:
 	int cardno;		// the current card number within record
-	const int FitsCardSize;
-	const int FitsMaxCard;
-	const int FitsRecSize;
+	static const int FitsCardSize = 80;
+	static const int FitsMaxCard = 36;
+	static const int FitsRecSize = 2880;
```
tammojan, line 998:
```suggestion
	static constexpr int FitsCardSize = 80;
	static constexpr int FitsMaxCard = 36;
	static constexpr int FitsRecSize = 2880;
```

tammojan (approved):
Looks good to me! I have not tested whether you've removed all sprintf's, but I assume you did a grep.
The constexpr change is optional.

PR #1357, opened 2024-08-11
Updating bin to CMAKE_INSTALL_BINDIR

preshanth:
I have updated all references of install and runtime directories to point to `CMAKE_INSTALL_BINDIR` which points to a path the user sets else will default to gnuinstalldirs i.e. install/bin


rtobar:
@preshanth I'm not a casacore maintainer myself, so I can't approve or reject this PR. I wanted to comment though on two points.

I tried your patch locally and I still see two executables under `$PREFIX/bin/`, namely `findmeastable` and `measuresdata-update`. They both are defined under `measuers/apps/CMakeLists.txt`, you might want to make sure they are included in your patch.

The second point is more interesting. I think that a better solution for the problem might be *not* to specify a `DESTINATION` value in the first place. From the install command [docs](https://cmake.org/cmake/help/latest/command/install.html#signatures):

> For regular executables, static libraries and shared libraries, the DESTINATION argument is not required. For these target types, when DESTINATION is omitted, a default destination will be taken from the appropriate variable from [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#module:GNUInstallDirs), or set to a built-in default value if that variable is not defined.

(with the built-in default being `bin` for `RUNTIME` destinations)

In other words, removing the `DESTINATION` option from the `install` commands you are concerned about will give us the best of both worlds: we get `bin` by default, and you can still indicate a different directory via `CMAKE_INSTALL_BINDIR`. In the case of the `measuresdata-update` `PROGRAMS` installation, it might be better to specify `TYPE BIN` rather than a `DESTINATION` option,

preshanth:
@rtobar Thanks for pointing out the two instances I have missed. I had them fixed locally but had committed before looks like. 

I fully understand the default install option is the `bin` based on `GNUInstalldirs`, I use the same in my code. The issue however is that casacore dumps a bunch of executables into `bin`. I only want the executables I write to exist in my bin. Here is an example of our current structure where all other dependencies put executables into `install/bin/XXX/` where XXX is the external dependency.
```
nemo:  ls -l
total 1223600
-rwxr-xr-x 1 pjaganna nmstaff    761384 Aug  2 14:57 acme
drwxr-xr-x 2 pjaganna nmstaff        28 Aug  5 11:45 bin
drwxr-xr-x 2 pjaganna nmstaff      4096 Aug  1 01:45 casacore
-rwxr-xr-x 1 pjaganna nmstaff   2279008 Aug  2 14:57 coyote
-rwxr-xr-x 1 pjaganna nmstaff   5224640 Aug  2 14:57 dataiter
-rwxr-xr-x 1 pjaganna nmstaff       442 Aug  6 11:30 exodus
drwxr-xr-x 2 pjaganna nmstaff        64 Aug  1 01:44 fftw
drwxr-xr-x 2 pjaganna nmstaff        33 Aug  1 01:44 hpg
-rwxr-xr-x 1 pjaganna nmstaff   1471528 Aug  2 14:57 hummbee
drwxr-xr-x 2 pjaganna nmstaff        87 Aug  1 01:44 kokkos
-rwxr--r-- 1 pjaganna nmstaff 618866854 Aug  6 11:35 libra_apps.sh
-rwxr--r-- 1 pjaganna nmstaff 618866928 Aug  5 12:12 libra.sh
-rwxr-xr-x 1 pjaganna nmstaff    810520 Aug  2 14:57 mssplit
-rwxr-xr-x 1 pjaganna nmstaff   3105560 Aug  2 14:57 roadrunner
-rwxr-xr-x 1 pjaganna nmstaff    894384 Aug  2 14:57 subms
-rwxr-xr-x 1 pjaganna nmstaff    657936 Aug  2 14:57 tableinfo
```
My external dependencies are casacoe, hpg, fftw and kokkos all of which have executables they provide which are installed optionally inside the `bin` when an install `dir` is not specified. The changes I have made are in the same spirit. If a user were to provide a `CMAKE_INSTALL_BINDIR` it would then end up there else will default to the `GNUInstalldirs`. That clarifies my approach ?

rtobar:
@preshanth yes, the requirement is clear, and I understand your approach. What I'm pointing out is that you get the same behaviour if you don't specify a `DESTINATION`: you can _still_ define `CMAKE_INSTALL_BINDIR` and it will do what you need. That's how I spotted the two missing executables: in my test I defined `-DCMAKE_INSTALL_DIR=bin/casacore` and found the two lingering files directly under `bin`.

To be clear, your approach also works, it just seems unnecessary as you're mimicking the default behaviour provided by cmake already.

BTW, I realise I forgot to point to the changes I came up with (on top of yours), maybe if you try them out it will give more confidence that things work as expected without explicitly stating a `DESTINATION`. Changes are at https://github.com/rtobar/casacore/commit/a8b75474196c466f1fe95823e89c011bc1567cd5

preshanth:
@rtobar The issue is following your approach you get `/bin/casacore/bin` and `/bin/casacore/lib` and `/bin/casacore/include` so on. I would rather have a singular directory for the libraries and subdivision only inside the `bin` dir. Hence the `-DCMAKE_INSTALL_BINDIR` which I think is the more general solution offered by `cmake`. Further more I think `casacore` needs and `install/lib/cmake/casacoreConfig.cmake` too. I am happy to put one of those in too should the admins agree.

rtobar:
@preshanth sorry, I mistyped in my previous comment. I meant I used `-DCMAKE_INSTALL_BINDIR=bin/casacore`. If you could give a try to my changes that'd confirm this works as intended.

preshanth:
@rtobar That works very well. I see the difference that your change has. It is a much nicer solution. Please push you patch on here. Else I am happy to do that too . Lets ask the admins to take in the change.

rtobar:
@preshanth glad to hear that! I can't push to your repo, but you should be able to merge the commit in my repo into yours. If you're not sure how to do that you can follow https://github.com/preshanth/casacore/compare/master...rtobar:casacore:master to create a PR and do the merge that way, which should then be reflected here.

preshanth:
@rtobar done merged.

tammojan:
@rtobar Could you confirm that the MR as it is now does the right thing? Then I'm happy to merge.

tammojan:
I see that CMake in the CI fails with [errors](https://github.com/casacore/casacore/actions/runs/10574321844/job/29295616640?pr=1357#step:3:1453) on all executables:
```
#12 2.804 CMake Error at tables/apps/CMakeLists.txt:5 (install):
#12 2.804   install TARGETS given no RUNTIME DESTINATION for executable target
#12 2.804   "showtableinfo".
#12 2.804 
#12 2.804 
#12 2.804 CMake Error at tables/apps/CMakeLists.txt:5 (install):
#12 2.804   install TARGETS given no RUNTIME DESTINATION for executable target
#12 2.804   "showtablelock".
#12 2.804 
#12 2.804 
#12 2.804 CMake Error at tables/apps/CMakeLists.txt:5 (install):
#12 2.804   install TARGETS given no RUNTIME DESTINATION for executable target "taql".
#12 2.804 
```
Could this be due to a CMake version dependency? I think CI uses CMake 3.10.2-1ubuntu2.18.04.2.

rtobar:
Ugh , yes, I hadn't realised this functionality was not there in earlier cmake (I didn't see any "New in version X.Y" anywhere, so I assumed it was introduced in 3.0, maybe even earlier). This was indeed introduced [in 3.14](https://cmake.org/cmake/help/latest/release/3.14.html#commands), explaining why it fails on CI.

I'm not sure what consideration would you give to updating the minimum cmake requirement? It's a broader question for sure, but 3.10 is 7 years old, while 3.14 is 5 (and neither of them is actively maintained). The top-level `CMakeLists.txt` states 3.3.0 as the minimum cmake required version, which I'm not sure is actually true or not.

If you want to keep using 3.10 and keep these changes, this PR will need further work. If on the other hand you're willing to update cmake to at least 3.14 then this PR needs to wait for that to happen first.

tammojan:
CMake 3.14 is a very reasonable update. I will remove ubuntu 18.04 from the CI in a separate MR (#1361). Could you bump the CMake requirement in this MR?

rtobar:
@preshanth could you please merge from my latest `master` again so that we bump the minimum cmake version required to build casacore as requested by @tammojan? Thanks!

preshanth:
I have pulled and merged. This is the expected new behavior with cmake where providing a `-DCMAKE_INSTALL_BINDIR=/bin/casacore` results in all the casacore binaries being placed inside it.

preshanth:
@tammojan and @rtobar let me know if there is anything else I can do here.

tammojan (approved):

PR #1362, opened 2024-08-27
Install newer numpy for python wheels

tammojan:
When compiling with `oldest_supported_numpy`, the resulting package is incompatible with numpy 2.0. Using the latest numpy should get us numpy 2.0 compatibility.


gmloose (approved):
LGTM.

PR #1366, opened 2024-09-03
Fix library name in FindSOFA.cmake

mpokorny:
The installed SOFA library is normally named `libsofa_c.a`. The alternative name is retained to allow backwards compatibility (if in fact the other name ever worked -- I have not searched SOFA archives to determine if the other name was valid at some point in the past.)

### DIFF
--- a/cmake/FindSOFA.cmake
+++ b/cmake/FindSOFA.cmake
@@ -9,7 +9,7 @@
 
 if(NOT SOFA_FOUND)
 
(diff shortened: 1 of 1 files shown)


mpokorny:
I suppose distros might rename the library -- I haven't checked that -- but `libsofa_c.a` is the name when it's built from source.

mpokorny:
Please ignore this PR. I didn't know it before, but the dependency for casacore is on the FORTRAN library and API, not the C version.

PR #1370, opened 2024-09-26
Properly reset class variables in FITS-IDI filler

kettenis:
In particular, clear the array name when processing the primary HDU. This makes sure converting multiple FITS-IDI files from different arrays will correctly identify the array and make the appropriate digital corrections for the various array/correlator combinations. Change the warning about conflicting observatory names to SEVERE as this now only applies to conflicting names within a single FITS-IDI file, which would be serious trouble as the wrong digital corrections would probably be applied.  This should not happen in real life.


dpetry (approved):
Looks correct. Thanks.

PR #1371, opened 2024-12-02
Issue 1369 MSMetaData ALMA specific update

dmehring:
CASA JIRA ticket CAS-14435 Update msmd.almaspws for edge case of ACAC 2 GHz spws with heavy online channel averaging


ms/MSOper/MSMetaData.h
```diff
@@ -1000,6 +1001,10 @@ class MSMetaData {
 
     vector<std::set<String> > _getSpwToIntentsMap();
 
+    // poarization ids will be sorted in ascending order in all
```
tammojan, line 1004:
```suggestion
    // polarization ids will be sorted in ascending order in all
```

ms/MSOper/MSMetaData.h
```diff
@@ -769,6 +769,7 @@ class MSMetaData {
     mutable std::map<ScanKey, std::set<Int> > _scanToStatesMap, _scanToFieldsMap, _scanToAntennasMap;
     mutable std::map<Int, std::set<Int> >    _fieldToStatesMap, _stateToFieldsMap, _sourceToFieldsMap;
     mutable std::map<std::pair<uInt, uInt>, uInt> _spwPolIDToDataDescIDMap;
+    mutable std::vector<std::vector<uInt>> _spwIDToPolIDMap;
```
tammojan, line 772: The name implies that this is a map, why not use an `std::map` instead?

ms/MSOper/MSMetaData.cc
```diff
@@ -4999,6 +4999,21 @@ std::shared_ptr<vector<int>> MSMetaData::_almaReceiverBands(uint nspw) const {
...
 }
 
+std::vector<std::vector<uInt>> MSMetaData::_getSpwToPolMap() const {
+    if (! _spwIDToPolIDMap.empty()) {
+        return _spwIDToPolIDMap;
+    }
+    const auto spwPolToDDID = getSpwIDPolIDToDataDescIDMap();
+    _spwIDToPolIDMap.resize(nSpw(true));
+    for (const auto &p : spwPolToDDID) {
+        _spwIDToPolIDMap[p.first.first].push_back(p.first.second);
```
tammojan, line 5009: Would be a bit more readable as
```suggestion
        uInt spwID = p.first.first;
        uInt polId = p.first.second;
        _spwIDToPolIDMap[spwId].push_back(polId);
```

ms/MSOper/MSMetaData.cc
```diff
@@ -5117,7 +5132,32 @@ vector<MSMetaData::SpwProperties>  MSMetaData::_getSpwInfo2(
             fdmSpw.insert(i);
         }
         else {
-            tdmSpw.insert(i);
+            const auto spwToPol = _getSpwToPolMap();
+            if (
+                spwToPol[i].size() == 1
+                && spwInfo[i].nchans
+                    * getNumCorrs()[*(spwToPol[i].begin())] > 256
```
tammojan, line 5139: Am I correct in assuming that `tdmSpw` and `fdmSpw` stand for time and frequency domain spectral window? And what is the magic number 256 here?

tammojan (commented):
Looks good to me, some questions.

ms/MSOper/MSMetaData.h
dmehring: The spw id is represented by the index of the (outer) vector. This is a common pattern used in this class, and often the variable name includes "Map" which refers the the index being one side of the map.

ms/MSOper/MSMetaData.cc
dmehring: Yes that's what the acronyms mean. The 256 is the number given to me by the stakeholder who requires this change. There is some complexity in this corner case that requires it. I cannot give much of a detailed explanation beyond that; I'm just the guy who codes the requirements I'm given :) .

dmehring:
@tammojan it looks like some CI runs failed or were cancelled for issues not related to my changes.

tammojan:
> @tammojan it looks like some CI runs failed or were cancelled for issues not related to my changes.

No idea what happened there. I restarted them just now.

vsuorant:
It seems like maybe a compatibility issue with Boost and CMake versions?  That seems odd since the workflow configuration sets "-DBoost_NO_BOOST_CMAKE=True" here 
https://github.com/casacore/casacore/blob/master/.github/workflows/osx.yml#L51C13-L51C40

PR #1373, opened 2024-12-04
correct proper motion unit conversion in MSFitsOutput.cc

dpetry:
Check the present units and if they are RAD/S, convert to DEG/DAY, otherwise write the values as they are and give warning if the original units were not DEG/DAY.

### DIFF
--- a/msfits/MSFits/MSFitsOutput.cc
+++ b/msfits/MSFits/MSFitsOutput.cc
@@ -2121,8 +2121,7 @@ Bool MSFitsOutput::_writeSU(std::shared_ptr<FitsOutput> output, const Measuremen
                             *lsrvel = sv(0);
                         }
                     }
-                    if (
-                        sourceTable->isColumn(MSSource::REST_FREQUENCY)
+                    if (sourceTable->isColumn(MSSource::REST_FREQUENCY)
                         && sourceColumns->restFrequency().isDefined(rownr)
                     ) {
                         Vector<Double>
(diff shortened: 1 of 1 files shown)


dpetry:
@tammojan , hi, I guess the Mac OS build failures are some infrastucture problem, right?

tammojan:
> @tammojan , hi, I guess the Mac OS build failures are some infrastucture problem, right?

Yes they are. We'll fix this, and until then I think we can ignore macos build failures as long as we don't see obvious mac related issues in the code change.

msfits/MSFits/MSFitsOutput.cc
```diff
@@ -2132,10 +2131,25 @@ Bool MSFitsOutput::_writeSU(std::shared_ptr<FitsOutput> output, const Measuremen
...
+			else{
+			    *pmra = pm(0);
+			    *pmdec = pm(1);
+			    if (unit.at(0,5)!="DEG/D"){
+			       os << LogIO::WARN
+				  << "Proper motion for source in SOURCE table row #"<<rownr<<" has unfamiliar units: "
+				  << unit << " .\n Value in uv fits table AIPS SU needs to be checked."
+				  << LogIO::POST;
+			    }
+			}
```
tammojan, line 2151: Why not use casacore's built-in unit conversion routines? There's probably a bug in the code below, if you want to use it please try it locally.
```suggestion
			ArrayQuantColumn<Double> pm = sourceColumns->properMotionQuant()(rownr);
			try {
			    *pmra = pm_quantity(0).getValue("deg/d", true);
			    *pmdec = pm_quantity(1).getValue("deg/d", true);
			} catch (const std::exception& e) {
			       os << LogIO::WARN
				  << "Proper motion for source in SOURCE table row #"<<rownr<<" has unfamiliar units: "
				  << unit << " .\n Value in uv fits table AIPS SU needs to be checked."
				  << LogIO::POST;
		             *pmra = pm(0).getValue();
			     *pmdec = pm(1).getValue();
			}
```

dpetry: Yes, I was also thinking about this but was not sure the casacore routines would do the right thing in all cases.
The only use-case I know of is the rad/s -> deg/day conversion.
But OK, l will try your code and, if it works, make a new commit. 
Thanks!

dpetry:
@tammojan , OK, I modified the code to use Quanta now. Looks a bit different from what you originally suggested but it works.

PR #1374, opened 2024-12-05
Support compilation with c++23 turned on

aroffringa:
Due to change of rules in implicit move in C++23, at least with GCC (14) there are a few lines that no longer compiled. This has to do with the move-only class ostringstream, which seem to hide the conversion constructor of String.

The changes are of course backward compatible, so that c++23 is not required.


tammojan (approved):
Looks good to me, thanks!

PR #1375, opened 2024-12-12
Remove redundant CMake Find modules

gmloose:
A number of CMake Find modules have been removed from this repository, because all these modules are part of CMake >= 3.12. The ones in this repository would take precedence, but are outdated, and should not be used.


tammojan:
Thanks for cleaning up! There is [one recent change](https://github.com/casacore/casacore/commit/d48a880d8c72c5f4b8ce7dea91db3aa65956a9d7) in casacore's HDF5 module, could you perhaps check if that is also done properly in CMake's FindHDF5?

gmloose:
`FindHDF5.cmake` has changed so much compared to the old one in Casacore that I think it is more efficient to just test if it still works. I don't see a MacOS-12 build in the list anymore. Both existing MacOS builds are failing for a different reason. How would you like to proceed?

tammojan:
It is fine with me to proceed with a failing macos build currently. I think this change is also sufficiently tested that we can trust that CASA is not affected too much.

gmloose:
So, can you please approve? Otherwise I need to use my cow powers to bypass branch protections.

tammojan (approved):
🐮

PR #1379, opened 2025-01-11
Solve some warnings & undefined behaviour

aroffringa:


casa/OS/Memory.h
```diff
@@ -110,17 +110,20 @@ class Memory
     // How much memory has been allocated by the programmer (via either
     // <src>new</src> or <src>malloc</src>. This can include some extra
     // overhead bytes, e.g. for alignment reasons.
-    static size_t allocatedMemoryInBytes();
+    [[deprecated("Results are not accurate (see help for mallinfo2()), so avoid this call")]]
+    static size_t allocatedMemoryInBytes() { return 0; }
```
tammojan, line 114: Instead of returning 0, shouldn't we raise an exception upon use, or just remove the function?

lattices/LatticeMath/LatticeConvolver.tcc
```diff
@@ -52,7 +52,7 @@ LatticeConvolver()
    itsPsf(0),
    itsCachedPsf(False)
 {
-  itsXfr->set(typename NumericTraits<T>::ConjugateType(1));
+  //itsXfr->set(typename NumericTraits<T>::ConjugateType(1));
```
tammojan, line 55: Slightly dubious to just comment this line out? What does it do?

casa/OS/Memory.h
aroffringa: I thought that this value is probably only useful to report to the user.  If e.g. casa uses it for that, it's probably better to show '0' with a compilation warning, then to actually crash...

lattices/LatticeMath/LatticeConvolver.tcc
aroffringa: My best estimate after reading the code, is that the code is never used. itsxFr is a nullptr (4 lines above) so this is undefined behaviour (dereferencing a nullptr with -> is undefined behaviour in C++).

itsxFr is of type TempLattice<typename NumericTraits<T>::ConjugateType>*. The set function is a virtual function:
```
  // Set all of the elements in the Lattice to the given value.
  virtual void set (const T& value);
```
It overrides this function from the Lattice class. Since the "this" is nullptr, at runtime this virtual function call can not succeed without something going seriously wrong, most likely a segfault -- but newer compilers recognize undefined behaviour and may do really confusing things (like conclude that function can never be called, so can be removed entirely). I can not see how this constructor can ever work... But since I do not understand one bit of this code I just commented it out, showing a trace of what happened. If you have a better solution let me know ;).

aroffringa: Digging a bit deeper, the code is there since 2004:
```
f2128823a0 code/lattices/implement/Lattices/LatticeConvolver.cc (Ralph Marson     1997-11-20 23:32:00 +0000  54) {
2f60c29757 code/lattices/implement/Lattices/LatticeConvolver.cc (Ger van Diepen   2004-10-14 12:11:24 +0000  55)   itsXfr->set(typename NumericTraits<T>::ConjugateType(1));
625de99fc1 code/lattices/implement/Lattices/LatticeConvolver.cc (Kumar Golap      2003-08-13 00:13:45 +0000  56)   doFast_p=False;
f2128823a0 code/lattices/implement/Lattices/LatticeConvolver.cc (Ralph Marson     1997-11-20 23:32:00 +0000  57) } 
```
...and this particular constructor IS actually tested, in tLatticeConvolver.cc:55. The test also runs... fine :-|.

aroffringa: So, the test indeed crashes:
```
:~/projects/casacore/build/lattices/LatticeMath/test$ ./tLatticeConvolver 
Segmentation fault
```
...but inside ctest it runs fine:
```
build$ ctest -R tLatticeConvolver
Test project /home/offringa/projects/casacore/build
    Start 399: tLatticeConvolver
1/1 Test #399: tLatticeConvolver ................   Passed    0.04 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.06 sec
```
Adding -V to ctest gives some more info:
```
$ ctest -V -R tLatticeConvolver
[...]
399: Test command: /home/offringa/projects/casacore/cmake/cmake_assay "./tLatticeConvolver"
399: Working Directory: /home/offringa/projects/casacore/build/lattices/LatticeMath/test
399: Test timeout computed to be: 1500
399: UNTESTED: ./tLatticeConvolver
1/1 Test #399: tLatticeConvolver ................   Passed    0.02 sec

The following tests passed:
	tLatticeConvolver
```
apparently ctest only calls `/home/offringa/projects/casacore/cmake/cmake_assay "./tLatticeConvolver"`, which... is fine with tLatticeConvolver giving a segfault?

aroffringa: ..and then the final bit why the test doesn't run:
```
$ cat lattices/LatticeMath/test/tLatticeConvolver.run
#! /bin/sh
echo "UNTESTED"
exit 3
```
It has a `.run`  that does not run the tests. So the lattice convolver test never runs. Even after removing the obviously wrong nullptr dereferencing, the test still crashes when run manually. So, this seems like dead code, and maybe the whole latticeconvolver class is unused? At least within casacore the file isn't included.

So far that commenting out the warning-causing line seems fine to me.

dmehring:
It looks to me like CASA doesn't call the default constructor in which it appears this issue lies. I'll need to investigate further to be certain that removing the default constructor implementation will not cause the CASA build to fail. Will report back

kgolap:
casa does use LatticeConvolver in deconvolution code mainly. This is usually used when memory is not available to do convolution in Arrays etc... As @dmehring states we always construct with a Lattice as parameter.

dmehring:
Thanks for the confirmation @kgolap. I therefore suggest
```
LatticeConvolver() = delete;
```
in LatticeConvolver.h, removing the default constructor implementatioin in LatticeConvolver.tcc, and removing the test code in tLatticeConvolver.cc which depends on the default constructor (essentially the first block of code in the first try block).

aroffringa:
Thanks, changed as suggested by dmehring. The test still fails, also without the default constructor test code. The test is however still ignored so the CI is not failing because of it.

tammojan (approved):
Looks good to me! Thanks for the quick replies about CASA use all.

PR #1381, opened 2025-01-12
Add tests for the columnar file classes

aroffringa:
This PR merges towards the stokes-storage-manager branch and is part of several PRs that together will add this stman. The work is split in several PRs to allow easier reviewing.

This adds tests for https://github.com/casacore/casacore/pull/1380.

### DIFF
--- a/tables/AlternateMans/test/tColumnarFile.cc
+++ b/tables/AlternateMans/test/tColumnarFile.cc
@@ -16,9 +16,9 @@ void TestEmptyConstructor() {
   const std::string filename = "columnar_file_test.tmp";
   ColumnarFile file_b = ColumnarFile::CreateNew(filename, 0, 0);
   file_b = std::move(file_a);
-  BOOST_CHECK_EQUAL(file.NRows(), 0);
-  BOOST_CHECK_EQUAL(file.Filename(), "");
-  BOOST_CHECK(!file.IsOpen());
+  BOOST_CHECK_EQUAL(file_b.NRows(), 0);
+  BOOST_CHECK_EQUAL(file_b.Filename(), "");
+  BOOST_CHECK(!file_b.IsOpen());
 
   unlink(filename.c_str());
 }
--- a/tables/AlternateMans/test/tColumnarFile.cc
+++ b/tables/AlternateMans/test/tColumnarFile.cc
@@ -79,9 +79,9 @@ void TestReadAndWrite() {
   BOOST_CHECK_EQUAL(file.Stride(), 17*8);
 
   // Generate random data
-  for(size_t i=0; i!=std::size(data_a); ++i)
+  for(size_t i=0; i!=data_a.size(); ++i)
       data_a[i] = std::complex<float>(i+3, i*2);
-  for(size_t i=0; i!=std::size(data_b); ++i)
+  for(size_t i=0; i!=data_b.size(); ++i)
       data_b[i] = std::complex<float>(i*3+41, i*-2.0);
 
   file.Write(2, 0, data_a.data(), data_a.size());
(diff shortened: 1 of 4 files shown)


tables/AlternateMans/test/tColumnarFile.cc
```diff
@@ -0,0 +1,164 @@
...
+void TestEmptyConstructor() {
+  ColumnarFile file;
+  BOOST_CHECK_EQUAL(file.NRows(), 0);
+  BOOST_CHECK_EQUAL(file.Filename(), "");
+  BOOST_CHECK(!file.IsOpen());
+
+  file = ColumnarFile();
+  BOOST_CHECK_EQUAL(file.NRows(), 0);
+  BOOST_CHECK_EQUAL(file.Filename(), "");
+  BOOST_CHECK(!file.IsOpen());
```
mnijhuis-tos, line 19: This second part of the test checks that assigning an empty ColumnarFile to an existing empty ColumnarFile yields an empty ColumnarFile. I don't see much value in this part. If you want to test assignment, please use two different objects in the assignment.

tables/AlternateMans/test/tColumnarFile.cc
```diff
@@ -0,0 +1,164 @@
...
+  BOOST_CHECK_EQUAL(file.Filename(), filename);
+  BOOST_CHECK(file.IsOpen());
+  file.AddRows(43);
+  BOOST_CHECK_EQUAL(file.NRows(), 43);
+  file.DeleteRow();
+  BOOST_CHECK_EQUAL(file.NRows(), 42);
+  BOOST_CHECK(file.IsOpen());
+  // Close by assigning to empty
+  file = ColumnarFile();
+  BOOST_CHECK(!file.IsOpen());
```
mnijhuis-tos, line 48: Here, you're just checking that calling `IsOpen()` after using default constructor yields `false`, which is already part of TestEmptyConstructor.

If you want to check that the file is really closed, you'd have to get the number of open file descriptors from the system before and after the assignment, for example. It doesn't seem trivial, though. Maybe you can try if `open(filename, ...)` with the proper options succeeds instead.

tables/AlternateMans/test/tColumnarFile.cc
```diff
@@ -0,0 +1,164 @@
...
+  const std::array<unsigned char, 4> header = {1, 9, 8, 2};
+  ColumnarFile file = ColumnarFile::CreateNew(filename, header.size(), kStride);
+  file.WriteHeader(header.data());
+  BOOST_CHECK_EQUAL(file.Stride(), 17*8);
+
+  // Generate random data
+  for(size_t i=0; i!=std::size(data_a); ++i)
+      data_a[i] = std::complex<float>(i+3, i*2);
+  for(size_t i=0; i!=std::size(data_b); ++i)
+      data_b[i] = std::complex<float>(i*3+41, i*-2.0);
```
mnijhuis-tos, line 85: I'm surprised casting a `std::array` to `std::size` just works. Still, using the array's `.size()` member is more straightforward:
```suggestion
  for(size_t i=0; i!=data_a.size(); ++i)
      data_a[i] = std::complex<float>(i+3, i*2);
  for(size_t i=0; i!=data_b.size(); ++i)
      data_b[i] = std::complex<float>(i*3+41, i*-2.0);
```

mnijhuis-tos (changes requested):

tables/AlternateMans/test/tColumnarFile.cc
aroffringa: size is not a type, it's a function (so it's not a cast). But I agree that .size() is nicer, I think I was using a regular c-array before, hence the std::size.

tables/AlternateMans/test/tColumnarFile.cc
aroffringa: Fair, done.

tables/AlternateMans/test/tColumnarFile.cc
aroffringa: Hmm really checking the nr of open files sounds a bit fragile. I think if the file is not closed properly, some of the other tests will fail, so I think that should be enough as a test.

mnijhuis-tos (approved):

PR #1382, opened 2025-01-12
Add the Stokes I storage manager classes

aroffringa:
Part of the Stokes I stman changes.

### DIFF
--- a/tables/AlternateMans/StokesIStMan.cc
+++ b/tables/AlternateMans/StokesIStMan.cc
@@ -1,0 +1,129 @@
+#include "StokesIStMan.h"
+
+#include "StokesIStManColumn.h"
+
+void register_stokesistman() { casacore::StokesIStMan::registerClass(); }
+
+namespace casacore {
+namespace {
+  /**
+   * Create an object with given name and spec.
+   * This methods gets registered in the StokesIStMan "constructor" map.
+   * The caller has to delete the object.
+   */
+  static casacore::DataManager *Make(const casacore::String &name,
+                                           const casacore::Record &spec) {
+    return new StokesIStMan(name, spec);
+  }
+}
+
+StokesIStMan::StokesIStMan(const casacore::String &/*name*/,
+                           const casacore::Record &/*spec*/)
+    : DataManager() {
+}
+
+StokesIStMan::StokesIStMan(const StokesIStMan &source)
+    : DataManager(),
+      name_(source.name_) {}
+
+StokesIStMan::~StokesIStMan() noexcept = default;
+
+casacore::Record StokesIStMan::dataManagerSpec() const {
+  return casacore::Record();
+}
+
+void StokesIStMan::registerClass() {
+  DataManager::registerCtor("StokesIStMan", Make);
+}
+
+void StokesIStMan::create64(casacore::rownr_t nRow) {
+  file_ = BufferedColumnarFile::CreateNew(fileName(), {}, CalculateAndUpdateStride());
+  file_.AddRows(nRow);
+}
+
+casacore::rownr_t StokesIStMan::open64(casacore::rownr_t /*n_row*/, casacore::AipsIO &) {
+  file_ = BufferedColumnarFile::OpenExisting(fileName(), 0);
+  return file_.NRows();
+}
+
+casacore::DataManagerColumn *StokesIStMan::makeScalarColumn(
+    const casacore::String & /*name*/, int dataType,
+    const casacore::String &dataTypeID) {
+  std::ostringstream s;
+  s << "Can not create scalar columns with StokesIStMan! (requested datatype: '"
+    << dataTypeID << "' (" << dataType << ")";
+  throw std::runtime_error(s.str());
+}
+
+casacore::DataManagerColumn *StokesIStMan::makeDirArrColumn(
+    const casacore::String & /*name*/, int dataType,
+    const casacore::String & /*dataTypeID*/) {
+
+  if (dataType == casacore::TpFloat || dataType == casacore::TpComplex)
+    return columns_.emplace_back(std::make_unique<StokesIStManColumn>(*this, file_, static_cast<DataType>(dataType))).get();
+  else
+    throw std::runtime_error(
+        "Trying to create a Stokes I data column with wrong type");
+}
+
+casacore::DataManagerColumn *StokesIStMan::makeIndArrColumn(
+    const casacore::String & /*name*/, int /*dataType*/,
+    const casacore::String & /*dataTypeID*/) {
+  throw std::runtime_error(
+      "makeIndArrColumn() called on StokesIStMan. StokesIStMan can only created "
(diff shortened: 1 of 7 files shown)


tables/AlternateMans/StokesIStMan.cc
```diff
@@ -0,0 +1,129 @@
...
+        "Trying to create a Stokes I data column with wrong type");
+}
+
+casacore::DataManagerColumn *StokesIStMan::makeIndArrColumn(
+    const casacore::String & /*name*/, int /*dataType*/,
+    const casacore::String & /*dataTypeID*/) {
+  throw std::runtime_error(
+      "makeIndArrColumn() called on StokesIStMan. StokesIStMan can only created "
+      "direct columns!\nUse casacore::ColumnDesc::Direct as option in your "
+      "column desc constructor");
```
tammojan, line 75:
```suggestion
      "makeIndArrColumn() called on StokesIStMan. StokesIStMan can only create "
      "direct columns!\nUse casacore::ColumnDesc::Direct as option in the "
      "column desc constructor");
```

tables/AlternateMans/StokesIStMan.cc
```diff
@@ -0,0 +1,129 @@
...
+
+casacore::rownr_t StokesIStMan::resync64(casacore::rownr_t nRow)
+{
+  return nRow;
+}
+
+void StokesIStMan::deleteManager() { unlink(fileName().c_str()); }
+
+void StokesIStMan::prepare() {
+}
```
tammojan, line 86:
```suggestion
void StokesIStMan::prepare() {}
```

tables/AlternateMans/StokesIStMan.h
```diff
@@ -0,0 +1,154 @@
...
+   * columns and/or data).
+   */
+  StokesIStMan &operator=(const StokesIStMan &source) = delete;
+
+  casacore::DataManager *clone() const final { return new StokesIStMan(*this); }
+  /**
+   * Create an object with given name and spec.
+   * This methods gets registered in the DataManager "constructor" map.
+   * The caller has to delete the object. New class will be
+   * initialized via @ref DyscoStMan(const casacore::String& name, const
```
tammojan, line 58:
```suggestion
   * initialized via @ref StokesIStMan(const casacore::String& name, const
```

tables/AlternateMans/StokesIStMan.h
```diff
@@ -0,0 +1,154 @@
...
+   * casacore::Record& spec).
+   * @returns A DyscoStMan with given specs.
```
tammojan, line 60:
```suggestion
   * @returns A StokesIStMan with given specs.
```

tables/AlternateMans/StokesIStMan.h
```diff
@@ -0,0 +1,154 @@
...
+    return new StokesIStMan(name, spec);
+  }
+
+  casacore::String dataManagerType() const final { return "StokesIStMan"; }
+
+  casacore::Record dataManagerSpec() const final;
+
+  casacore::Bool canAddRow() const final { return true; }
+
+  casacore::Bool canRemoveRow() const final { return true; }
```
tammojan, line 73: Would it make sense to put this to `False`? If I understand correctly, it can only remove rows at the end, not sure how big the use case for that is.

tables/AlternateMans/StokesIStManColumn.h
```diff
@@ -0,0 +1,240 @@
...
+
+namespace casacore {
+
+class StokesIStMan;
+
+/**
+ * Expands @p n values from single Stokes I values
+ * to have 4 values, in place. This implies the array
+ * should have place to store n*4 values.
+ */
```
tammojan, line 21: Another use case for this storage manager would be to store `WEIGHT_SPECTRUM` (which has four polarizations by default) as one polarization. In that case, the expansion would be to `[value, value, value, value]`. Could you consider making the code generic enough to also handle this?

tables/AlternateMans/StokesIStManColumn.h
```diff
@@ -0,0 +1,240 @@
...
+    const size_t index = n - i - 1;
+    const T value = data[index];
+    data[index * 4] = value;
+    data[index * 4 + 1] = T();
+    data[index * 4 + 2] = T();
+    data[index * 4 + 3] = value;
+  }
+}
+
+template <typename T>
```
tammojan, line 34: Could you insert some documentation here that mentions the factor 0.5?

tables/AlternateMans/StokesIStManColumn.h
```diff
@@ -0,0 +1,240 @@
...
+
+template <>
+inline bool *TransformToStokesI(const bool *input, char *buffer, size_t n) {
+  for (size_t i = 0; i != n; ++i) {
+    new (&buffer[i]) bool(input[i * 4] || input[i * 4 + 3]);
+  }
+  return reinterpret_cast<bool *>(buffer);
+}
+
+constexpr size_t SizeOfType(DataType dtype) {
```
tammojan, line 52: This should exist somewhere else, I hope?

tables/AlternateMans/StokesIStManColumn.h
aroffringa: Yes, that's a good point. To make this stman handle both cases, it would have to know if it is a weight col or a data col. This could be determined by the type (double vs complex), by name (`WEIGHT_SPECTRUM` vs anything else) or through the "specification record", but none of these is very elegant. Dysco uses name and type I think, but it has limitations and I don't really like that approach, as it hard-codes some behaviour. 

Given that we also don't need the weight compression at this point, we could ignore the weight scenario for now. After some thinking, I think the most elegant solution is to make a ScalarWeightStMan, and move most of the StokesIStMan code to a base class that these stmans share.

tables/AlternateMans/StokesIStManColumn.h
aroffringa: It doesn't look like it, I couldn't find it. The way the read/write functions in the other cols seem to work is through overloading, e.g.:
```
    case TpUChar:
        uChar valuc;
        get (rownr, &valuc);
        value = DComplex ((double) valuc);
        return;
    case TpShort:
        Short vals;
        get (rownr, &vals);
        value = DComplex ((double) vals);
        return;
[..]
```
This causes a large number of long `case` statements so I'd say this method is more elegant :). However, this function is more appropriately placed in the `casa/Utilities/DataType.h` file, that way it can be found easier next time it is needed.

tables/AlternateMans/StokesIStManColumn.h
aroffringa: (moved)

tables/AlternateMans/StokesIStManColumn.h
aroffringa: Good point, added. I've now also added a check for non-zero xy/yx values, which should make the dataman a bit safer to use.

tammojan (approved):

PR #1383, opened 2025-01-12
Stokes I st manager: Buffered columnar file

aroffringa:
Part of the Stokes I stman changes.


tables/AlternateMans/BufferedColumnarFile.h
```diff
@@ -0,0 +1,307 @@
...
+  void Close() {
+    if (IsOpen()) {
+      if (block_changed_) {
+        const uint64_t start_row = active_block_ * rows_per_block_;
+        const size_t n_rows_to_write =
+            std::min(rows_per_block_, std::max(NRows(), start_row) - start_row);
+        Seek(start_row * Stride() + DataLocation(), SEEK_SET);
+        WriteData(block_buffer_.data(), n_rows_to_write * Stride());
+        block_changed_ = false;
+      }
```
mnijhuis-tos, line 95: Can you create a (private) function for this code, since it's now duplicated in the destructor, here and in `ActivateBlock`.

tables/AlternateMans/BufferedColumnarFile.h
```diff
@@ -0,0 +1,307 @@
...
+    }
+  }
+
+  VarBufferedColumnarFile& operator=(VarBufferedColumnarFile&& rhs) {
+    RowBasedFile::operator=(std::move(rhs));
```
mnijhuis-tos, line 72: `Close()` is not virtual, so the `Close()` in `RowBasedFile::operator=` doesn't call `VarBufferedColumnarFile::Close()`.
```suggestion
    Close();
    RowBasedFile::operator=(std::move(rhs));
```

tables/AlternateMans/BufferedColumnarFile.h
```diff
@@ -0,0 +1,307 @@
...
+        block_changed_ = false;
+      }
+
+      const uint64_t start_row = block * rows_per_block_;
+      const size_t n_rows_to_read =
+          std::min(rows_per_block_, std::max(NRows(), start_row) - start_row);
+      if (n_rows_to_read > 0) {
+        Seek(start_row * Stride() + DataLocation(), SEEK_SET);
+        ReadData(block_buffer_.data(), n_rows_to_read * Stride());
+      }
```
mnijhuis-tos, line 257: -  I don't see when `n_rows_to_read` can be zero here.
-  Even if it can become zero, the `if` statement is not needed:`ReadData` will just read nothing in that corner case.
```suggestion
      Seek(start_row * Stride() + DataLocation(), SEEK_SET);
      ReadData(block_buffer_.data(), n_rows_to_read * Stride());
```

tables/AlternateMans/BufferedColumnarFile.h
```diff
@@ -0,0 +1,307 @@
...
+      std::fill(block_buffer_.begin() + n_rows_to_read * Stride(),
+                block_buffer_.end(), 0);
```
mnijhuis-tos, line 259: Filling the remaining rows in `block_buffer_` with zeroes does not seem necessary: When writing the last block, `n_rows_to_write` already ensures you're not writing that part of `block_buffer_`.

tables/AlternateMans/BufferedColumnarFile.h
```diff
@@ -0,0 +1,307 @@
...
+    const uint64_t block_row = active_block_ * rows_per_block_;
+    unsigned char* position =
+        block_buffer_.data() + (row - block_row) * Stride() + column_offset;
+    std::copy_n(reinterpret_cast<const unsigned char*>(data),
+                n * sizeof(ValueType), position);
+    SetNRows(std::max(row + 1, NRows()));
+    block_changed_ = true;
+  }
+
+  std::vector<unsigned char> packed_buffer_;
```
mnijhuis-tos, line 295: Maybe you could document that this buffer is only used temporarily for (un)packing booleans, and that defining it as a member avoids memory allocations.

tables/AlternateMans/BufferedColumnarFile.h
```diff
@@ -0,0 +1,307 @@
...
+ * alias BufferedColumnarFile, which uses a reasonable buffer size.
+ *
+ * This class aims to be as fast as possible for large data files that are read
+ * from or written to consecutively. It does not try to optimize random access,
+ * partial data access and access is not transactional.
+ *
+ * The class uses exceptions to handle any I/O errors.
+ */
+template <uint64_t BufferSize = 100 * 1024>
+class VarBufferedColumnarFile : private RowBasedFile {
```
mnijhuis-tos, line 33: Please add tests for this new class.

mnijhuis-tos (changes requested):

tables/AlternateMans/BufferedColumnarFile.h
aroffringa: Wouldn't this make it necessary?
- Active a new block which is not (entirely) on disk, e.g. it has only 1 row. Then row 1 is stored in the buffer, the rest of the buffer has the old active block.
- Inside that block, write row 5. This only updates row 5 in the buffer and extends the nrows, but won't update rows 2-4, which will have data from the previous active block in that case.
- Then read row 3. I think this will return results from the previous active block instead of zeros. 

This is an edge case that is dubious to start with, but I think reading zeros is best in this case.

mnijhuis-tos: Good edge case. It makes me think if the following case works:
- Create a file with 3 rows. (Or a number of rows that results in `block_buffer_` being fully used.)
- Write a single value to second row and first column of the file.

aroffringa: I don't understand the possible issue with your scenario, can you explain what might go wrong?

As far as I can tell it should work: when writing the second row of the file, that block is activated. No rows exist yet in the file, so the buffer is set entirely to zero. Then, writing the value will update the buffer as expected, and the buffer is consistent: reading any other value returns zero, reading back the written value returns the written value. 

I tried to add some test variations like this one, and it succeeds:
```
BOOST_AUTO_TEST_CASE(buffered_file_edge_case) {
  constexpr size_t kColumnSize = sizeof(float) * 2;
  constexpr size_t kStride = kColumnSize * 2;
  const std::string filename = "columnar_file_test.tmp";
  casacore::VarBufferedColumnarFile file = casacore::VarBufferedColumnarFile<kStride*2>::CreateNew(filename, 0, kStride);
  const float values[2] = { 3, 4 };
  // Polute the buffer with some values
  file.Write(2, 0, values, 2);
  file.Write(3, 0, values, 2);
  file.Write(2, kColumnSize, values, 2);
  file.Write(3, kColumnSize, values, 2);
  // At this point, block 1 (rows 2-3) is activated.
  file.Write(5, 0, values, 2);
  // Now, block 2 (rows 4-5) is activated
  float result[2];
  file.Read(5, kColumnSize, result, 2);
  BOOST_CHECK_EQUAL(result[0], 0);
  BOOST_CHECK_EQUAL(result[1], 0);
  file.Read(5, 0, result, 2);
  BOOST_CHECK_EQUAL(result[0], 3);
  BOOST_CHECK_EQUAL(result[1], 4);
  file.Close();
  unlink(filename.c_str());
}
```

mnijhuis-tos: Where does it set the buffer to zero? I searched for `fill` in this PR, but could only find this fill and some fills that occur when the row number is out of range.

tables/AlternateMans/BufferedColumnarFile.h
aroffringa: It will be zero when `NRows() <= start_row`, which happens when creating a file and e.g. activating the first block (NRows() == 0, start_row = 0), or a block that is not existing yet.

You're right the if is not strictly necessary, which I hadn't realized. It may be an optimization though; it prevents calling Seek() everytime a new block is started (so when writing a file from start to end, for all blocks). Even though a seek is a relatively expensive operation (it's a kernel call, but might even perform I/O), with a large buffer size, that overhead is relatively small though, but in the write case it is pure overhead: the `if` statement never evaluates to false. However, when reading an existing file, it returns false every time, so there the if is always overhead. I think I would indeed then just leave the if out as you suggest, as on average it won't help much, if any. So, I'll remove it.

tables/AlternateMans/BufferedColumnarFile.h
```diff
@@ -0,0 +1,307 @@
...
+namespace casacore {
+
+/**
+ * Class that provides binary table I/O. It is rather low-level, and requires
+ * the user to keep track of the columns and their datatype. It uses a simple
+ * binary format where the columns are interleaved into the output file.
+ * Booleans are written with bit-packing.
+ *
+ * This class limits calls to the read/write functions by caching a block of
+ * data around the last accessed row. The size of the buffer is templated to
```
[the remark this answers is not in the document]
aroffringa: The OS cache is certainly important. But accessing the OS cache still requires a kernel call (may go out of user mode?) and synchronization on the cache, which all together is relatively expensive. I've timed this class against the SimpleColumnarFile, when using the new Dp3 branch to write StokesIStMans, and it was on the order of 20% faster on my desktop. It can be expected that reads, which are generally faster (making the kernel call a larger chunk of the time), it will be even more important, but I haven't tested that. No doubt it highly depends on disk speed, access pattern, etc., but my experience from the wsclean read things is also that calling small read functions many times causes high CPU usage and (at least sometimes) slower processing.

I've doubted too whether it was necessary or desirable, but given my first timings I tend to say it is, and at the very least it doesn't hurt, apart from having more and more complex code.

tables/AlternateMans/BufferedColumnarFile.h
aroffringa: You've already reviewed those tests ;). In the unit tests you reviewed, the tests for SimpleColumnarFile are also all executed on the BufferedColumnarFile.

mnijhuis-tos: I didn't see this PR yet when reviewing the tests... There's no specific test for the BufferedColumnarFile, for example a test that explicitly fills the internal buffer only partially, but the generic tests should already provide decent coverage.

tables/AlternateMans/BufferedColumnarFile.h
aroffringa: This is the fill that does that. First write will call ActivateBlock(0), (because the active_block is set to max uint64_t), there are no rows to read so the entire buffer is filled.

tables/AlternateMans/BufferedColumnarFile.h
aroffringa: Good idea, done

tables/AlternateMans/BufferedColumnarFile.h
aroffringa: I've added this test in a separate PR: https://github.com/casacore/casacore/pull/1387

mnijhuis-tos: It's nice that you now handle many cases using this single fill, but, as you noticed, that's not immediately obvious. Therefore please add a comment, like "Fill the remainder of block_buffer_ with zeroes. Doing it here makes the code robust and avoids the need for inserting zeroes when adding rows out-of-order, e.g., when adding row 5 while NRows() is 2."

aroffringa: Good idea, added!

mnijhuis-tos (approved):

PR #1384, opened 2025-01-13
Verify files are proper FITS-IDI before conversion

kettenis:
The astropy.io.fits module inadvertedly converts FITS-IDI files into random group structure FITS.  Since casacore interprets FITS keywords differently for such files this results in misinterpretation of the origin of the FITS-IDI and therefore the digital corrections aren't applied correctly.  Prevent this by generating an error if the user tries to convert random groups FITS using the FITS-IDI conversion code.

### DIFF
--- a/msfits/MSFits/MSFitsIDI.cc
+++ b/msfits/MSFits/MSFitsIDI.cc
@@ -276,6 +276,12 @@ void MSFitsIDI::readFITSFile(Bool& atEnd)
     os << LogIO::SEVERE << "Error reading FITS input" << LogIO::EXCEPTION;
   }
 
+  // Make sure this is a FITS-IDI file; the Astropy FITS code may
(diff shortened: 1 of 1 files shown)


tammojan (approved):
Assuming that `LogIO::EXCEPTION` throws an exception (`std::runtime_error` (or a casacore variant). That looks good to me.

kettenis:
When called from CASA's importfitsidi task this does indeed generate an exception:
`RuntimeError: 2025-01-13 10:40:37	SEVERE	MSFitsIDI::readFITSFile() (file /home/kettenis/src/casa6/casatools/casacore/msfits/MSFits/MSFitsIDI.cc, line 270)	Not a FITS-IDI file`

dpetry:
Yes, looks good!

PR #1385, opened 2025-01-15
Stokes I storage manager

aroffringa:
This is the main branch for a Stokes I stman. PRs #1380 , #1381 , #1382 and #1383 merge towards this branch. Once all those PRs are merged and this branch is fully functional, it will be merged to master.

### DIFF
--- a/casa/Utilities/DataType.h
+++ b/casa/Utilities/DataType.h
@@ -214,6 +214,56 @@ DataType asScalar(DataType type);
 DataType asArray(DataType type);
 // </group>
 
+/**
+ * Returns the number of bytes that this type takes when serialized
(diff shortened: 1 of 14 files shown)


tammojan (approved):
This looks good to me. The implementation is a bit different from the other casacore storage managers, which I think can be justified given recent opportunities in C++. Good to put it in a separate directory.
Making this a storage manager in casacore (rather than a shared library in its own repository) will make it easier for users to install. For casacore users who do not use this storage manager, the extra compile time and library file size is negligible.
There is a bit of interplay between MS and storage manager here. In principle, storage managers are generic, and agnostic of visibilities and MS, but I think Stokes I is the exception to this rule (like dysco).
All in all, very nice addition, I look forward to experimenting with this.

PR #1386, opened 2025-01-15
added support for beam-waveguide mount

flep198:
Added support for Beam-Waveguide mounts as discussed in https://arxiv.org/abs/2210.13381

### DIFF
--- a/msfits/MSFits/MSFitsOutput.cc
+++ b/msfits/MSFits/MSFitsOutput.cc
@@ -1840,7 +1840,7 @@ Bool MSFitsOutput::_writeAN(std::shared_ptr<FitsOutput> output, const Measuremen
                 *mntsta = 8;
             }
             else {
-                *mntsta = 9; // fits does not use anyway, put it 7
+                *mntsta = 9; // fits does not use anyway, put it 9
             }
             *staxof = inantoffset(antnum)(IPosition(1, 0));
             // OK, try to find if we're L/R or X/Y
--- a/msfits/MSFits/MSFitsOutput.cc
+++ b/msfits/MSFits/MSFitsOutput.cc
@@ -1815,6 +1815,12 @@ Bool MSFitsOutput::_writeAN(std::shared_ptr<FitsOutput> output, const Measuremen
             else if (mount.contains("ALT-AZ+NASMYTH-L")) {
                 *mntsta = 5;
             }
+	    else if (mount.contains("ALT-AZ+BWG-R")) {
+		*mntsta = 6;
+	    }
+	    else if (mount.contains("ALT-AZ+BWG-L")) {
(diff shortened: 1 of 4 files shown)


msfits/MSFits/MSFitsOutput.cc
```diff
@@ -1828,13 +1828,19 @@ Bool MSFitsOutput::_writeAN(std::shared_ptr<FitsOutput> output, const Measuremen
...
+	    else if (mount.contains("ALT-AZ+BWG-L")) {
+		*mntsta = 7;
+	    }
             else if (mount.contains("BIZARRE")) {
-                *mntsta = 6;
+                *mntsta = 8;
             }
             else {
-                *mntsta = 7; // fits does not use anyway, put it 7
+                *mntsta = 9; // fits does not use anyway, put it 7
```
kettenis, line 1843: Comment still mentions the value 7.

flep198: fixed through latest commit

kettenis:
Thanks @flep198, the code looks good to me. I'm currently testing a CASA build that integrates these changes (found some public LBA data that includes the Warkworth 30m antenna) and will come back to this pull request when that is done.

msfits/MSFits/MSFitsOutput.cc
```diff
@@ -1828,13 +1828,19 @@ Bool MSFitsOutput::_writeAN(std::shared_ptr<FitsOutput> output, const Measuremen
...
             else if (mount.contains("SPACE-HALCA")) {
-                *mntsta = 7;
+                *mntsta = 9;
             }
+	    else if (mount.contains("ALT-AZ+BWG-R")) {
+		*mntsta = 6;
+	    }
+	    else if (mount.contains("ALT-AZ+BWG-L")) {
+		*mntsta = 7;
+	    }
```
kettenis, line 1838: While testing `exportuvfits` in my CASA build, I found out this doesn't actually work.  Since the code does `match.contains` the checks for the extended "ALT-AZ+" types needs to come before the "ALT-AZ" one. Can you move these new "BWG" ones up such that they come straight after the "NASMYTH" ones?

kettenis:
Here are some plots of the RR phase minus the LL phase after running `applycal` with `parang=True` in a CASA build with these changes:
bwg
Some of the data is a bit noisy (I didn't make any effort to flag this data set) but the plot clearly shows that phase difference is constant over time which shows the the correct parallectic angle correction is applied.

msfits/MSFits/MSFitsOutput.cc
flep198: that makes a lot of sense, I just fixed this through the latest commit.

PR #1388, opened 2025-01-16
Update upstream URLs for IERS, in effect since 2023-03-06

tammojan:


dpetry (approved):
Thanks for updating. I understand this has been in production for some time and is therefore already well tested.

PR #1391, opened 2025-01-19
Fix the IPosition std::vector<ll> constructor

### LINKED ISSUE #1389: IPosition error with debug build
When `casacore` is compiled with debug **on**, the following example exits with an assertion error:
```
#include <vector>
#include <casacore/casa/Arrays/IPosition.h>

int main( ) {
    std::vector<long long> vec = {3,4};
    casacore::IPosition shape(vec);
}
```

aroffringa:
Fixes #1389.


schiebel:
Thanks for the fix @aroffringa !

tammojan (approved):

PR #1396, opened 2025-01-24
Add UvwFile writer for Uvw column compression

aroffringa:


tables/AlternateMans/test/tUvwFile.cc
```diff
@@ -0,0 +1,187 @@
...
+      ++row;
+    }
+  }
+}
+
+BOOST_AUTO_TEST_CASE(write_and_direct_read) {
+  UvwFile file = UvwFile::CreateNew(kFilename);
+  WriteTwoTimesteps(file, kBaselines, true);
+  ReadTwoTimesteps(file, kBaselines, true);
+  unlink(kFilename.c_str());
```
mnijhuis-tos, line 133: Shouldn't you `file.Close();` before unlinking it? Also in the other tests?

mnijhuis-tos (changes requested):

tables/AlternateMans/UvwFile.h
```diff
@@ -0,0 +1,351 @@
...
+    source.reference_antenna_ = 0;
+    source.start_antenna_2_ = 0;
+    source.n_antenna_ = 0;
+    source.block_uvws_.clear();
+    source.block_is_changed_ = false;
+  }
+
+  ~UvwFile() noexcept { Close(); }
+
+  UvwFile& operator=(UvwFile&& rhs) {
```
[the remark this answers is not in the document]
aroffringa: Good catch, done

tables/AlternateMans/UvwFile.h
```diff
@@ -0,0 +1,351 @@
...
+    n_antenna_ = reinterpret_cast<uint64_t&>(data[24]);
+  }
+
+  void WriteHeader() {
+    unsigned char data[kHeaderSize] = "Uvw-col";
+    reinterpret_cast<uint64_t&>(data[8]) = rows_per_block_;
+    reinterpret_cast<uint64_t&>(data[16]) = reference_antenna_;
+    reinterpret_cast<uint64_t&>(data[24]) = n_antenna_;
+    file_.WriteHeader(data);
+  }
```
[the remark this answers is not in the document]
aroffringa: The issue with structs are that they may have non-portable and inconsistent packing/padding bytes to align types. While in this case this appears unlikely, even here the compiler could decide to e.g. align the first uint64_t on a 16 byte start (somewhat unlikely though). There are compiler-specific ways to enforce consistent packing, but I prefer this approach over compiler specific things.

tables/AlternateMans/UvwFile.h
```diff
@@ -0,0 +1,351 @@
...
+
+  bool IsSet(size_t antenna) const {
+    return block_uvws_.size() > antenna &&
```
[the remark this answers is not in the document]
aroffringa: These assumptions aren't entirely accurate, but it's good that you mention this. The second point, that the reference antenna is not 0, wasn't correct: the absolute position never matters, but the relative position is always 0.

You do correctly point out that in the way it works, there couldn't be two antennas with the same position (if one of the antennas would by accident also be the reference antenna). This may however occur in theory; it's a border case but it wasn't so hard to support. I therefore now use a different value (kUnsetPosition) to denote an unset position. So, all in all, good catch.

tables/AlternateMans/test/tUvwFile.cc
```diff
@@ -0,0 +1,187 @@
...
+      std::array<double, 3> uvw;
+      file.ReadUvw(row, baselines[i][0], baselines[i][1], uvw.data());
+      for(size_t z=0; z!=3; ++z)
+        BOOST_CHECK_CLOSE_FRACTION(kUvwPerBaseline[i][z], uvw[z], 1e-6);
+      ++row;
+    }
+  }
+}
+
+BOOST_AUTO_TEST_CASE(write_and_direct_read) {
```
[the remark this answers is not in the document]
aroffringa: I think that's a bit of a too "white-box" / specific (non-behavioural) way of testing, as the exact size depends on the internals of the columnar file and the internals of UvwFile. Even a more generic "approximately less than" test would require more data (with the test sizes, the file might not be smaller than the uncompressed one). So I think I prefer not to test this in CI ; I've done a manual check so at present it works, and it's at least hard to break the compression while passing all the other checks.

tables/AlternateMans/test/tUvwFile.cc
aroffringa: Done

tables/AlternateMans/UvwFile.h
```diff
@@ -0,0 +1,403 @@
...
+   * rows in one block, i.e. the number of set elements in the correlation
+   * matrix.
+   */
+  uint64_t rows_per_block_ = 0;
+  uint64_t active_block_ = 0;
+  size_t reference_antenna_ = 0;
+  // This value is used to determine the first baseline in the data, which is
+  // the baseline (reference_antenna_, start_antenna_2_).
+  size_t start_antenna_2_ = 0;
+  size_t n_antenna_ = 0;
```
mnijhuis-tos, line 395: Although `n_antenna_` is a clear name, `n_antennas_` would be more consistent with `n_rows_`.
```suggestion
  size_t n_antennas_ = 0;
```

mnijhuis-tos (changes requested):

tables/AlternateMans/UvwFile.h
```diff
@@ -0,0 +1,403 @@
...
+          } else {
+            block_uvws_.emplace_back(std::array<double, 3>{0.0, 0.0, 0.0});
+          }
+        }
+      } else {
+        std::fill(block_uvws_.begin(), block_uvws_.end(), kUnsetPosition);
+        block_uvws_[reference_antenna_] = {0.0, 0.0, 0.0};
+      }
+      active_block_ = block;
+    }
```
mnijhuis-tos, line 301: Optional: Since there's also a `WriteActiveBlock()` function, you could move most this code to a `ReadActiveBlock()` function, so it will look like:
```
     active_block_ = block;
     ReadActiveBlock();
}

void ReadActiveBlock() { 
     const uint64_t block_start_row = (n_antenna_ - 1) * active_block_;
      if (block_start_row < file_.NRows()) {
        block_uvws_.clear();
        for (size_t antenna = 0; antenna != n_antenna_; ++antenna) {
          if (antenna != reference_antenna_) {
            const uint64_t row = antenna < reference_antenna_
                                     ? block_start_row + antenna
                                     : block_start_row + antenna - 1;
            std::array<double, 3>& uvw = block_uvws_.emplace_back();
            file_.Read(row, 0, uvw.data(), 3);
          } else {
            block_uvws_.emplace_back(std::array<double, 3>{0.0, 0.0, 0.0});
          }
        }
      } else {
        std::fill(block_uvws_.begin(), block_uvws_.end(), kUnsetPosition);
        block_uvws_[reference_antenna_] = {0.0, 0.0, 0.0};
      }
    }
}
```

tables/AlternateMans/UvwFile.h
```diff
@@ -0,0 +1,403 @@
...
+    if (rows_per_block_ != 0) {
+      const uint64_t block = row / rows_per_block_;
+      ActivateBlock(block);
+    }
+    uvw[0] = block_uvws_[antenna2][0] - block_uvws_[antenna1][0];
+    uvw[1] = block_uvws_[antenna2][1] - block_uvws_[antenna1][1];
+    uvw[2] = block_uvws_[antenna2][2] - block_uvws_[antenna1][2];
+  }
+
+  void Close() {
```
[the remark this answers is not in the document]
aroffringa: Adding asserts to a class that's only used in a specific way in it's stman feels a bit like gold plating ;), but I've added asserts.

tables/AlternateMans/UvwFile.h
aroffringa: Sure, that indeed improves the structure.

tables/AlternateMans/UvwFile.h
```diff
@@ -373,7 +380,7 @@ class UvwFile {
    * Uvw row using @ref WriteUvw(). This concept of a "row" is different
    * from a row in the BufferedColumnarFile (i.e., @c file_ ), as for each
    * block (see below), only the uvw per antenna are stored, and these form
-   * the rows in that file.
+   * the rowsn_antenna_ in that file.
```
mnijhuis-tos, line 383:
```suggestion
   * the rows in that file.
```

mnijhuis-tos (approved):

PR #1400, opened 2025-02-01
Various finishing touches for the compressing storage managers

aroffringa:
These changes add support for opening a read-only file, and make sure that no writing or truncation is performed when no changes are made. This is e.g. important for supporting MS-es as input in e.g. CWL tasks.

It also adds a magic header mark to the Stokes I data column, such that Stokes I, UVW and Antenna cols can be distinguished from each other already from the header. This changes the format of the Stokes I columns, but since it is not used yet, now is the time (and it should be the last change). This makes the Stokes I stman in line with the other compressing storage managers.

A test set is added for the operations done in the Stokes I stman.

Finally there are some fixes to make sure everything remains robust when failures occur. The RO changes showed that e.g. if a close fails but the state of the file is not updated accurate, the destructor might then also throw, thereby causing a terminate instead of an exception. This should be all fixed now.

### DIFF
--- a/docker/ubuntu2204_clang.docker
+++ b/docker/ubuntu2204_clang.docker
@@ -12,6 +12,8 @@ RUN export DEBIAN_FRONTEND=noninteractive \
     wcslib-dev \
     libfftw3-dev \
     gfortran \
+    libc++-dev \
+    libc++abi-dev \
     libncurses5-dev \
     libreadline6-dev \
     libhdf5-serial-dev \
@@ -37,6 +39,7 @@ RUN useradd -ms /bin/bash casacore \
 USER casacore
 RUN mkdir /code/build
 WORKDIR /code/build
+ENV CXXFLAGS="${CXXFLAGS} -stdlib=libc++"
 RUN cmake .. \
     -DBUILD_TESTING=ON \
     -DUSE_OPENMP=OFF \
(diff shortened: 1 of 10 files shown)


docker/ubuntu2204_clang.docker
```diff
@@ -37,6 +39,7 @@ RUN useradd -ms /bin/bash casacore \
 USER casacore
 RUN mkdir /code/build
 WORKDIR /code/build
+ENV CXXFLAGS="${CXXFLAGS} -stdlib=libc++"
```
tammojan, line 42: Why use libc++ instead of libstdc++? I think it would be good if also the tests work with libstdc++. If a part of it does not, we could work around that by ifdef'ing that part.

aroffringa: This is not strictly required anymore, so it could be removed again. However, this is because of the extra 'boost_system' requirement, which results in that a part of the tests no longer is compiled. If they are compiled, the version of libstdc++ seemed at least too old to use the filesystem things in the test (I couldn't quite figure out why, but I thought that not running a test on an older platform is fine-ish)._ Also, clang is really made to compile with libc++, plus having on test run with that is a nice extra compatibility check, hence I kept these changes. If you rather not I'm also happy to remove them again.

tammojan (approved):

PR #1407, opened 2025-03-15
Fix an out of bounds issue in Stokes I StMan

aroffringa:
In boolean (bit-packed) columns, the code confused nr of bytes with nr bits.


mnijhuis-tos (approved):

tammojan (approved):
What is the status of files that have been compressed with the StokesIStMan with this bug? Is there a test that should have caught this bug?

aroffringa:
> What is the status of files that have been compressed with the StokesIStMan with this bug? Is there a test that should have caught this bug?

Data that has been compressed is fine. (Also only the experimental branch of Dp3 is using it so far, so it's not in formal production yet.)

The Dp3 tests seem to catch this bug "now", but not before, which I don't understand (there are successful builds of the Dp3 metadata-compression on CI, but it started failing later).

PR #1409, opened 2025-03-27
Support UVW compression on sets with only auto-correlations

aroffringa:
When writing a file with only auto-correlations, enabling UVW compression would cause issues. When writing measurement sets with multiple auto-correlations, the UvwFile would never be changed (as there's nothing to write) and therefore the file would never be written. Technically such a file is not supported by the given requirements for UVW compression, as it says:

> Also, it requires the baselines to be ordered such that a relation to the
> reference antenna can be made.

...and that's not the case when there are more than one antennas. However, it's easy to support files with auto-correlations, so that's added in this MR.

A second issue was that if a file has only one auto-correlation and therefore also only one antenna, the nr of rows in the file was not properly calculated. This is also fixed.

### DIFF
--- a/tables/AlternateMans/UvwFile.h
+++ b/tables/AlternateMans/UvwFile.h
@@ -140,6 +140,7 @@ class UvwFile {
         ActivateBlock(1);
       } else {
         n_antennas_ = std::max({antenna1 + 1, antenna2 + 1, n_antennas_});
+        block_uvws_.resize(n_antennas_, kUnsetPosition);
       }
     } else {
       const uint64_t block = row / rows_per_block_;
@@ -173,6 +174,16 @@ class UvwFile {
             "Baselines are written in a non-ordered way: they need to be "
             "ordered either by antenna 1 or by antenna 2");
       }
+    } else {
+      // In the special case of a file with only auto-correlations, the
+      // positions of the antennas can not be established in this case, it's
+      // also not necessary, because a zero uvw can be returned for
+      // auto-correlations. However, it will cause StoreOrCheck() to be
+      // never called, and therefore the block is never written to the file, and
+      // upon read the nr. of rows can not be determined. So, if this is an
+      // auto-correlation that has not yet been written, mark the block as
+      // changed so it gets written.
+      block_is_changed_ = block_is_changed_ || !IsSet(antenna1);
     }
     n_rows_ = std::max(n_rows_, row + 1);
   }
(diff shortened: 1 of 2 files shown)


tables/AlternateMans/UvwFile.h
```diff
@@ -173,6 +174,16 @@ class UvwFile {
...
+    } else {
+      // In the special case of a file with only auto-correlations, the
+      // positions of the antennas can not be established in this case, it's
+      // also not necessary, because a zero uvw can be returned for
+      // auto-correlations. However, it will cause StoreOrCheck() to be
+      // never called, and therefore the block is never written to the file, and
+      // upon read the nr. of rows can not be determined. So, if this is an
+      // auto-correlation that has not yet been written, mark the block as
+      // changed so it gets written.
+      block_is_changed_ = block_is_changed_ || !IsSet(antenna1);
```
tammojan, line 186: How is `!IsSet(antenna1)` an indicator for being an autocorrelation?

tammojan (approved):
Just curious: will the UVW compression also work if we have, say, 6 antennas, with correlations in two disjoint groups, say 0, 1, 2 all connected and 3, 4, 5 all connected, but no baselines between these groups?

tables/AlternateMans/UvwFile.h
aroffringa: This is in the `else` of `if (antenna1 != antenna2)`, hence this condition only runs on autocorrelations.

In theory we could use `block_is_changed_ = true` here, but then everytime a auto-correlation is set, it will cause rewrite of the whole block, which isn't necessary. So the `!IsSet(antenna1)` is a slight optimization, to only write if the position of the antenna is unknown, which makes sure that on a 'normal' MS, setting the auto-correlations again won't update the file needlessly.

aroffringa:
> Just curious: will the UVW compression also work if we have, say, 6 antennas, with correlations in two disjoint groups, say 0, 1, 2 all connected and 3, 4, 5 all connected, but no baselines between these groups?

No, that won't work with the current implementation. This is caught though, and tested in the `invalid_baselines` test, so a reasonably understandable error is printed in that case.

It would be possible to support that use case with the same file format, basically by allowing multiple antennas to be a reference antenna. Doing so however leads to the complex situation that if you allow any ordering of baselines, it might be necessary to "join" two disjoint groups if they have a shared baseline later on, e.g. if the baselines are:
```
0, 1
0, 2
3, 4
3, 5
0, 3
```
The 3rd line would have to recalculate the positions of antennas 3, 4 and 5. This is not super hard, but does increase the required code quite a bit.

I guess one example where this in theory could occur is if all long baselines are removed. In that case, you might get two clusters of baselines that are disjoint. So it's not unthinkable, but still rare. An of course it's relatively easily solved by either turning metadata compression on, or by flagging the baselines instead of removing them.

PR #1410, opened 2025-04-06
Fix an indexing bug in DataManInfo::removeDminfoColumns()

aroffringa:
There's an indexing bug in `DataManInfo::removeDminfoColumns()`, which triggered an error in the test that was added in Dp3 here: https://git.astron.nl/RD/DP3/-/merge_requests/1431 .

The function used the same index for the input and output info structures. However, if dms are removed somewhere in the middle, the output info has fewer elements, so the wrong index is used. This causes this error to be thrown:

```
std exception detected: RecordInterface::define - new fieldNumber exceeds #fields
```

This test compresses a MS with meta-data compression, and then decompresses it. This causes a call to `DataManInfo::removeDminfoColumns()` in which the dm for the antenna columns is not the first dm in the set, which triggered this bug.

### DIFF
--- a/tables/DataMan/DataManInfo.cc
+++ b/tables/DataMan/DataManInfo.cc
@@ -452,23 +452,24 @@ Vector<String> DataManInfo::removeDminfoColumns (Record& dminfo,
   // Keep track which columns are removed.
   Vector<String> remCols(columns.size());
   uInt ncols = 0;
-  for (uInt j=0; j<dminfo.nfields(); j++) {
-    Record rec = dminfo.subRecord(j);
+  uInt newdm_index = 0;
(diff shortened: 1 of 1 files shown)


tammojan (commented):
Well spotted, good change! Would it be possible to add a short test for this case in tDataManInfo.cc ?

aroffringa:
> Well spotted, good change! Would it be possible to add a short test for this case in tDataManInfo.cc ?

Writing a test that specifically tests the bugfix seems a bit behind the fact ;), and performing a full testing of the `removeDminfoColumns()` function is a bit beyond the scope of this fix (apart from that it is a quite an "ad-hoc" function;  wouldn't be surprised if this function was added specifically for Dp3, because it's not very generic.) I've at least added a test-set to Dp3 that catches the bug, which I think is sufficient.

PR #1411, opened 2025-04-16
Add cronjobs for measures update

tammojan:
These files are running the daily update of "WSRT_Measures.ztar" at ASTRON. Copied into the repository for reference.


dpetry (approved):
I suppose all this was tested well. I'll approve.
I suggest you modify the automatic mailing of reports in a future version
such that the subject line of the mail indicates if there were problems.
In this way, one doesn't have to actively look at log files every day.

PR #1412, opened 2025-04-16
Retire ubuntu 20.04 build on CI

tammojan:
See https://github.com/casacore/casacore/actions/runs/14494768719 : ubuntu 20.04 has been retired on github since a few days.


gmloose (commented):
LGTM

PR #1422, opened 2025-07-20
Fix race condition bugs in initialization of measures

aroffringa:


measures/Measures/Aberration.cc
```diff
@@ -63,12 +56,12 @@ Aberration &Aberration::operator=(const Aberration &other) {
...
 void Aberration::init() {
     method = Aberration::STANDARD;
-    fill();
+    checkEpoch = 1e30;
 }
 
 void Aberration::init(AberrationTypes type) {
     method = type;
-    fill();
+    checkEpoch = 1e30;
```
mnijhuis-tos, line 64: Optional: Calling the overload without arguments removes a little bit of duplicate code.
```
init(Aberration::STANDARD)
```

measures/Measures/Aberration.h
```diff
@@ -161,15 +162,16 @@ class Aberration
 // Last calculation
     MVPosition result[4];
 // Interpolation interval
-    static uInt interval_reg;
+    inline static uInt interval_reg;
 // JPL use
-    static uInt usejpl_reg;
+    inline static uInt usejpl_reg;
+    inline static std::once_flag initialize_once_flag;
```
mnijhuis-tos, line 168: After reading the documentation for inline static member variables, I think using 'inline' could lead to errors here.

Since Aberration.cc initializes these members, only the copies of this variable in Aberration.o will be initialized. If another foo.cc file uses a function from Aberration.h which has the implementation in the header, foo.cc will get another copy of the inline variable which is not initialized.

Since all member function implementations are currently in Aberration.cc, it will work now. Adding a simple `get_interval_reg()` function, which returns `interval_reg`, will break things, though.

Also, since Aberration.cc is  the only user of these members, inline doesn't have much effect; There will still be one copy of the  members after linking all object files together.

Of course, the same reasoning applies to the new `inline static` members in other files.

aroffringa: I'm not sure I follow... static and inline static variables have external linkage so only one copy exists, e.g. from https://en.cppreference.com/w/cpp/language/static.html :

> Static data members
> Static data members are not associated with any object. They exist even if no objects of the class have been defined. There is only one instance of the static data member in the entire program with static [storage duration](https://en.cppreference.com/w/cpp/language/storage_duration.html), unless the keyword [thread_local](https://en.cppreference.com/w/cpp/keyword/thread_local.html) is used, in which case there is one such object per thread with thread storage duration(since C++11).

There's also info there about inline static members functions. The addition of inline stops the requirement of having to define the variable inside a unit file. Other than that, it is exactly the same. Maybe you're confused with a unit-scope static variable? (i.e. not being in a class) They get internal linkage, and those may be redeclared and "see" different values in different unit files.

measures/Measures/Aberration.cc
aroffringa: Good idea. I gave the `init` function with parameter a default so that it covers both: `init(AberrationTypes type = Aberration::STANDARD)`.

mnijhuis-tos (approved):

PR #1426, opened 2025-07-27
Simplify MeasFrame for improving thread-safety and add independentCopy()

aroffringa:
I did get at some point thread sanitizer errors on the 'reference counter' variable that's used. I'm not entirely sure they were correct, but std::shared_ptr uses an atomic counter by default and therefore solves this issue. It also simplifies the class a lot, making it more obvious what happens ; and makes move construct/assign trivial.

### DIFF
--- a/measures/Measures/MeasFrame.h
+++ b/measures/Measures/MeasFrame.h
@@ -175,9 +175,12 @@ class MeasFrame {
 	    const Measure &meas3);
   // </group>
   // Copy constructor (reference semantics)
-  MeasFrame(const MeasFrame &other);
+  MeasFrame(const MeasFrame &other) = default;
+  MeasFrame(MeasFrame &&other) = default;
+  
   // Copy assignment (reference semantics)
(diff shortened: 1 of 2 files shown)


measures/Measures/MeasFrame.h
```diff
@@ -287,13 +291,19 @@ class MeasFrame {
   // Get the comet coordinates
   Bool getComet(MVPosition &tdb) const;
   // </group>
+
+  // Make a value copy of this MeasFrame, such that it
+  // contains no reference to the old MeasFrame. This
+  // can be useful for insuring thread safety in conversions.
```
mnijhuis-tos, line 297:
```suggestion
  // can be useful for ensuring thread safety in conversions.
```

mnijhuis-tos (approved):
Nice improvement. I always enjoy replacing manual locking code by `std::shared_ptr`s. I only found a minor typo. If you fix it, you may immediately merge this PR.

PR #1430, opened 2025-07-30
Use a cyclic ptr in MeasFrame to solve cyclic references

### LINKED ISSUE #1082: MeasFrame accesses deleted object in destructor.
The following very basic test causes an error in the ~MeasFrame destructor:

```
#include <casacore/measures/Measures.h>
int main(int argc, char** argv) {
  casacore::MeasFrame mf;
  return 0;
}
```

I added some print statements to the ~MeasFrame destructor, so it looks like:
```
MeasFrame::~MeasFrame() {
  cout << "Delete rep = " << rep << " cnt = " << rep->cnt << endl;
  if (rep && rep->cnt && --rep->cnt == 0) {
    cout << "Actual delete" << endl;
    delete rep;
  }
}
```

The resulting output is:
```
Delete rep = 0xbdb9f20 cnt = 1
Actual delete
Delete rep = 0xbdb9f20 cnt = 0
```

The destructor for MeasFrame is thus called twice. On the first invocation, `rep->cnt` equals one and the destructor thus deletes rep. On the second invocation, the destructor reads the cnt member from a deleted `rep` object, which is an error.

The reason why cnt equals 1, and not 2, in the first destructor call, is as follows:
- MeasFrame::create executes `lock(locker)`, which stores the current `rep->cnt` value.
- MeasFrame::create executes `rep->mymcf = new MCFrame(*this);`.
- The MCFrame constructor copies the MeasFrame object. The MeasFrame copy constructor increases rep->cnt to 2.
- MeasFrame::create executes `unlock(locker)`, which restores original `rep->cnt` value to 1, even though there are two MeasFrame objects that use the same `rep` pointer.

I haven't dug deeper into the reasons for having the lock and unlock functions in MeasFrame. For properly fixing this issue, I would advice using a `std::shared_ptr` for `rep` instead of using manual reference counting.

aroffringa:
MeasFrame was recently converted (in #1426) to use a shared_ptr for the implementation ("rep") member. While the cycle to MCFrame was previously already removed, it turns out there are several other ownership cycles,e.g. MeasFrame->MCFrame->MeasConvert->MeasRef->MeasFrame. Therefore #1426 has introduced a bug that leaks in certain cases, as was also observed in a previous effort to cleanup MeasFrame (#1111).

The old system of locking functions was undocumented and highly confusing. However, it is evidently not easy to replace it with a shared_ptr. Using weak_ptrs would also be really challenging and may require heavy restructuring, mostly because currently the route through which the cycle can form is not constant and not always present (and as such a weak ptr may cause early destruction). Therefore, I went for a new solution, and implemented a shared_ptr-like class called CyclicPtr, which allows freezing the counter temporarily to let some references not count in ownership. It adds a bit of code and I considered if that's worth it, but I think it is. It separates the concerns of reference counting into a separate class, and no longer mixes that with the task of the MeasFrame class. It makes the reference counting easier to test, better documented, and solves some issues. As listed in the CyclicPtr code:

> Originally, the FrameRep class held a counter and was destructed when the counter reached zero. While this
> worked, it had issues:
>
>  - It was hard to understand the implementation. It looked like a 
>    "standard" manual reference-counting implementation and it therefore
>    seemed like it could be replaced by a shared_ptr. This class helps
>    to annotate the intention.
>  - It had bugs: it would sometimes access the counter after destruction
>    of the MeasFrame (see #1082). By holding the counter inside the pointer
>    class, this is avoided.
>  - The counter was not implemented in a safe manner, hence copying the
>    same frame to different threads would cause a race condition even
>    when the frame was only read from (/copied).

This (finally) fixes #1082 without introducing the bug in #1111. 

BTW I've introduced a Boost test for CyclicPtr. The casa/arrays module already used boost testing so it's not a new dependency. The measures test could be a lot simpler and easier to understand if they would all use boost test, so now there's a first bit that could be interesting for the feature.

### DIFF
--- a/measures/Measures/CyclicPtr.h
+++ b/measures/Measures/CyclicPtr.h
@@ -134,7 +134,7 @@ public:
 
   /**
    * Freeze the reference counter. The object is guaranteed not to
-   * be destructed  until Unfreeze() is called.
+   * be destructed until Unfreeze() is called.
    *
    * The caller is responsible for calling Unfreeze(); letting CyclicState go
    * out of scope without unfreezing will not cause an error, so will lead to
--- a/measures/Measures/MCFrame.cc
+++ b/measures/Measures/MCFrame.cc
@@ -441,6 +441,9 @@ void MCFrame::makeEpoch(MeasFrame& frame) {
   impl_->epTDBp.reset();
   impl_->epUT1p.reset();
   impl_->epTTp.reset();
+  // Initializing the MeasConvert may cause a cycle:
+  // MeasConvert -> MeasRef -> MeasFrame and
+  // MeasConvert -> Measure -> MeasRef -> MeasFrame
   const details::CyclicState state = frame.rep.Freeze();
   impl_->epConvLAST = MEpoch::Convert(*(frame.epoch()),
 				   MEpoch::Ref(MEpoch::LAST, frame));
@@ -472,6 +475,7 @@ void MCFrame::makeDirection(MeasFrame& frame) {
   static const MDirection::Ref REFJ2000 = MDirection::Ref(MDirection::J2000);
   static const MDirection::Ref REFB1950 = MDirection::Ref(MDirection::B1950);
 
+  // Initializing the MeasConvert may cause a cycle (see makeEpoch).
   details::CyclicState state = frame.rep.Freeze();
   impl_->dirConvJ2000 = MDirection::Convert(*frame.direction(),
           MDirection::Ref(MDirection::J2000, frame));
--- a/measures/Measures/MeasFrame.cc
+++ b/measures/Measures/MeasFrame.cc
@@ -436,6 +436,8 @@ void MeasFrame::create() {
 }
 
 void MeasFrame::fill(const Measure *in) {
+  // The Measure values may own a MeasRef that owns this MeasFrame, causing a
+  // cycle. The Freezing/Unfreezing calls prevent counting these links.
   if (in) {
     if (dynamic_cast<const MEpoch*>(in)) {
       const details::CyclicState state = rep.Freeze();
--- a/measures/Measures/test/tCyclicPtr.cc
+++ b/measures/Measures/test/tCyclicPtr.cc
@@ -77,6 +77,7 @@ BOOST_AUTO_TEST_CASE(non_empty) {
 }
 
 BOOST_AUTO_TEST_CASE(copy_empty) {
+  Counter::Reset();
   CyclicPtr<Counter> empty;
   CyclicPtr<Counter> empty_copy(empty);
   BOOST_CHECK(!empty_copy);


measures/Measures/MCFrame.cc
```diff
@@ -466,19 +468,22 @@ void MCFrame::makePosition(const MeasFrame& frame) {
...
   static const MDirection::Ref REFJ2000 = MDirection::Ref(MDirection::J2000);
+  details::CyclicState state = frame.rep.Freeze();
   impl_->dirConvJ2000 = MDirection::Convert(*frame.direction(),
-					 MDirection::Ref(MDirection::J2000,
-							 frame));
+          MDirection::Ref(MDirection::J2000, frame));
+  frame.rep.Unfreeze(state);
 
   static const MDirection::Ref REFB1950 = MDirection::Ref(MDirection::B1950);
+  state = frame.rep.Freeze();
```
mnijhuis-tos, line 479: Can't you remove these new statements, so only one Freeze..UnFreeze block remains?
```suggestion

  static const MDirection::Ref REFB1950 = MDirection::Ref(MDirection::B1950);
```

measures/Measures/MeasFrame.cc
```diff
@@ -415,24 +430,32 @@ Bool MeasFrame::getComet(MVPosition &tdb) const {
...
 void MeasFrame::create() {
   if (!rep) {
-    rep = std::make_shared<FrameRep>();
+    rep = details::MakeCyclic<FrameRep>();
     rep->mymcf = std::make_unique<MCFrame>();
   }
 }
 
 void MeasFrame::fill(const Measure *in) {
   if (in) {
```
mnijhuis-tos, line 439: Please add a comment here why you need to `Freeze()` and `UnFreeze()` here. As you explained in the description of this PR, the dependency cycle isn't trivial.

measures/Measures/MCFrame.cc
```diff
@@ -441,8 +441,10 @@ void MCFrame::makeEpoch(const MeasFrame& frame) {
   impl_->epTDBp.reset();
   impl_->epUT1p.reset();
   impl_->epTTp.reset();
+  const details::CyclicState state = frame.rep.Freeze();
```
mnijhuis-tos, line 444: Please document the cyclic dependency here, since it's not clear.

measures/Measures/MCFrame.cc
```diff
@@ -466,19 +468,22 @@ void MCFrame::makePosition(const MeasFrame& frame) {
   impl_->posGeop = MVPosition();
 }
 
-void MCFrame::makeDirection(const MeasFrame& frame) {
+void MCFrame::makeDirection(MeasFrame& frame) {
   static const MDirection::Ref REFJ2000 = MDirection::Ref(MDirection::J2000);
+  details::CyclicState state = frame.rep.Freeze();
```
mnijhuis-tos, line 473: Please document the cyclic dependency here, since it's not clear.

measures/Measures/CyclicPtr.h
```diff
@@ -0,0 +1,191 @@
...
+  explicit CyclicPtr(T* object) : data_(std::make_shared<Data>(object))
+  {}
+
+  CyclicPtr(const CyclicPtr& source) : data_(source.data_)
+  {
+    Increase();
+  }
+
+  CyclicPtr(CyclicPtr&& source) : data_(std::move(source.data_))
+  {}
```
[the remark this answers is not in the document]
aroffringa: I think it's an interesting design question... but I think I prefer the current approach.

Your suggestion would add allocation in the move constructor and reset method. Also the empty constructor would have to be changed, and then allocate. These are (currently) also used by MeasFrame.
- From complexity, not sure I prefer one over the other.
- From a performance perspective, the allocations can't be optimized away, and are relatively expensive. The if statement is cheap and in obvious cases can be optimized away... So I think I prefer emptying the shared_ptr when the object is empty.

shared_ptr has chosen the no-allocating approach (see https://gcc.gnu.org/onlinedocs/gcc-4.6.0/libstdc++/api/a01034_source.html line 00756) ; it does not allocate when empty constructing.

measures/Measures/MeasFrame.cc
aroffringa: Agree; though I don't know 100% sure why some things need the frozen counter (they're coming from the 'lock' statements that were there before #1426), but I've added the possible cycles I identified.

gervandiepen:
I always avoided using the Measures in a concurrent way. I took care each
thread had its own MeasFrame. This may also avoid that a thread has to wait
for another when using MeasFrame.

aroffringa:
> I always avoided using the Measures in a concurrent way. I took care each thread had its own MeasFrame. This may also avoid that a thread has to wait for another when using MeasFrame.

That's what we're trying to make possible, but just using different MeasFrames in different threads was not guaranteed thread safe (i.e. free of race conditions) due to various small and less small issues, see the recent pull requests. We're not trying to put mutexes in MeasFrame as I guess you're assuming.

measures/Measures/CyclicPtr.h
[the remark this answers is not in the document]
mnijhuis-tos: Ok, fair points.

PR #1431, opened 2025-07-31
Cleanup and smart pointer use in the MeasConvert class

aroffringa:


measures/Measures/MeasConvert.tcc
```diff
@@ -437,17 +409,15 @@ void MeasConvert<M>::setOut(typename M::Types mr) {
 
 template<class M>
 void MeasConvert<M>::set(const M &val, const typename M::Ref &mr) {
-  delete model; model = 0;
-  model = new M(val);
+  model = std::make_unique<M>(&val);
```
mnijhuis-tos, line 412: The original code passed the reference, and didn't convert it to a pointer.
```suggestion
  model = std::make_unique<M>(val);
```

measures/Measures/MeasConvert.tcc
```diff
@@ -437,17 +409,15 @@ void MeasConvert<M>::setOut(typename M::Types mr) {
...
   unit = val.unit;
   outref = mr;
   create();
 }
 
 template<class M>
 void MeasConvert<M>::set(const M &val, typename M::Types mr) {
-  delete model; model = 0;
-  model = new M(val);
+  model = std::make_unique<M>(&val);
```
mnijhuis-tos, line 420:
```suggestion
  model = std::make_unique<M>(val);
```

mnijhuis-tos (changes requested):
Using std classes simplifies this code. Nice changes!

I mainly checked if the new code should behave the same, and found two differences, where the new code passes a pointer instead of a reference.

PR #1434, opened 2025-08-22
FFT server fix

tammojan:
Supersedes #1282, local rebased branch.


dpetry:
@tammojan , just to confirm. Should this PR be used or the old one (#1282)?
If this one, shall I go ahead and review it and merge or would you like to do something additional?

tammojan:
Since the other one (#1282) is almost ready, let's continue there.

PR #1438, opened 2025-08-29
Sisco fixes: taql support and skipping rows

aroffringa:
Two fixes:

- Taql determines if a column has data by calling `isShapeDefined()`, which always used to return false in Sisco (because a defined shape is somewhat awkward in Sisco when writing). This is fixed.
- WSClean doesn't write all rows when not all rows are selected, introducing empty rows and empty shapes, and this wasn't correctly covered in reading/writing.


tammojan (approved):
Maybe add a test for taql writing and reading Sisco data?

aroffringa:
> Maybe add a test for taql writing and reading Sisco data?

My suggestion is to add a test to Dp3, because that is (at least for me) a bit easier to do...

PR #1443, opened 2025-10-02
Remap spws in SYSCAL table during concat

kettenis:


dpetry (commented):
Hi, thanks for getting to work on this quickly.
Can we exclude that the condition
   newSPWIndex_p.find(spwIDs[k]) != newSPWIndex_p.end()
is never True, i.e. for none of the possible values of k?
If not, I think we might need some error handling here.

kettenis:
I'm pretty sure that if there are references to non-existing spectral windows in the SYSCAL table, we will not update the spectral window correctly. So yes, some error handling would be appropriate. Probably should do what we do for mismatched antenna numbers here and drop the table in the concatenated MS.

Now I just copied and adjusted the code already in use for merging the GAIN_CURVE and PHASE_CAL tables.(I wrote that code, so I can't really shift the blame ;). So those cases need to be fixed as well.

kettenis:
CASA tests also still pass with this.

PR #1445, opened 2025-10-05
Fix SOFA tarball URL

rtobar:
The old URL doesn't work anymore, as it seems some redesign has occurred on the SOFA website. The precise version used before is also not currently available (http://www.iausofa.org/archive currently says that the archive is "coming soon").


tammojan (approved):
Thanks!

PR #1452, opened 2025-11-27
Fix read/write of Adios2StMan by row chunks

rtobar:
This PR fixes the `{get,put}ArrayColumnCellsV` implementation for the Adios2StMan. It previously was broken in subtle ways, which became evident during further testing. The PR adds a new class that iterates over the row slices in a `RefRows` object, adjusting the start/offsets of the underlying adios Variable, and returning the corresponding offset into the user-provided buffer. This is then used to correctly read/write consecutive row slices from/to the adios variable to/from the user-provided buffer.

No tests have been added at this stage, but this has been tested at a higher level externally via python-casacore's `table.{get,put}col(nrow=..., startrow=...)`.

### DIFF
--- a/tables/DataMan/Adios2StMan.cc
+++ b/tables/DataMan/Adios2StMan.cc
@@ -489,7 +489,6 @@ rownr_t Adios2StMan::impl::open64(rownr_t aNrRows, AipsIO &ios)
       ios >> dummy;
     }
     ios.getend();
-    itsRows = aNrRows;
     return itsRows;
 }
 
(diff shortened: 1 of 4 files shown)


rtobar:
Rebasing on top of latest `master` to ensure changes are still valid

tammojan (approved):

rtobar:
Hi @tammojan thanks for approving. I don't have write access to this repo, would you mind merging if you're happy with the changes? In the meantime I rebased once again, just making sure nothing is broken after the last change on `master`.

PR #1454, opened 2026-01-23
perf: avoid/defer construction of instances in MCFrame

tnakazato:
An attempt to improve the performance of frame conversion using measures. It turned out that performance degradation originated from recent update was actually due to costly constructor calls during (re-)initialization of `MCFrame` instance and associating `MCFrameImplementations` instance. Major change in this pull request is the use of `std::optional` to avoid/defer construction of some instances. The change should not affect the result.

I'm not sure this is sufficient to make the performance equivalent to the one before measures update. But I think it improves the performance at some level.


aroffringa:
Merged -- thanks a lot @tnakazato

PR #1464, opened 2026-03-06
Fix missing frame bias in IAU2000 JNAT<->APP conversions

bglenden:
## Summary

This fixes an IAU2000-only frame-bias omission in the shared combined
precession/nutation helper used by the `JNAT <-> APP` conversion paths.

With `measures.iau2000.b_use=true`, casacore's `J2000 -> APP` conversion
disagreed with the IAU SOFA reference calculation by about 16 mas for a
representative test direction. The root cause was that
`MeasMath::applyPrecNutat()` and `MeasMath::deapplyPrecNutat()` did not apply
`MeasTable::frameBias00()`, even though the standalone IAU2000 precession path
already does.

This is not a generic SOFA-vs-casacore discrepancy, and it does not affect the
older non-IAU2000 or B1950 paths.

## Root Cause

In the IAU2000 path:

- `MeasMath::applyPrecession()` applies `frameBias00()` and then precession.
- `MeasMath::applyPrecNutat()` previously applied only precession and
  nutation.

That meant the combined helper was using a different frame chain from the
sequential helper. The missing term is the IAU2000 frame-bias rotation, and
its projected size matches the observed ~16 mas discrepancy.

## Scope

This bug is in the shared helper, so the same fix applies to all IAU2000
`JNAT <-> APP` routes that call it:

- `MCDirection`
- `MCBaseline`
- `MCEarthMagnetic`
- `MCuvw`

The regression added in this PR demonstrates the problem on `MDirection`.
Because the other call sites use the same broken helper in the same part of the
route chain, they were affected by the same omission as well.

This is specific to the IAU2000 mode:

- affected: IAU2000A
- affected: IAU2000B
- unchanged: older non-IAU2000 precession/nutation path
- unchanged: B1950 path

## Reference Calculation

SOFA is the IAU-maintained reference software collection, with official
Fortran 77 and ANSI C implementations. ERFA is a BSD-licensed C library
derived from the SOFA C code and is used here as a practical fallback when
SOFA is not available.

The regression compares casacore `J2000 -> APP` against:

1. `pnm00a`
2. `epv00`
3. `ab`

That is sufficient to expose the missing frame-bias term.

casacore still differs from SOFA/ERFA at the mas level after the fix because
casacore uses full solar deflection and different Earth-velocity series, while
the SOFA/ERFA `ab()` step only includes a very small gravity term. That small
residual is expected; the ~16 mas jump is not.

## Before / After

Before the fix, the new regression reproduces a discrepancy of roughly:

- `16.17 mas` at J2000.0
- `16.08 mas` at 2020-01-01

After restoring the missing frame-bias step, the same check gives:

- `1.398 mas` at J2000.0
- `1.480 mas` at 2020-01-01

## Changes

- apply `MeasTable::frameBias00()` in
  `MeasMath::applyPrecNutat()` for the IAU2000 branch
- apply `MeasTable::frameBias00()` in
  `MeasMath::deapplyPrecNutat()` for the IAU2000 branch
- add `tIAU2000App` regression test against SOFA, or ERFA when SOFA is not
  available
- add minimal ERFA discovery for the new regression

## Testing

Verified with:

```sh
ctest --test-dir build-fix-iau2000 -R tIAU2000App --output-on-failure
```

and direct execution of `tIAU2000App`, which reports:

```text
J2000.0 separation to SOFA/ERFA reference: 1.39807 mas
2020-01-01 separation to SOFA/ERFA reference: 1.4799 mas
OK
```

## Background

(I (BG) discovered this while doing some AI assisted coding (converting casacore to
rust just for fun) and it found discrepancies in Measures while doing some cross-checking vs ERFA. I am not expert in this area, so while I tried to be careful, sorry if it's just nonsense for some reason).


VoronkovMA:
Just my 2c worth. It is certainly very plausible we have a real bug on the casacore's side. At the time of ASKAP's early science I had a casacore-based delay model working (now it uses CALC) and there were issues with accuracy which we had to work around by altering the conversion path slightly (and they were discovered the hard way in the first place). More recently, the question of the accuracy of UVWs appeared when we commissioned the new correlator for ATCA which writes data in an ASDM format. As it turned out, there is no place in the ASDM to store UVWs (although we do have them available in the online system). Instead, they're regenerated from first principles during the conversion of an ASDM into MS (via importasdm). And the accuracy was lacking.

aroffringa:
I can't really assess the fix itself, but the tests look good and are for me convincing enough that the fix is correct, as it compares correctly against ERFA. So I would recommend merging the PR. Any objections?

aroffringa (approved):

VoronkovMA:
I am happy for it to be merged. But it would be good to hear from actual VLBI astrometry experts.

aroffringa:
I don't see objections from the VLBI people, so I'll merge. Thanks for this, @bglenden !

bglenden:
Nice to make a technical contribution again after (probably) 30 years.

PR #1468, opened 2026-03-31
Handle deflate include directory

tammojan:
I think the current main branch only builds if libdeflate is in the standard include path. This fixes that, in a way somewhat consistent with the rest of the cmake modules in casacore. I also removed some legacy code in other modules.

### DIFF
--- a/cmake/FindDeflate.cmake
+++ b/cmake/FindDeflate.cmake
@@ -4,7 +4,7 @@ if(NOT DEFLATE_FOUND)
 
   find_path(DEFLATE_INCLUDE_DIR libdeflate.h)
   find_library(DEFLATE_LIBRARY NAMES deflate libdeflate)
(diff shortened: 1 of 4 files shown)


cmake/FindDeflate.cmake
```diff
@@ -0,0 +1,14 @@
+# Find libdeflate
+
+if(NOT DEFLATE_FOUND)
+
+  find_path(DEFLATE_INCLUDE_DIR libdeflate.h)
+  find_library(DEFLATE_LIBRARY NAMES deflate libdeflate)
+  mark_as_advanced(WCSLIB_INCLUDE_DIR WCSLIB_LIBRARY M_LIBRARY)
```
aroffringa: What is this for?

tammojan: Ahum, well caught, this should be deflate not wcslib. From what I understand `mark_as_advanced` makes sure the variable shows only under 'advanced variables' in `ccmake`.

aroffringa (approved):

PR #1470, opened 2026-04-03
Add a Sisco Stokes I mode

aroffringa:
When using Sisco for model data that has Stokes I values, it will still compress the values independently. Most of the redundancy gets compressed by Sisco, but not all, but what's more important, compressing 4 values takes about 4 times longer. By having a Stokes I mode, only a fourth of the data is sent to Sisco, leading to much faster compression.

Some results of Stokes I mode: in a simple test I find it is 3x faster to write Stokes I values compared to writing the full data and let Sisco find the redundancy – which is thus considerable.

Additionally, it saves 8% of space. This is because Sisco can find some of the redundancy in the polarizations when the data is Stokes I, but apparently not all.

### DIFF
--- a/tables/AlternateMans/SiscoReader.cc
+++ b/tables/AlternateMans/SiscoReader.cc
@@ -157,10 +157,10 @@ void SiscoReader::Request(size_t baseline_index, size_t n_values) {
     request.chunk = current_chunk_;
     request.n_values = 0;
     request_queue_.Push(std::move(request), lock);
-    lock.unlock();
     
     results_in_chunk_counter_ = 0;
     current_chunk_ = &chunks_.emplace_back();
+    lock.unlock();
     GetNextChunk(*current_chunk_);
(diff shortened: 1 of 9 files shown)


tables/AlternateMans/SiscoReader.cc
```diff
@@ -157,10 +157,10 @@ void SiscoReader::Request(size_t baseline_index, size_t n_values) {
     request.chunk = current_chunk_;
     request.n_values = 0;
     request_queue_.Push(std::move(request), lock);
-    lock.unlock();
     
     results_in_chunk_counter_ = 0;
     current_chunk_ = &chunks_.emplace_back();
+    lock.unlock();
```
aroffringa, line 163: This was actually a synchronization bug; the chunks_ member is written to without lock, and other threads also read/write to it.

tammojan (approved):

PR #1471, opened 2026-04-05
Make Block movable and deprecate PtrBlock

aroffringa:
This improves performance of Block slightly.

The deprecation of PtrBlock causes many changes in casacore. The motivation for PtrBlock is however confusing and based on old compiler behaviour, and removing it makes the code more consistent.
    
If Casa still uses PtrBlock, this change will cause deprecation warnings there, which can be solved using a find-replace of PtrBlock with Block.

The code also listed all the documentation in comments twice -- hence the removal of the comments.

### DIFF
--- a/casa/Containers/Block.h
+++ b/casa/Containers/Block.h
@@ -44,64 +44,6 @@
 
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
-// <summary>simple 1-D array</summary>
-// <use visibility=export>
-//
-// <reviewed reviewer="UNKNOWN" date="before2004/08/25" tests="" demos="">
-// </reviewed>
-//
-// <etymology>
-//     This should be viewed as a <em>block</em> of memory without sophisticated
-//     manipulation functions. Thus it is called <src>Block</src>.
-// </etymology>
-//
-// <synopsis>
-// <src>Block<T></src> is a simple templated 1-D array class. Indices are always
-// 0-based. For efficiency reasons, no index checking is done unless the
-// preprocessor symbol <src>AIPS_ARRAY_INDEX_CHECK</src> is defined. 
-// <src>Block<T></src>'s may be assigned to and constructed from other
-// <src>Block<T></src>'s.
-// As no reference counting is done this can be an expensive operation, however.
-//
-// The net effect of this class is meant to be unsurprising to users who think
-// of arrays as first class objects. The name "Block" is  intended to convey
-// the concept of a solid "chunk" of things without any intervening "fancy"
-// memory management, etc. This class was written to be
-// used in the implementations of more functional Vector, Matrix, etc. classes,
-// although it is expected <src>Block<T></src> will be useful on its own.
-//
-// The Block class should be efficient. You should normally use <src>Block</src>.
-//
-// <note role=warning> If you use the assignment operator on an element of this
-// class, you may leave dangling references to pointers released from
-// <src>storage()</src>.
-// Resizing the array will also have this effect if the underlying storage
-// is actually affected.
-// </note>
-//
-// If index checking is turned on, an out-of-bounds index will
-// generate an <src>indexError<uInt></src> exception.
-// </synopsis>
-//
-// <example> 
-// <srcblock>
-// Block<Int> a(100,0);  // 100 ints initialized to 0
-// Block<Int> b;         // 0-length Block
-// // ...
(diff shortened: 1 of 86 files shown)


tammojan (commented):
Looks like a nice cleanup to me. I have not checked the move constructor yet.

casa/Containers/Block.h
```diff
@@ -327,6 +266,16 @@ template<class T> class Block: public BlockTrace
...
   }
   
+  Block(Block<T> &&other) noexcept :
+      allocator_p(other.allocator_p), capacity_p(other.capacity_p), used_p(other.used_p),
+          array(other.array), destroyPointer(other.destroyPointer) {
+    // The allocator of other is left as is; the storage is emptied so no need to change it.
+    other.capacity_p = 0;
+    other.used_p = 0;
+    other.array = nullptr;
+    other.destroyPointer = true;
```
mnijhuis-tos: Optional:
The value of `destroyPointer` doesn't matter, since `array` also becomes null, and functions like `deinit()` and `dealloc()` check both `array` and `destroyPointer`. Still, `false` would be more logical, since there's nothing left to destroy in `other`.
```suggestion
    other.destroyPointer = false;
```

casa/Containers/Block.h
```diff
@@ -343,8 +292,27 @@ template<class T> class Block: public BlockTrace
...
+    allocator_p = other.allocator_p;
+    capacity_p = other.capacity_p;
+    used_p = other.used_p;
+    destroyPointer = other.destroyPointer;
+    array = other.array;
+    // The allocator of other is left as is; the storage is emptied so no need to change it.
+    other.capacity_p = 0;
+    other.used_p = 0;
+    other.array = nullptr;
+    other.destroyPointer = true;
```
mnijhuis-tos: Optional, see comment in move constructor.
```suggestion
    other.destroyPointer = false;
```

casa/Containers/Block.h
```diff
@@ -782,82 +689,31 @@ template<class T> class Block: public BlockTrace
   T *array;
   // Can we delete the storage upon destruction?
   Bool destroyPointer;
-  // Can we change allocator or not?
-  Bool keep_allocator_p;
```
mnijhuis-tos, line 786: Nice removal. Since `prohibitChangingAllocator` was never called, except in the test, removing it should be safe. In case anybody still uses `prohibitChangingAllocator`, they should simply remove it.

Since most code supposedly doesn't use `Block` directly, I'm even unsure if we should mention in the change log that `Block::prohibitChangingAllocator`  and `Block::permitChangingAllocator` were removed.

casa/Containers/Block.h
```diff
@@ -665,51 +612,11 @@ template<class T> class Block: public BlockTrace
   }
 
  private:
-  friend class Array<T>; // to allow access to following constructors.
```
mnijhuis-tos, line 668: I'm surprised you could remove this `friend` and the code below. If CI accepts it, it's just another nice removal.

casa/Containers/Block.h
```diff
@@ -327,6 +266,16 @@ template<class T> class Block: public BlockTrace
     }
   }
   
+  Block(Block<T> &&other) noexcept :
```
mnijhuis-tos, line 269: Adding the move constructor and move assignment operator are nice improvements. Can you also add tests for them?

casa/Containers/Block.h
aroffringa: The reason to use true is because the default constructor also uses true. I do agree false is slightly nicer, so I'll change that :). Many of the constructors could actually also do that, but might be good to not introduce changes that could possible change behaviour of Block inside this PR.

casa/Containers/Block.h
aroffringa: The Array classes were extensively updated for C++11 at some point, I think that changed its dependence on this class, but I guess this friend declaration wasn't noticed at the time.

casa/Containers/Block.h
aroffringa: I've added a simple test. The Containers module uses the rather outdated testing approach so I think it is good to keep those tests simple...

casa/Containers/Block.h
aroffringa: Given it could be a breaking change, mentioning it can't hurt. I've added it to the changelog.

mnijhuis-tos (approved):

PR #1472, opened 2026-04-05
Solve a few Clang warnings

aroffringa:


mirlib/uvio.c
```diff
@@ -760,7 +760,7 @@ void uvopen_c(int *tno,Const char *name,Const char *status)
     CHECK(iostat,(message,"Error accessing visdata for %s, in UVOPEN(old)",name));
 #ifdef MIR4
     /* figure out if to read old MIR3 or new MIR4 */
-#if true
+#if 1
```
aroffringa, line 763: this one is suspicious... true is an unknown symbol during preprocessing, so evaluates to zero (with a warning). Hence this used to call rdhdi_c (=int form) instead of RDHDL_C (=long form). I think reading a long is intended so this was actually a bug (though probably not causing any effect).

python/Converters/PycBasicData.h
```diff
@@ -556,7 +556,7 @@ namespace casacore { namespace python {
         bool is_same = PyRange_Check(obj_ptr) ||
                   (PySequence_Check(obj_ptr)
                && !PyTuple_Check(obj_ptr) && !PyList_Check(obj_ptr));
-    int i = 0;
+    [[maybe_unused]] int i = 0;
```
tammojan, line 559: Could be a while true here as well, like in `PycRecord`?

tammojan (commented):
Thanks, looks good!

python/Converters/PycBasicData.h
aroffringa: Variable i is used after the loop in an assert, so it can't be completely eliminated. However I agree that a while true loop is better describing what it does, so I changed that.

tammojan (approved):

PR #1473, opened 2026-04-05
Fix tests when running on Debian with debugging asserts

aroffringa:
A few tests failed under Debian, and a few others failed when building with NDEBUG (such that debug asserts run).
Fixes:
- MultiFile would, via STLIO, call the indexing operator on an empty std::vector to get the address of the vector buffer ( &vec[0] ). This works (when debug asserts are off) but is officially not allowed on an empty vector; changed to vec.data(), which is allowed.
- casacore_floatcheck, one of the python testing helper scripts, has a shebang line calling python instead of python3. This doesn't work on Debian.
- One of the Dysco tests to see if Dysco doesn't crash when writing past the end of the table. In debug mode, this causes somewhere in casa an assert to fail in a check for when writing past the end of file, so this test does something which it shouldn't do.

After these changes, all tests succeed both with or without debug asserts.

### DIFF
--- a/build-tools/casacore_floatcheck
+++ b/build-tools/casacore_floatcheck
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 #-----------------------------------------------------------------------------
 # floatcheck: Check files within floating point accuracy
--- a/casa/BasicSL/STLIO.tcc
+++ b/casa/BasicSL/STLIO.tcc
@@ -56,7 +56,7 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
     uInt nr;
     ios >> nr;
     v.resize(nr);
-    getAipsIO(ios, nr, &(v[0]));
+    getAipsIO(ios, nr, v.data());
     ios.getend();
     return ios;
   }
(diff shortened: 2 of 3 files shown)


tammojan (approved):
Just disabling the test is a bit dubious; how are we sure that the thing the assert tries to check does not lead to errors in non-debug mode?

aroffringa:
> Just disabling the test is a bit dubious; how are we sure that the thing the assert tries to check does not lead to errors in non-debug mode?

The test that was disabled seems to explicitly test something which Casacore doesn't allow (read past the end of the MS). It's something that I added while testing Sisco, as I think at some point DP3 did read/write past the end and Sisco crashed. This was then fixed such that Sisco doesn't crash -- but you're right that DP3 may still trigger that assert if it reads/writes past the end of an MS. If that's the case (I haven't seen it since) it means, because Casacore doesn't allow it, it's a bug in DP3.

tammojan:
> > Just disabling the test is a bit dubious; how are we sure that the thing the assert tries to check does not lead to errors in non-debug mode?
> 
> The test that was disabled seems to explicitly test something which Casacore doesn't allow (read past the end of the MS). It's something that I added while testing Sisco, as I think at some point DP3 did read/write past the end and Sisco crashed. This was then fixed such that Sisco doesn't crash -- but you're right that DP3 may still trigger that assert if it reads/writes past the end of an MS. If that's the case (I haven't seen it since) it means, because Casacore doesn't allow it, it's a bug in DP3.

Thanks, that makes sense.

PR #1476, opened 2026-04-20
Switch to new EOP files

tammojan:
The eopc04_14 timeseries for Earth Orientation Parameters was no longer [updated](https://hpiers.obspm.fr/iers/eop/eopc04_14/eopc04_IAU2000.26) after 2026-01-05. With this MR, we switch to the EOPC04_20 timeseries. This MR is aimed to be rolled out at ASTRON, where 'WSRT_Measures.ztar' is created.

### DIFF
--- a/measures/apps/measuresdata/measuresdata.cc
+++ b/measures/apps/measuresdata/measuresdata.cc
@@ -436,7 +436,7 @@ const tableProperties allProperties[] = {
     "IERS EOPC04_20 Earth Orientation Data from IERS",	"eop97",
     True,			"",			vector<String>(),
     0, 				vector<formatDescr>(),
-    { "hpiers.obspm.fr", "iers/eop/eopc04", "eopc04.1962-now" } },
+    { "hpiers.obspm.fr", "iers/eop/eopc04", "eopc04.dPsi_dEps.1962-now" } },
 
   //**********************************************************************//
 
--- a/measures/apps/measuresdata/measuresdata.cc
+++ b/measures/apps/measuresdata/measuresdata.cc
@@ -436,7 +436,7 @@ const tableProperties allProperties[] = {
     "IERS EOPC04_20 Earth Orientation Data from IERS",	"eop97",
     True,			"",			vector<String>(),
     0, 				vector<formatDescr>(),
-    { "hpiers.obspm.fr", "iers/eop/eopc04", "eopc04.dPsi_dEps.1962-now" } },
+    { "hpiers.obspm.fr", "iers/eop/eopc04", "eopc04.dPsi_dEps.IAU1980.1962-now" } },
 
   //**********************************************************************//
 
--- a/measures/apps/measuresdata/measuresdata.cc
+++ b/measures/apps/measuresdata/measuresdata.cc
@@ -55,6 +55,7 @@
 #include <casacore/casa/stdio.h>
 #include <map>
 #include <vector>
+#include <algorithm>
 
 #include <casacore/casa/namespace.h>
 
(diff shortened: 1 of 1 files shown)


gmloose:
Are these data from the same source? I mean, could we expect a sudden jump in directions and positions?

tammojan:
There are some jumps in some of the fields, but not in the x and y, which are (I think) the only ones that are used. The jumps are on January 2026, which is the point where the new data starts.
IERSeop97
IERSeop2000

This is work in progress (together with @dpetry).

dpetry:
Hi Tammo, 
thanks for posting the plots.

I still have some doubts about the dPsi and dEps values. All others are fine I think.

There seems to have been a change at the beginning of 2026 (around MJD 61050) in the procedure of how the error estimates of the various quantities are calculated.

dpetry:
@tammojan , hi again, please do a test and produce the plots again but using the file
    eopc04.dPsi_dEps.1962-now
instead of 
   eopc04.1962-now
(in line 439 of the code)

I checked that you shouldn't have to change the column numbers. They are the same in both tables.

tammojan:
This is with `hpiers.obspm.fr", "iers/eop/eopc04", "eopc04.dPsi_dEps.1962-now`

IERSeop97

dpetry:
Mmh. At first glance, this looks exactly the same as for eopc04.1962-now .

The suspicious thing is that dPsi and dEps jump to nearly zero and after that the variations seem to be much smaller.

Looks like they changed the units. (But the unit in the table header is still indicated as arcsec.)

Maybe degrees instead of arcsec?
Could you try multiplying the two values by 3600?

On the other hand, the error estimates of the values seem to have become larger if anything!
So, probably no change of units there.

I sent an email to Paris again to ask about this ...

kettenis:
Looking at the archived IERS data product I see that the only EOP 14 C04 data products offered there that include dPsi and dEps are those based on the IAU1980 precession/nutation model. The EOP 20u24 C04 data products are all based on the IAU2000A precession nutation model. That could explain the jump in dPsi and dEps you're seeing. That said, I don't fully understand what dPsi and dEps in the context of IAU2000A mean, as my understanding of IAU2000A is that it eliminated those parameters.

dpetry:
OK, I got confirmation from the Observatoire de Paris. They say
"
eopc04.dPsi_dEps.1962-now  is referred to IAU 2000 model.
 
eopc04_14/eopc04.26  is referred to IAU 1980 model, leading to much larger celestial pole offsets.

That's all.
"
So we can conclude that all is fine.

dpetry:
But we are updating two tables here:
One explicitly based on the IAU2000 model - that update is working.
And one based on the older model.
In a second email, I was told by the Observatoire de Paris now that we should consider using
   https://hpiers.obspm.fr/iers/eop/eopc04/eopc04.dPsi_dEps.IAU1980.1962-now

And I find that the dPsi and dEps values here agree much better with ours.

So I suggest to replace 
eopc04.1962-now
(in line 439 of the code)
by 
eopc04.dPsi_dEps.IAU1980.1962-now

dpetry (changes requested):
I double checked all the column numbers and your test runs confirm that this is working.
The discontinuities in the plots which you sent me by email
occur around the beginning of January
and only concern the error estimates as far as I can see.
This points to a change in procedure in Paris.

The large jump of the dPsi and dEps values has been confirmed by
Christian.Bizouard@obspm.fr to be due to the model change.
We should use file 
eopc04.dPsi_dEps.IAU1980.1962-now
instead of
eopc04.1962-now

dpetry:
@tammojan , sorry, but the file name is still wrong. It should be 
   eopc04.dPsi_dEps.IAU1980.1962-now
to retain the properties of the 97 table.
You have inserted eopc04.dPsi_dEps.1962-now .

dpetry (changes requested):
On line 438, it should be 
    eopc04.dPsi_dEps.IAU1980.1962-now

Please produce the parameter plots one last time with this change. Thanks!

dpetry:
Hi @tammojan , we discussed further at the Measures WG telecon today and all agree that
using eopc04.dPsi_dEps.IAU1980.1962-now for the 97 table is the way to go for the moment.

Concerning the nice plots you produced, we just want to make sure we understand:
Until 5 Jan 2026, you are plotting the present value of the IERSeop97 and 2000 tables respectively, right?
Until that date, you are _not_ plotting the new values taken from Paris, right?

After 5 Jan 2026, of course you are plotting the new values from Paris taken from the corresponding input table.

This means that the one remaining jump we are seeing for dPsi and dEps in the 97 table at the moment,
is _not_ a feature of the Paris data but caused by the switch from one input table to another at that date, right?

dpetry:
Anyway, everybody at CASA and ALMA is eagerly waiting for this pull request to be resolved.
Please make the last requested change and the diagnostic plots as soon as convenient for you
so I can sign off on it and the procedure update can take place.

Meanwhile, the CASA team wants to investigate which of the two tables IERSeop97 and 2000 is actually being
used downstream for the important tasks. And if it is IERSeop97, we need to think of a path towards switching
to using IERSeop2000.

dpetry (approved):
Thanks for making this last change.
Could you produce the continuity plots again and post them here before you deploy this at ASTRON?

dpetry:
@tammojan , I'm sick at home today. You, too?
I hesitate to merge this pull request before having seen your continuity plots made with this latest version.
Could you still produce them?

tammojan:
This is the continuity plot with the latest URL `eopc04.dPsi_dEps.IAU1980.1962-now`.
iersop97

dpetry:
Looks good! As expected. Please go ahead and deploy!

tammojan:
I updated the astron machine to run docker with a casacore build which has this update. I've run a manual `measuresdata-update` now and deployed it to https://www.astron.nl/iers/, the daily update should work as well. I'll check that the coming days. I'll make a pull request for the updated cron jobs etc.

PR #1477, opened 2026-04-21
Sync the Dysco code with the github repo

aroffringa:
The [Dysco github repo](https://github.com/aroffringa/dysco) uses the updated UVector from aocommon (rename from ao::uvector to aocommon::UVector) and has applied clang formatting with 100 chars/line to UVector. This PR syncs those changes to keep the code (almost) identical. There's no change in behaviour.

The Dysco repo still doesn't include the 64-bit rownr support to make it build with ancient casacore versions, which results in minor differences.


gmloose (approved):
LGTM.

PR #1479, opened 2026-04-24
Fix "no matching function call to arrayTransform" in Array2Math.cc

vsuorant:
This fixes build error with Apple clang 21 that fails with "error: no matching function for call to 'arrayTransform'
97 | arrayTransform (carray, rarray, std::abs<float>);" in Array2Math.cc.


aroffringa (approved):
Not sure if the compiler is correct here, but solution seems fine to me :). I'll merge the PR.

PR #1480, opened 2026-06-11
Update GitHub actions and docker build

gmloose:
* Updated GitHub workflow to use latest actions, to avoid deprecation errors (or worse).
* Updated Docker build: 
  * dropped Python 3.8 and 3.9, added Python 3.14; 
  * updated manylinux image;
  * dropped docker build/test for Ubuntu 22.04, added Ubuntu 26.04


tammojan (approved):
Nice work, thanks!

PR #1481, opened 2026-06-22
Build Sisco by default

gmloose:
Enable Sisco by default. 

This change requires that we build `libdeflate` ourselves for our Docker base images, because the version in the `manylinux` base image repository is too old.

### DIFF
--- a/tables/AlternateMans/SiscoWriter.cc
+++ b/tables/AlternateMans/SiscoWriter.cc
@@ -269,7 +269,7 @@ void SiscoWriter::WriteLoop()
   assert(scheduled_tasks.empty());
 }
 
-void SiscoWriter::WriteChunk(uint64_t uncompressed_size, std::span<const std::byte> data) {
+void SiscoWriter::WriteChunk(size_t uncompressed_size, std::span<const std::byte> data) {
   file_.write(reinterpret_cast<const char*>(&uncompressed_size), sizeof(uncompressed_size));
   const uint64_t compressed_size = data.size();
   file_.write(reinterpret_cast<const char*>(&compressed_size), sizeof(compressed_size));
--- a/.github/workflows/osx.yml
+++ b/.github/workflows/osx.yml
@@ -26,7 +26,7 @@ jobs:
           # Delete GitHub Python, See https://github.com/orgs/Homebrew/discussions/3895#discussioncomment-4130560
           find /usr/local/bin -lname '*/Library/Frameworks/Python.framework/*' -print -delete
 
-          brew install bison fftw hdf5 boost-python3 numpy cfitsio wcslib gsl ninja gcc
+          brew install bison fftw hdf5 boost-python3 numpy cfitsio wcslib gsl ninja gcc libdeflate
 
           # ensure we use homebrew python and gcc
           export PATH=$(brew --prefix python)/bin:${PATH}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,7 +46,7 @@ option (BUILD_PYTHON3 "Build the python3 bindings" YES)
 option (BUILD_DEPRECATED "build and install deprecated classes (such as Map)" NO)
 option (BUILD_FFTPACK_DEPRECATED "build FFTPack" NO)
 option (BUILD_DYSCO "Build the Dysco compression storage manager" YES)
-option (BUILD_SISCO "Build the Sisco compression storage manager" NO)
+option (BUILD_SISCO "Build the Sisco compression storage manager" YES)
 
 # By default build shared libraries
 option (ENABLE_SHARED "Build shared libraries" YES)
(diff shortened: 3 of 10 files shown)


kettenis:
Is this just "build Sisco by default"? Or does this actually result in creating MSv2s that enable Sisco by default?

aroffringa:
> Is this just "build Sisco by default"? Or does this actually result in creating MSv2s that enable Sisco by default?

It only makes it build by default ; once build, the Sisco storage manager still needs to be explicitly selected to use it. Sisco is mostly aimed at compressing noisless data (e.g. model data).

kettenis:
> > Is this just "build Sisco by default"? Or does this actually result in creating MSv2s that enable Sisco by default?
> 
> It only makes it build by default ; once build, the Sisco storage manager still needs to be explicitly selected to use it. Sisco is mostly aimed at compressing noisless data (e.g. model data).

Good, I'd probably change the commit message to "Build Sisco by default" to avoid confusion.

tables/AlternateMans/SiscoWriter.cc
```diff
@@ -269,7 +269,7 @@ void SiscoWriter::WriteLoop()
   assert(scheduled_tasks.empty());
 }
 
-void SiscoWriter::WriteChunk(uint64_t uncompressed_size, std::span<const std::byte> data) {
+void SiscoWriter::WriteChunk(size_t uncompressed_size, std::span<const std::byte> data) {
```
aroffringa, line 272: It's a bit dangerous to slip in a change like this in a build-system PR ;) ; I checked it though and I think this is okay, as `size_t` is used consistently. This does mean that 32-bit systems won't support files > 4 Gb, but making that possible is more work and probably not worth the effort.

Btw how come the Mac build uses a 32-bit system?

aroffringa (approved):

aroffringa:
Do note that this MR:
- Will cause casacore to require libdeflate to be present
- Enables C++20, and therefore requires a slightly newer version of gcc.

Personally I think it is fine, but I don't know if everyone agrees. It may also be good to document this.

tables/AlternateMans/SiscoWriter.cc
gmloose: Agree. But it was a one-liner and looked innocent enough. I assumed the declaration was correct, and that the implementation needed fixing.
I don't know if/why Mac uses a 32-bit system. But the compiler barfed, saying it couldn't find a matching function, which only makes sense if `size_t` and `uint64_t` are different types.

gmloose:
> Do note that this MR:
> 
>     * Will cause casacore to require libdeflate to be present
> 
>     * Enables C++20, and therefore requires a slightly newer version of gcc.
> 
> 
> Personally I think it is fine, but I don't know if everyone agrees. It may also be good to document this.

I'm happy to disable it again. At least we now know it works _if_ we decide to enable it. Who should we ask for a final verdict if these extra requirements are too much of a burden?

aroffringa:
There's not really a good communication system in place to discuss Casacore design choices -- people that care probably follow the pull requests to some level (like kettenis above). So if no one shouts I would merge it, with the caveat that if it causes major issues for someone we may have to look at it again.

aroffringa:
Maybe one more addition: If the version of libdeflate is too old, I don't think that an understandable error is printed? This might be something that is good to do, otherwise we might see that people report this as a bug.

tables/AlternateMans/SiscoWriter.cc
kettenis: So probably `size_t` is `long` and `uint64_t` is `long long` (or the other way around) on macOS. Might be better to use `uint64_t` in the declaration such that this is a 64-bit type everywhere.

gmloose:
> Maybe one more addition: If the version of libdeflate is too old, I don't think that an understandable error is printed? This might be something that is good to do, otherwise we might see that people report this as a bug.

I'm happy to fix the CMake module. Do you know the minimal version you need? I should probably do that in a separate PR.

tables/AlternateMans/SiscoWriter.cc
aroffringa: I agree, Macos should be using the LP64 model, and so size_t, long and long long should be all the same size, and probably the compiler does not accept interchanging long and long long even though they are the same size.

I agree uint64_t is probably better, but I think that's better not done in this MR. Also if all of these types are 64-bit, it doesn't really matter, so I don't think it's not worthwhile.

aroffringa:
The version in Ubuntu 22 is too old (libdeflate-dev version is 1.10).
The version of Ubuntu 24 works (libdeflate-dev version is 1.19).
So 1.19 as lower limit seems like a good setting.

tables/AlternateMans/SiscoWriter.cc
kettenis: Getting off-topic here, but it is Linux that does things wrong. Making `uint64_t` `unsigned long long` means that type is consistent across 32-bit and 64-bit architectures.

PR #1483, opened 2026-06-23
Add CMake Find module for LibDeflate

gmloose:
Added a new CMake Find module for `libdeflate`. This module:
  * Replaces `FindDeflate.cmake`.
  * Matches with the config module that ships with newer `libdeflate` packages, which makes a smooth transition easier.
  * Also checks for the correct version number of `libdeflate`.

The `CMakeLists.txt` files that used `FindDeflate.cmake` have been updated to use the new `FindLibDeflate.cmake`.


aroffringa (approved):
Nice! Looks all good to me.

gmloose:
I made some changes and had to close/reopen the PR in order to get rid off the status checks.

PR #1486, opened 2026-06-25
Require a C++-20 compatible compiler

gmloose:
Make the use of a C++-20 compatible compiler required unconditionally. Before, when not building Sisco, C++-17 would suffice, but it is better to always use the same compatibility level.


tammojan (approved):
Approved by me, but I'd like it if someone from the CASA side could comment here.

aroffringa (approved):
Given no objections I think it is fine to merge.

PR #1491, opened 2026-08-09
Remove 'using std::string' et al from String.h

aroffringa:
This doesn't change every occurrence of std in header files, it just changes String.h and fixes (using find and replace) non-std usage in parts of the code that failed.

### DIFF
--- a/casa/BasicSL/String.h
+++ b/casa/BasicSL/String.h
@@ -26,16 +26,10 @@
 #ifndef CASA_STRING_H
 #define CASA_STRING_H
 
-//# Includes
 #include <casacore/casa/aips.h>
 
-//# Includes
+#include <sstream>
 #include <string>
-
-using std::string;
-
-#include <casacore/casa/iosstrfwd.h>
-#include <casacore/casa/sstream.h>
 
 namespace casacore { //# NAMESPACE CASACORE - BEGIN
 
@@ -61,7 +55,7 @@ public:
   //# Friends
   friend class String;
   // Make a string
-  operator const string() const { return string(ref_p, pos_p, len_p); }
+  operator const std::string() const { return std::string(ref_p, pos_p, len_p); }
   // Default copy constructor.
   SubString (const SubString&) = default;
   // Assignment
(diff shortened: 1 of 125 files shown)


casa/BasicSL/String.h
```diff
@@ -220,90 +214,90 @@ class SubString {
...
+
+  typedef std::string::reference 		reference;
+  typedef std::string::const_reference 	const_reference;
+  typedef std::string::pointer		pointer;
+  typedef std::string::const_pointer		const_pointer;
+
+  typedef std::string::iterator iterator;
+  typedef std::string::const_iterator const_iterator;
+  typedef std::string::reverse_iterator reverse_iterator;
+  typedef std::string::const_reverse_iterator const_reverse_iterator;
```
mnijhuis-tos, line 236: Optional:
- Convert these `typedef`s into `using` statements, since you're changing them anyway.
- Or, even better: Try removing these `typedef`s completely, since they merely duplicate definitions from the base `std::string` class. The class can probably already use these definitions without redefining them.

ms/MSSel/test/tMSAntennaGram.cc
mnijhuis-tos, line 1: Please convert `cout` to `std::cout` as well in this file.

ms/MSSel/test/tMSAntennaGram2.cc
mnijhuis-tos, line 1: Please convert `cout` to `std::cout` as well in this file.

ms/MSSel/test/tMSAntennaGram3.cc
mnijhuis-tos, line 1: Please convert `cout` to `std::cout` as well in this file.

ms/MSSel/MSSpwIndex.cc
```diff
@@ -122,7 +122,7 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
     if (IDs.nelements() != sourceId.nelements())
       {
 	vector<int> tt = IDs.tovector();
```
mnijhuis-tos, line 124: I noticed you changed `vector` to `std::vector` elsewhere. Should we update all `vector`s as well?
```suggestion
	std::vector<int> tt = IDs.tovector();
```

mnijhuis-tos (approved):
- In the beginning, I see you replaced `cout` by `std::cout`, however, later I still see `cout`. Since it seems you've replaced all `endl` by `std::endl`, please consider replacing `cout` as well.
- You can then also remove `using std::cout;` and `using std::endl;` in casa/iostream.h.
- Please consider removing `using std::[io]stringsteam;` as well.
- I see there are many other `using std::xxx` in casacore. Are you planning removing all of them?

casa/BasicSL/String.h
aroffringa: I intend to deprecate the entire String class in the end (work in progress #1493 ), to remove it in the end. Hence I rather do the minimal required now to the class.

ms/MSSel/test/tMSAntennaGram.cc
aroffringa: I haven't changed all endls, I did a somewhat targetted find-and-replace for both to make sure that the code compiles again. I rather do this in steps; the intend is to slowly remove all unnecessary `using namespace std` and such, so this will be dealt with in further PRs.

ms/MSSel/MSSpwIndex.cc
aroffringa: There are usages of std without std:: of many types throughout casacore, I don't want to cover them all in one PR but do this step by step.

PR #1493, opened 2026-08-10
Add new functions that can replace extra String functionality

aroffringa:
This PR also adds a macro that can be defined to deprecate all extra functions. This is for now not enabled as this would cause many warnings. Once the extra functions can be replaced, the deprecation can be turned on.

This PR also removes all forwarding methods to std::string from casacore::String; they are not necessary, as String publically derives from std::string. For overloaded methods, they are now made available with `using` statements.


casa/BasicSL/test/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ tComplex
 tConstants
 tSTLIO
 tSTLMath
-tString
+tStringOld
```
mickveldhuis, line 6: You removed tString.cc from this list, but the file is still there. Should that be removed?

mickveldhuis: Ah I see it's moved to `casa/CMakeLists.txt`, why is it moved?

casa/BasicSL/test/tStringOld.cc
```diff
@@ -0,0 +1,494 @@
...
+#include <unordered_set>
+
+#include <casacore/casa/namespace.h>
+// Generally used variables
+String X = "Hello";
+String Y = "world";
+String N = "123";
+String c;
+const Char*  s = ",";
+Regex  r ("e[a-z]*o");
```
mickveldhuis, line 46: Should this be moved to an anonymous namespace?

casa/BasicSL/test/tStringOld.cc
```diff
@@ -0,0 +1,494 @@
...
+
+  /*
+  AlwaysAssertExit(X.OK());
+  AlwaysAssertExit(Y.OK());
+  AlwaysAssertExit(x.OK());
+  AlwaysAssertExit(y.OK());
+  AlwaysAssertExit(z.OK());
+  AlwaysAssertExit(n.OK());
+  AlwaysAssertExit(r.OK());
+  */
```
mickveldhuis: Why is this commented out?

casa/BasicSL/test/tStringOld.cc
```diff
@@ -0,0 +1,494 @@
...
+  };
+
+  //# remove cout's of hash values because they can differ between arch's
+  for (xp=x; xp->length() > 0; xp++)  xp->hash();
+  //  cout << "hash(" << *xp << ") = " << xp->hash() << endl;
+
+
+  a = ""; a.hash();
+  //  cout << "hash() = " << a.hash() << endl;
+
```
mickveldhuis, line 453: Same here, why is this commented out? Can this be removed?

casa/BasicSL/String.cc
```diff
@@ -27,24 +27,54 @@
...
+    iter = str.end() - 1;
+    while (iter != str.begin()  &&  std::find(characters.begin(), characters.end(), *iter) != characters.end()) {
+        --iter;
+    }
+    ++iter;
+    str.erase (iter, str.end());
+  }
+}
+
+std::string FormatString (const char* picture, ...)
```
mickveldhuis, line 54: Why is the first parameter called a `picture`?

casa/BasicSL/String.h
```diff
@@ -1009,21 +950,184 @@ inline SubString::SubString(const std::string &str, std::string::size_type pos,
...
+  while(p != str.end()) {
+    bool at_word;
+    if (islower(*p)) {
+      *p = toupper(*p);
+      at_word = true;
+    } else {
+      at_word = isupper(*p) || isdigit(*p);
+    }
+    ++p;
+    // atword is now true if the previous *p is a character or digit
```
mickveldhuis: I guess that should be:
```suggestion
    // at_word is now true if the previous *p is a character or digit
```

casa/CMakeLists.txt
```diff
@@ -649,7 +649,68 @@ install (FILES ${top_level_headers}
 DESTINATION include/casacore/casa
 )
 add_subdirectory (test ${EXCL_ALL})
-foreach (casa_module Arrays BasicMath BasicSL Containers
```
mickveldhuis, line 652: Why is `Arrays` removed here?

mickveldhuis (commented):
Looks good to me overall! Just have a few small questions.

casa/BasicSL/test/CMakeLists.txt
aroffringa: It's now a boost unit test instead of the casacore style test (the casacore like tests only compare the console output instead of have proper tests, which is rather hard for debugging). Because the casa module already had other boost unit tests that I wrote previously (from the "arrays" directory), it was easiest to combine those.

casa/BasicSL/test/tStringOld.cc
aroffringa: tStringOld.cc is what was first in tString.cc ; I didn't change this. The idea is that tStringOld.cc can be largely removed once we remove the String class, and tString.cc will test all new free functions.

I guess this is commented out because the String class no longer has an OK() method... but that is probably done long ago.

aroffringa: (I'll remove it)

casa/BasicSL/test/tStringOld.cc
aroffringa: I'm not sure, but the Hash function for String will be removed once String is removed, so I don't think we have to resuscitate this test...

casa/CMakeLists.txt
aroffringa: Arrays has only boost tests, and those tests moved up inside casa/CMakeLists.txt (so that they could be merged with the new boost BasicSL/String.h tests), so that it doesn't need to be separately called here.

casa/BasicSL/test/tStringOld.cc
aroffringa: Good idea, fixed

casa/BasicSL/String.cc
aroffringa: I copied it from the old function String::Format... but that's a good question. I've renamed it to "format_string".

casa/BasicSL/String.h
aroffringa: Sharp :) . Fixed.

mickveldhuis (approved):

PR #1496, opened 2026-08-12
Replace non-std string use in measures

aroffringa:


measures/apps/measuresdata/measuresdata.cc
```diff
@@ -1280,7 +1280,7 @@ Bool IERSpred(tableProperties &tprop, inputValues &inVal) {
     if (lines[i].size() >= tprop.fdesc.back().start +  tprop.fdesc.back().n) {
       field.resize(0);
       for (uInt j=0; j<tprop.fdesc.size(); ++j) {
-	field.push_back(lines[i].at(tprop.fdesc[j].start, tprop.fdesc[j].n));
+	field.push_back(lines[i].substr(tprop.fdesc[j].start, tprop.fdesc[j].n));
```
tnakazato: When `pos` (1st arg) exceeds the size of the string, `string::substr` throws `std::out_of_range` exception. 

https://en.cppreference.com/cpp/string/basic_string/substr

On the other hand, `String::at` adjusts the start position to suppress the exception (possibly). 

https://github.com/casacore/casacore/blob/master/casa/BasicSL/String.h#L945-L949

Therefore, I'm not sure if this replacement is straightforward in general. Although it's an edge case,  possibly we need to check if `pos` is in the correct range.

measures/Measures/MDirection.cc
```diff
@@ -477,7 +477,7 @@ String MDirection::toString() const {
 	) {
 		String ra = MVTime(longitude).string(MVTime::TIME, 12);
 		String dec = MVAngle(abs(lat)).string(MVAngle::ANGLE_CLEAN, 11);
-		dec.trim();
+		TrimInPlace(dec);
```
tnakazato, line 480: not directly related to this line, but is there any reason not to use `kWhiteSpaceCharacters` for the default value of the second argument of `TrimInPlace`? You should be able to turn `kWhiteSpaceCharacters` into `constexpr` if necessary since you are using `string::starts_with` which is available as of C++20. 

Also, is it possible to implement `TrimInPlace` as  a combination of `LTrimInPlace` and `RTrimInPlace` like the following? 

```c++
inline void TrimInPlace(std::string& str, std::string_view characters = " \t\n\r") {
    RTrimInPlace(str, characters);
    LTrimInPlace(str, characters);
}
```

tnakazato (commented):
As a CASA team member, I have reviewed this pull request. I left a few comments. I would appreciate it if you could take a look.

measures/Measures/MDirection.cc
aroffringa: I think that would make sense, but I didn't want to change the behaviour of the old trim function. The old function has `" \t\n\r"` as default, whereas `kWhiteSpaceCharacters` is `" \n\t\r\v\f"`.

I added `kWhiteSpaceCharacters` because these characters were used for trimming in some of the other modules. I'd prefer using the first string everywhere (without vertical tab and line feed), but I don't know if the vertical table or line feed symbols are still relevant there. I'll have a look and if I think it's save it remove `\v\f` I'll change `kWhiteSpaceCharacters` and use it as the default.

aroffringa: Oh and yes I also like your other suggestion, to implement TrimInPlace using the R/LTrim functions. I've changed that in #1502 (since that PR also needed another change to that function ; the char overload).

measures/apps/measuresdata/measuresdata.cc
aroffringa: Agree -- I tried to pay attention to that but maybe I've not always done that (like here). Indeed, in this place it looks like that if the file isn't consistent with the header, it could lead to substr() throwing. I've changed it now to what the old method did, i.e. to ignore this and push an empty string into `field`.

(It does look a bit like this functionality doesn't do error handling very well -- it could be argued that throwing is better if the file is misformed, but I'm not entirely sure so probably best to retain old behaviour).

measures/Measures/MDirection.cc
aroffringa: The (only) use of `kWhiteSpaceCharacters` including form feed and vertical tab is in coordinates. I'll rename it there to `kExtendedWhiteSpaceCharacters`, and then add `kWhiteSpaceCharacters` that equals `" \t\n\r"` . Does that make sense?

I'm not sure form feeds or vertical tabs can be part of fits files, so I think it's best not touch the trimmed strings in the coordinates module.

tnakazato: > The (only) use of kWhiteSpaceCharacters including form feed and vertical tab is in coordinates. I'll rename it there to kExtendedWhiteSpaceCharacters, and then add kWhiteSpaceCharacters that equals " \t\n\r" . Does that make sense?

That makes senes to me. Thanks!

tnakazato: I confirmed that `TrimInPlace` was re-implemented in the other PR. Thanks!

measures/apps/measuresdata/measuresdata.cc
tnakazato: We might see the exception at some point (hopefully in CI). We can fix it each time adaptively.

tnakazato (approved):
I'll approve the PR since all of my comments have been addressed. I didn't check if there are any oversights (i.e., I didn't try to find the lines that was not updated although they should be). I think any oversights should be detected by CI if test coverage is high enough.

PR #1498, opened 2026-08-14
Add test for splitName() in TableParseUtil and fix segfault on empty string

aroffringa:
The string changes require modification of this function, but because its behaviour was somewhat difficult to understand I first added a test set.

This uncovered that the function crashes when supplied with an empty string. I'm not sure this could ever occur but better to fix it, which is done in this PR.


tammojan (approved):
Thanks! I'm afraid I don't know all the use cases of the output of this function. But adding a test for current behavior is a good thing, specially before changing the implementation.

PR #1499, opened 2026-08-14
Replace specialized String functions with std replacements in tables

aroffringa:


tables/DataMan/test/tStManAll.cc
```diff
@@ -127,9 +127,9 @@ template<> Array<String> makeArray (const IPosition& shape,
 {
   Array<String> arr(shape);
```
enrique-alma-eso, line 128: This is still using the casa String. Is this intentional? Should it be replaced by std::vector<std::string> ?

tables/DataMan/StIndArrAIO.cc
```diff
@@ -143,7 +143,7 @@ StIndArray* StManColumnIndArrayAipsIO::getShape (rownr_t rownr)
     StIndArray* ptr = STMANINDGETBLOCK(rownr);
     if (ptr == 0) {
         throw (DataManInvOper ("ASM: no array in row " +
-			       String::toString(rownr) +
+			       String(std::to_string(rownr)) +
```
enrique-alma-eso, line 146: String(std::to_string(rownr))  -> std::to_string(rownr)

tables/DataMan/TiledStMan.cc
```diff
@@ -405,7 +405,7 @@ TSMCube* TiledStMan::getTSMCube (uInt hypercube)
 {
     if (hypercube >= nhypercubes()  ||  cubeSet_p[hypercube] == 0) {
       throw (AipsError ("TiledStMan::getTSMCube - hypercube nr "
-			+ String::toString(hypercube) + " does not exist in " +
+			+ String(std::to_string(hypercube)) + " does not exist in " +
```
enrique-alma-eso, line 408: String(std::to_string(hypercube)) -> std::to_string(hypercube)

tables/DataMan/TSMCoordColumn.cc
```diff
@@ -132,7 +132,7 @@ IPosition TSMCoordColumn::shape (rownr_t rownr)
     }
     if (! hypercube->valueRecord().isDefined (columnName())) {
         throw (DataManInvOper ("TSMCoord: no array in row " +
-			       String::toString(rownr) +
+			       String(std::to_string(rownr)) +
```
enrique-alma-eso, line 135: String(std::to_string(rownr)) -> std::to_string(rownr)

tables/DataMan/TSMDataColumn.cc
```diff
@@ -138,7 +138,7 @@ IPosition TSMDataColumn::shape (rownr_t rownr)
     IPosition cubeShape = hypercube->cubeShape();
     if (cubeShape.nelements() == 0) {
         throw (DataManInvOper ("TSM: no array in row " +
-			       String::toString(rownr) +
+			       String(std::to_string(rownr)) +
```
enrique-alma-eso, line 141: String(std::to_string(rownr)) -> std::to_string(rownr)

tables/Tables/ColumnSet.h
```diff
@@ -264,7 +264,7 @@ class ColumnSet
 
     // Make a unique data manager name by appending a suffix _n if needed
     // where n is a number that makes the name unique.
-    String uniqueDataManagerName (const String& name) const;
+    String uniqueDataManagerName (const std::string& name) const;
```
enrique-alma-eso, line 267: There are quite a number of public functions in this class that also use String rather than std::string. I understood that this set of PRs would change the internal use of String but not the public API. It is probably fine still to have this change since the implicit conversion, right? But in this case, why not for the other public methods?

tables/TaQL/TableParseProject.cc
```diff
@@ -134,24 +134,24 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
...
     } else {
       if (!negate) {
-        shInx = str.index("\\.");
-        if (shInx >= 0) {
-          shorthand = str.before(shInx);
-          str       = str.after(shInx+1);
+        shInx = str.find("\\.");
+        if (shInx != std::string::npos) {
+          shorthand = str.substr(0, shInx);
+          str       = str.substr(shInx+2);
```
enrique-alma-eso, line 154: Could this pos=shInx+2 argument in substr() be larger than size() ?

tables/DataMan/test/tStManAll.cc
aroffringa: The current pull requests don't try to already replace `casacore::String` by `std::string`, but rather avoid all use of methods that std::string doesn't have. Once that is done for all modules, casacore::String can be replaced by std::string. 

There are some places where I already did the replacement, because in some places it became easier ; I'll mention that in your next comment...

tables/DataMan/StIndArrAIO.cc
aroffringa: I've added the String cast temporarily to make this compile. The old toString() function returned a casacore String, whereas std::to_string returns a std::string. The String class doesn't allow concatenation with std::string, only with casacore::String (or const char*). So the old code concatenated:
a const char* plus String plus const char* plus String.
Without the cast, the new code would add:
a const char* plus std::String plus const char* plus String, which isn't possible. Once the last parameter is a std::string, the cast becomes unnecessary.

I have a note for myself to, once replaced,  make sure to look at all string casts and remove those if they are unnecessary.

tables/DataMan/TiledStMan.cc
aroffringa: This is for the same reason as above; hypercolumnName_p still is a casacore::String.

tables/DataMan/TSMCoordColumn.cc
aroffringa: Also here, columnName() is still a casacore::String and so can't be concatenated with a std::string

tables/DataMan/TSMDataColumn.cc
aroffringa: Also the same; columName is still a casacore::String

tables/Tables/ColumnSet.h
aroffringa: Fair point; I might not have always been consistent (nor could you know my motivation, which I hadn't really explained). The reason I've changed this to std::string is because it made the implementation easier:
```
String ColumnSet::uniqueDataManagerName (const std::string& name) const
{
    String dmName = name;
    Int nr = 0;
    while (! checkDataManagerName (dmName, 0, String(), False)) {
        nr++;
	dmName = name + '_' + std::to_string(nr);
    }
    return dmName;
}
```
If name were a casacore::String, a cast would be required to make the concatenation compile.

I agree that I've not always been consistent in what approach to take ; move towards std::string or solve it with a cast for now. I picked std::string here because it was relatively easy to do. The intention is though to in the end have std::string s anywhere (and no unnecessary casts).

tables/TaQL/TableParseProject.cc
aroffringa: Good to check indeed. From what I can tell, it can't be larger than size; it is the result of a find, which is a position at which the two character string "\\." starts, and so it must have two more characters. E.g., with the string "`abc\\.`" of size 5, shInx will be 3 and so shInx+2 will exactly be size. Would you agree or am I making a mistake?

tables/Tables/ColumnSet.h
enrique-alma-eso: Thanks for explaining!

aroffringa:
Thanks for the review! If you're fine with having it merged, could you approve the PR, @enrique-alma-eso ?

enrique-alma-eso (approved):

PR #1500, opened 2026-08-14
Replace specialized string methods in the casa module

aroffringa:


casa/BasicMath/Random.cc
```diff
@@ -439,13 +439,13 @@ String Random::asString(Random::Types type) {
 
 Random::Types Random::asType(const String& str) {
   String canonicalCase(str);
-  canonicalCase.upcase();
+  ToUpperCaseInPlace(canonicalCase);
```
schiebel, line 442: I think it would be worth considering updating `ToUpperCaseInPlace`. `std::toupper` accepts an `int` (rather than a `char`). Typically `char` is signed. `toupper` et al expect a value that is representable as an unsigned character so passing a negative (`char`) value is undefined behavior... and the fix is easy. The current:
```cpp
inline void ToUpperCaseInPlace(std::string& str) {
  std::transform(str.begin(), str.end(), str.begin(), [&](char c) { return toupper(c); });
}
```
becomes:
```cpp
inline void ToUpperCaseInPlace(std::string& str) {
  std::transform(str.begin(), str.end(), str.begin(),
                 [](unsigned char c) { return static_cast<char>(std::toupper(c)); });
}
```
This might be an issue for other string helpers. Also, `<casa/BasicSL/String.h>` doesn't seem to have a `using` statement for `std::toupper` (maybe because `<cctype>` leaks it?)... so it probably should be qualified or explicitly exposed.

casa/Exceptions/Error.h
```diff
@@ -436,7 +436,7 @@ class SystemCallError : public AipsError
   // and the errno.
   SystemCallError(const String &funcName, int error, Category c=GENERAL);
 
-  SystemCallError (int error, const String &msg, const String &filename,
+  SystemCallError (int error, const std::string &msg, const String &filename,
```
schiebel: Was leaving `const String &filename` intentional?

casa/Exceptions/Error2.cc
```diff
@@ -201,10 +201,10 @@ SystemCallError::SystemCallError(const String& funcName, int error, Category c)
   : AipsError("Error in " + funcName + ": " + errorMessage(error), c),
     itsError (error)
 {}
-SystemCallError::SystemCallError (int error, const String &msg,
+SystemCallError::SystemCallError (int error, const std::string &msg,
                                   const String &filename,
```
schiebel: Another `const String &filename`... I'm now suspecting that this is indeed intentional... :-)

casa/Inputs/Input.cc
```diff
@@ -271,12 +271,12 @@ Bool Input::put (const String& key, const String& value)
 Bool Input::put (const String& key)
 {
   String k = key;                  // Need non-const string
-  String::size_type inx = key.index("=");
+  String::size_type inx = key.find('=');
```
schiebel: Could be changed to `std::string::size_type`?  Same for `String::npos`...

schiebel (approved):
These were all that I found. I'll leave it to you to decide whether to make any changes or not.

casa/BasicMath/Random.cc
aroffringa: Good catch! I've changed both ToUpperCaseInPlace and ToLowerCaseInPlace as suggested, as well as CapitalizeStringInPlace.

casa/Exceptions/Error.h
aroffringa: Partly; I've changed parameters only when it was necessary inside the function. Because some of the new string functions return std::string instead of String, there were some issues when doing e.g. casacore::String + std::string, and so in those cases I made sure the first string became a std::string.

Once all use of String-specialized methods is gone, these left over String's will change, so in the end they will be converted, but there will be a in-between stage where it isn't the case and there may be some inconsistency.

casa/Exceptions/Error2.cc
aroffringa: Agree, made both std::string

casa/Inputs/Input.cc
aroffringa: Good to change indeed ; fixed!

casa/IO/StreamIO.cc
```diff
@@ -53,27 +53,25 @@ StreamIO::StreamIO(const String& hostname, uShort portNumber)
   memset(&serverInfo, 0, sizeof(serverInfo));
   serverInfo.sin_family = AF_INET;
 
-  Regex anyLetters("[A-Za-z]");
-  if(hostname.contains(anyLetters)){
-    struct hostent *hp = gethostbyname(hostname.chars());
+  const std::regex anyLetters("[A-Za-z]");
```
enrique-alma-eso, line 56: Not really related to this PR, but this seems to me that is trying to distinguish between hostnames and IP addresses. But this regexp won't work for IPv6 addresses. Maybe as an alternative getaddrinfo() can be considered. In fact gethostbyname() was deprecated in POSIX (https://pubs.opengroup.org/onlinepubs/009695399/functions/gethostbyname.html) and already removed in POSIX.1-2008.

casa/OS/Path.cc
```diff
@@ -240,7 +240,7 @@ String Path::baseName() const
     Int i=len;
     while (--i >= 0  &&  name[i] != '/') {}
     // The base name is the part from the slash till the end.
-    return name(i+1, len-i-1);
```
enrique-alma-eso, line 243: This used to create a SubString, which is just a reference to the String. Under STL this creates a new std::string, which might (not tested!) have worst performance in terms of CPU and memory. I guess the likelihood is very small, but if there is a a hot path that calls basename() a lot maybe there is some performance considerations.

PR #1501, opened 2026-08-14
Replace non-std string functions in ms module

aroffringa:


derivedmscal/DerivedMC/MSCalEngine.cc
```diff
@@ -397,9 +397,9 @@ void MSCalEngine::fillAntPos (Int calDescId, Int calInx)
   antMB.reserve  (tab.nrow());
   for (rownr_t i=0; i<tab.nrow(); ++i) {
     String mount = mountCol(i);
-    mount.downcase();
+    ToLowerCaseInPlace(mount);
     Int mountType = 0;
-    if (mount.size() >= 6  &&  mount(0,6) == "alt-az") {
+    if (mount.substr(0,6) == "alt-az") {
```
enrique-alma-eso: As reported by @tnakazato somewhere else, this would raise an exception if the mount string has less than 7 characters, so a column in the MS with a value that has less than 7 characters would raise.

ms/MSOper/MSSummary.cc
```diff
@@ -1050,7 +1050,7 @@ void MSSummary::listField (LogIO& os, Record& outrec,  Bool verbose, Bool fillRe
                     String err;
```
mnijhuis-tos, line 1050:
```suggestion
                    std::string err;
```

ms/MSOper/MSSummary.cc
```diff
@@ -1050,7 +1050,7 @@ void MSSummary::listField (LogIO& os, Record& outrec,  Bool verbose, Bool fillRe
...
                     mh.toRecord(err, dirrec);
                     fieldrec.defineRecord("direction", dirrec);
-                    String fieldrecid=String("field_")+String::toString(fld);
+                    String fieldrecid="field_"+std::to_string(fld);
```
mnijhuis-tos:
```suggestion
                    std::string fieldrecid="field_"+std::to_string(fld);
```

ms/MSSel/MSSelector.cc
```diff
@@ -854,13 +854,14 @@ Record MSSelector::getData(const Vector<String>& items, Bool ifrAxis,
...
 						Int ant2 = ifrAxis_p(k)%1000;
-						ifrName(k)=antName(ant1).before('\0');
+            constexpr std::string_view kTerminatingZero("\0", 1);
+						ifrName(k)=std::string(GetStringUpToExcluding(antName(ant1), kTerminatingZero));
 						ifrName(k)+="-";
-						ifrName(k)+=antName(ant2).before('\0');
+						ifrName(k)+=GetStringUpToExcluding(antName(ant2), kTerminatingZero);
 						if (prefix.length()>1) {
-							sName(k)=String(antName(ant1).after(prefix)).before('\0');
+							sName(k)=std::string(GetStringUpToExcluding(GetStringAfter(antName(ant1), prefix), kTerminatingZero));
```
mnijhuis-tos, line 862: Since the `+=`  below works without an extra conversion to `std::string`, assigning a `string_view` to a `std::string` variable should probably work as well.
```suggestion
							sName(k)=GetStringUpToExcluding(GetStringAfter(antName(ant1), prefix), kTerminatingZero);
```

ms/MSOper/MSSummary.cc
```diff
@@ -765,7 +765,7 @@ void MSSummary::listAntenna (LogIO& os, Bool verbose) const
     if (verbose) {
         // Detailed antenna list
         String title;
```
mnijhuis-tos:
```suggestion
        std::string title;
```

ms/MSOper/MSSummary.cc
```diff
@@ -765,7 +765,7 @@ void MSSummary::listAntenna (LogIO& os, Bool verbose) const
...
-        title="Antennas: " + String::toString(nAnt) + ":";
+        title="Antennas: " + std::to_string(nAnt) + ":";
         String indent("  ");
```
mnijhuis-tos:
```suggestion
        std::string indent("  ");
```

ms/MSOper/MSMetaData.cc
```diff
@@ -2525,10 +2525,10 @@ std::set<Int> MSMetaData::getFieldIDsForField(
     String name = field;
     vector<String> fieldNames = getFieldNames();
```
mnijhuis-tos, line 2526:
```suggestion
    std::string name = field;
    std::vector<std::string> fieldNames = getFieldNames();
```

mnijhuis-tos (approved):
Looks good. I only have suggestions for further improvement, but as is it is already ok. One main thing is that there are still `String` objects remaining, which should be replaced by `std::string` in the end, but that may also happen later.

sanbw (approved):
The changes look quite reasonable and understandable. I have visually inspected all the files and have some minor comments/suggestions -- just so that we don't forget when we get to the cleanup stage later.

1. In several files, `casacore::String` is still in use.  However, as @aroffringa explained, these will be replaced with `std::string` later when these changes are propagated to user packages.
2. Same with the use of `casacore::Regex` instead of `std::regex`.
3. Code in `MSLister.cc` could be reduced with a generic function that could serve the purpose of `IsAsciiLetter()` and other string comparison (like via `RegexMatches()`).
4. `MSSel/MSSelector.{cc,h}`: This was superseded by `MSSelection` class (a long time ago!), but is still here since it is _still_ used in a single file in `CASA`.  This could be deprecated in a later clean-up operation (and the changes propagated to the `CASA` code base).

ms/MSOper/MSSummary.cc
aroffringa: I'll skip this one -- Changing this causes a compile error, because err is sent to a function accepting a casacore::String reference.

ms/MSSel/MSSelector.cc
aroffringa: Unfortunately sName has still element type casacore::String, and assigning a string_view to that doesn't work (that could have been added in retrospect, but for now we've used casts in the other PRs -- once everything is std::string I'll go through the casts).

ms/MSOper/MSMetaData.cc
aroffringa: As of yet, getFieldNames() returns a vector<String>, which can't be converted to a std::vector<std::string>.

aroffringa:
Thanks for the review @sanbw . Point 1 and 2 will be performed in the next string-refactor steps. For step 3, I'll see if I can make IsAsciiLetter() into a generic function in a separate PR. Step 4, very useful to know that MSSelector is almost ready to be deprecated. Would giving it a [[deprecated]] annotation be acceptable? That will cause the deprecation warning while compiling Casa, but doesn't break compilation, so it's a good middle step.

Thanks again -- I'll merge the PR.

sanbw:
Personally, I think giving `MSSel/MSSelector` a [[deprecated]] annotation will be useful (it won't let us forget to propagate the necessary changes!).

PR #1502, opened 2026-08-14
Replace non-std string functions in coordinates module

aroffringa:


coordinates/Coordinates/FITSCoordinateUtil.cc
```diff
@@ -595,14 +595,14 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
...
-	        if(header[i].contains("-GLS")){
+	        if(StringContains(header[i], "-GLS")){
 	  	    os << LogIO::WARN << "Note: The GLS projection is deprecated. Use SFL instead." << LogIO::POST;
 	        }
 		all = all.append(header(i));
 	    }
 	    delete [] tmp;
 	}
-	char* pChar2 = const_cast<char *>(all.chars());
+	char* pChar2 = all.data();
```
tnakazato, line 606: I found that `c_str()` is used elsewhere. What is the proper use of these two methods? Is my understanding correct that `c_str()` is for read-only (const), while `data()` is for writable (non-const) output?

coordinates/Coordinates/Projection.cc
```diff
@@ -360,39 +360,28 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
...
+  // Strip trailing spaces
+  RTrimInPlace(ctypeLong, kWhiteSpaceCharacters);
+  RTrimInPlace(ctypeLat, kWhiteSpaceCharacters);
 
-   Int i1 = ctypeLong.index(RXwhite,0);
-   if (i1>=0) ctypeLong = String(ctypeLong.before(i1));
-//
-   i1 = ctypeLat.index(RXwhite,0);
-   if (i1>=0) ctypeLat = String(ctypeLat.before(i1));
+  String proj1(ctypeLong.substr(4));
```
tnakazato: `std::out_of_range` will be thrown when `ctypeLong.size() < 4`.

coordinates/Coordinates/Projection.cc
```diff
@@ -360,39 +360,28 @@ namespace casacore { //# NAMESPACE CASACORE - BEGIN
...
-   }
-//
-   return type(proj1);
+    // Get rid of leading -'s
+  LTrimInPlace(proj1, '-');
+  LTrimInPlace(proj2, '-');
+
+  // Get rid of spaces
+  LTrimInPlace(proj1, ' ');
+  LTrimInPlace(proj2, ' ');
```
tnakazato: If intention of the original code was to remove all whitespaces including the ones in the middle, this is incompatible change.

https://casacore.github.io/casacore/classcasacore_1_1String.html#adfc11278d6871fca0860dd31cc165927

tnakazato (commented):
As an CASA team member, I have reviewed this pull request. I posted a few comments. I would appreciate it if you could take a look.

coordinates/Coordinates/FITSCoordinateUtil.cc
aroffringa: That's correct. I generally prefer using `.c_str()`, but that returns const char*, in this case that didn't work (without a const cast), and so here I thought the use of `.data()` is cleaner.

I've added a comment to explain this.

coordinates/Coordinates/Projection.cc
aroffringa: Good catch! I now made the substr call conditional.

coordinates/Coordinates/Projection.cc
aroffringa: Good point, this indeed changes the behaviour, and will indeed not remove spaces that are inside a projection keyword, where as the old code would. That means I think that `RA--T AN` would be accepted before (with one too few dash and a space in DEC), whereas now it will now longer be accepted (and will raise an exception in the Project(string,string,vector) constructor.

From reading the code, I thought the new behavior is intended, though I could be wrong. The old code also does:
```
   proj1.gsub(Regex(" *"), String(""));
   proj2.gsub(String(" "), String(""));
```
I don't see a reason why proj2 is treated differently here compared to proj1 (though I guess they may have the same behaviour of removing all spaces), so I suspect this might have been not fully as intended: I think trimming spaces is sufficient.

However, I used LTrimInPlace, whereas I think right spaces should also be removed, so I've changed that now (I also added a TrimInPlace overload with char parameter in this PR now).

I suspect (hope) there aren't any fits files around that deliberately used things like `RA--TA N`, so I don't think this change affects anyone. If it does affect someone, I think reporting that the fits header is wrong is probably good (and the fix for them shouldn't be too hard), so I think this is better. Would you agree?

coordinates/Coordinates/FITSCoordinateUtil.cc
tnakazato: Understood. Thanks!

coordinates/Coordinates/Projection.cc
tnakazato: Yes, I agree. Thank you for sharing your thought and updating the code. Trimming whitespaces from both side should be sufficient practically. I agree that `RA--TA N` should be a bug in the application that generated FITS. No need to support it.

tnakazato (approved):
Although I didn't check if there are any oversights, I would like to approve this PR. Any oversight should be detected in CI here of the ones in the subsequent PRs (if test coverage is high enough).

aroffringa:
Thanks!

PR #1504, opened 2026-08-15
Replace non-std string in image, lattices, scimath, msfits, python mods

aroffringa:


images/Images/FITSImage.cc
```diff
@@ -257,12 +257,12 @@ uInt FITSImage::get_hdunum(const String &fullname)
 		// check for the comma
 		comma_pos = extstring.rfind(",", extstring.length());
 		if (comma_pos < 0) {
-			extstring.trim();
+			TrimInPlace(extstring);
 
 			// check whether an index is given
-			if (String::toInt(extstring)){
+			if (StringToInt(extstring)){
```
hmuellergoe: StringToInt() is not equivalent to String::toInt() when nonthrowing conversion fails: its return-by-value helper declares T value without initialization, so empty input may produce an indeterminate result instead of zero (as previously). Maybe value-initialize that helper with T value{} and preferably use the Boolean output-parameter overload here so parsing is performed once and zero is not treated as conversion failure. The same concern applies at lines 281 and in FITSImgParser.cc:182.

msfits/MSFits/MSFitsOutput.cc
```diff
@@ -3070,14 +3071,18 @@ void MSFitsOutput::_handleAntNumbers(const MeasurementSet& ms,
 
     for (Int iant = 0; iant < nAnt; ++iant) {
         String name;
-        if (arrayName.contains("VLA"))
+        if (StringContains(arrayName, "VLA")) {
             // Trim leading EA/VA, if present
-            name = antname(iant).from(RXint);
-        else
             name = antname(iant);
+            const size_t int_pos = antname(iant).find_first_of(kIntegerCharacters);
```
hmuellergoe, line 3077: Minor compatibility note: searching for the first character in kIntegerCharacters is not fully equivalent to searching for the first RXint match. For example, the old code extracts 12 from EA-X12, while the new code starts at the earlier hyphen, produces -X12, and rejects the name. Conventional EA<number> and VA<number> names are unaffected; is narrowing support to those forms intentional?

msfits/apps/ms2uvfits.cc
```diff
@@ -95,7 +95,7 @@ int main (int argc, const char* argv[])
 	}
 	if (fitsfile == "") {
 	  fitsfile = msin;
-	  fitsfile = fitsfile.before(Regex("\\.MS$")) + ".UVF";
+	  fitsfile = fitsfile.substr(0, fitsfile.size() - 3) + ".UVF";
```
hmuellergoe: This now removes the final three characters unconditionally, while the old code removed them only when the name ended in uppercase .MS. Inputs without that suffix, such as "dataset", change from "dataset.UVF" to "data.UVF". I am concerned about external pipelines which do not always follow strict naming.

hmuellergoe: proposed code change:
      constexpr std::string_view kExtension = ".MS";
	  if (fitsfile.ends_with(kExtension)) {
	    fitsfile.resize(fitsfile.size() - kExtension.size());
	  }
	  fitsfile += ".UVF";

images/Images/MIRIADImage.cc
```diff
@@ -587,24 +587,22 @@ void MIRIADImage::getImageAttributes (CoordinateSystem& cSys,
...
     proj2 = ctype(latAxis);
 
-    if (proj1.contains("GLON")) isGalactic = True;
+    if (StringContains(proj1, "GLON")) isGalactic = True;
 
     // Get rid of the first 4 characters, e.g., RA--
 
     const Int l1 = proj1.length();
     const Int l2 = proj2.length();
-    proj1 = String(proj1.at(4, l1-4));
```
hmuellergoe, line 596: String::at() previously clamped an out-of-range starting position, while std::string::substr() throws. Conventional CTYPE values such as RA---SIN and DEC--SIN are safe, but short recognized values such as RA and DEC previously produced empty projections and reached the existing LinearCoordinate fallback; these lines now throw first. Is stricter rejection of such legacy or malformed MIRIAD headers intended?

msfits/MSFits/FitsIDItoMS.cc
```diff
@@ -1079,8 +1075,8 @@ void FITSIDItoMS1::describeColumns()
 		// Vector of strings or degenerated vector.  Use the
 		// substring inside the parentheses as shape.
 		//
-		dimstr = SHAPEstr(1,SHAPEstr.length()-2);
-		shape(0) = atoi(dimstr.chars());
+		dimstr = SHAPEstr.substr(1,SHAPEstr.length()-2);
```
hmuellergoe: The old SHAPEstr(1, ...) operation clamped an excessive starting position, while substr(1, ...) throws if a malformed or empty MSK SHAPE value reaches this branch. Is the new stricter behavior intentional, or should the value be validated before extracting its contents? A similar risk exists at FitsIDItoMS.cc:752 if an unusually short keyword begins with MSK.

images/apps/imageregrid.cc
```diff
@@ -78,7 +78,7 @@ int main(int argc, const char* argv[]) {
       out = "regridded_"+in;
       cout << "No output name give using '" << out << "'." << endl;
     }
-    Bool outisfits = downcase(out).after(out.size()-6) == ".fits";
+    Bool outisfits = downcase(out).substr(out.size()-5) == ".fits";
```
hmuellergoe: For output names shorter than five characters, out.size() - 5 underflows and substr() throws. Probably better to use a direct suffix test instead. The identical issue also occurs in images/apps/imageslice.cc:60.

hmuellergoe: e.g.: Bool outisfits = downcase(out).ends_with(".fits");

hmuellergoe (commented):
The standard-string migration generally looks good. I found three, quite minor potentially actionable behavior changes involving failed integer conversion, unconditional output-name truncation, and short output names. I also left compatibility questions for unusual VLA names and malformed or legacy MIRIAD/FITS headers.

images/Images/FITSImage.cc
aroffringa: Thanks, that's a very good catch. I've initialized the `value` variable now and changed this to
```
			if (int value; StringToValue<int>(extstring, value, false)){
				// get the index
				extindex = value;
			}
```
I've also added a check for the return value of StringToInt() in the unit tests when an error is found.

Does this indeed solve your points? This changes the behaviour (it now accepts a value of zero), but I think that's an improvement and assume was your suggestion. I changed the other two places you point out likewise.

msfits/MSFits/MSFitsOutput.cc
aroffringa: I don't think I did this intentionally, so again good catch. Upon reading the code and the comments, the intention is just to remove EA or VA from the start of the string, in which case the regex is probably a bit overkill. Moreover, I didn't make a replacement for the `String.from(Regex)` because it wasn't necessarily elsewhere. I therefore have a slight preference not to introduce it here to be consistent with the old code, but to keep the find_first_of() code as it is now. Would you agree?

msfits/apps/ms2uvfits.cc
aroffringa: Another mistake by me, thanks, I've taken your proposed change.

images/Images/MIRIADImage.cc
aroffringa: You're right -- I think I overlooked this subtle difference between substr and at. I've now added checks to the cases where it is necessary, including this one.

msfits/MSFits/FitsIDItoMS.cc
aroffringa: Also good find, I've also made this conditionally

images/apps/imageregrid.cc
aroffringa: Agree, nice change -- applied.

msfits/MSFits/MSFitsOutput.cc
hmuellergoe: Agreed. Current VLA and ALMA antenna names are unaffected. I will simply make a note for future ngVLA naming.

hmuellergoe (approved):
looks all fine.

aroffringa:
Thanks for the review!

PR #1505, opened 2026-08-20
Fix some small string-transformation issues

aroffringa:
- Makes sure that the in #1493 newly added functions behave as their old counterparts.
- Fix an issue in RegexSubStr() when no matches are found
- Refer to std::to_string in deprecation message of toString()
- Add a kIntegerCharacters constant that can be used to replace regex searches.

### DIFF
--- a/casa/BasicSL/test/tString.cc
+++ b/casa/BasicSL/test/tString.cc
@@ -264,9 +264,14 @@ BOOST_AUTO_TEST_CASE(StringToInt) {
   BOOST_CHECK_EQUAL(StringToInt("-0"), 0);
   BOOST_CHECK_EQUAL(StringToInt("-32768"), -32768);
 
-  BOOST_CHECK_THROW(StringToInt(""), std::runtime_error);
-  BOOST_CHECK_THROW(StringToInt("0?"), std::runtime_error);
-  BOOST_CHECK_THROW(StringToInt("1e12"), std::runtime_error);
+  BOOST_CHECK_THROW(StringToInt("", true), std::runtime_error);
+  BOOST_CHECK_THROW(StringToInt("0?", true), std::runtime_error);
+  BOOST_CHECK_THROW(StringToInt("1e12", true), std::runtime_error);
+
+
+  BOOST_CHECK_NO_THROW(StringToInt("", false));
+  BOOST_CHECK_NO_THROW(StringToInt("0?", false));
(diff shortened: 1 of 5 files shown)


casa/Utilities/Regex.h
```diff
@@ -312,6 +312,7 @@ extern const Regex RXuppercase;      //# = "[A-Z]+"
 extern const Regex RXalphanum;       //# = "[0-9A-Za-z]+"
 extern const Regex RXidentifier;     //# = "[A-Za-z_][A-Za-z0-9_]*"
 
+constexpr std::string_view kIntegerCharacters = "-0123456789";
```
mickveldhuis, line 315: Is this used anywhere?

mickveldhuis: Ah guessing from the PRs summary, this is prep for later?

casa/BasicSL/test/tString.cc
```diff
@@ -264,9 +264,14 @@ BOOST_AUTO_TEST_CASE(StringToInt) {
   BOOST_CHECK_EQUAL(StringToInt("-0"), 0);
   BOOST_CHECK_EQUAL(StringToInt("-32768"), -32768);
 
-  BOOST_CHECK_THROW(StringToInt(""), std::runtime_error);
-  BOOST_CHECK_THROW(StringToInt("0?"), std::runtime_error);
-  BOOST_CHECK_THROW(StringToInt("1e12"), std::runtime_error);
+  BOOST_CHECK_THROW(StringToInt("", true), std::runtime_error);
+  BOOST_CHECK_THROW(StringToInt("0?", true), std::runtime_error);
+  BOOST_CHECK_THROW(StringToInt("1e12", true), std::runtime_error);
```
mickveldhuis, line 269: Why does this fail? Intuitively, I'd imagine that 10^12 should be convertible to an integer, or is it just not aware of exponential notation?

mickveldhuis (commented):
LGTM! Got a few very small questions.

casa/Utilities/Regex.h
aroffringa: Indeed, it's used in the 'rest' modules branch. It can replace the Regex in most cases.

casa/BasicSL/test/tString.cc
aroffringa: Indeed, integers are not allowed to be written in scientific notation.

mickveldhuis (approved):

PR #1506, opened 2026-09-03
Deprecate Fallible class in favour of std::optional

aroffringa:
The Fallible class implements functionality that can be almost one-to-one replaced by `std::optional`. Given that std::optional is now available, it's better to use that instead. It has r-value support and doesn't construct the type when it is empty, and is generally more accessible for programmers.

Of note is that Fallible always checks the access, and throws an exception if it is unset. For `std::optional<int> a;`, calling `*a` doesn't do this check, but `a.value()` does check and throws an exception if the value is unset. So in places in the code where it is unknown if a value is set, the `value()` method is used, and in places where it is known the value is set use `*a`.


casa/Utilities/test/tFallible.cc
mnijhuis-tos, line 1: Since Fallible.h is still there, and still works, please don't remove its test yet. We can remove it later, when removing Fallible.h.

scimath/Mathematics/StatAcc.tcc
```diff
@@ -168,64 +168,66 @@ uInt StatAcc<T>::getCount() const               // get number of samples
...
 	if (ms >= 0) {
-	    return Fallible<Double>(sqrt(ms));  // valid
+	    return std::optional<Double>(sqrt(ms));  // valid
 	} else {
-	    return Fallible<Double>(0);         // .....?
+	    return std::optional<Double>(0);         // .....?
 	}
     } else {
-	return Fallible<Double>();              // 
+	return std::optional<Double>();              //
```
mnijhuis-tos: Optional: Remove the (empty) comments here.

casa/Utilities/test/CMakeLists.txt
```diff
@@ -11,7 +11,6 @@ tCOWPtr
 tDataType
 tDefaultValue
 tDynBuffer
-tFallible
```
mnijhuis-tos, line 14: After restoring the test, please also undo this change.

aroffringa: That would cause the deprecation warnings in the casacore build, so I rather not. Fallible is no longer used anywhere inside Casacore, it is just in the code for if it is used by others. They now get a warning and should move to std::optional. I therefore don't see much use to keep running the unit tests.

mnijhuis-tos: Keeping the test code but disabling its compilation is also not very useful. The whole idea of the test is verifying that the code still compiles and works, since others may still use it.

However, I agree that deprecation warnings aren't nice either. You could disable those warnings by adding `target_compile_options(tFallible PRIVATE -Wno-deprecated-declarations)`. I would also add a comment saying that the warning is disabled for this test since it tests deprecated functionality.

aroffringa: I think we better not do that -- there's no real chance that the code stops working, and if it does it will only stop working for code that was already producing warnings. At most a new compiler could change e.g. include requirements, but if that happens it's not a big issue. Adding extra machinery, which have to be removed again when fallible is definite removed, to run the tests without warnings is imho not really pragmatic.

mnijhuis-tos: You can indeed also view it from a different angle: Since it's deprecated, we don't guarantee anymore that this code still works. I would then prefer removing the test again: Since we don't even compile it anymore, it becomes dead code, so we should remove it.

aroffringa: Sounds fair :+1: . I removed the test.

mnijhuis-tos (approved):

mnijhuis-tos (approved):

PR #1510, opened 2026-09-18
Replace macros in the DataManagerColumn class

aroffringa:
The DataManagerColumn class used some macros that could be replaced by functions. The macros also stopped working when casacore::String would be replaced by std::string, so besides cleaning it up it also helps that transition.


tables/DataMan/DataManagerColumn.cc
```diff
@@ -163,104 +163,109 @@ void DataManagerColumn::putOther (rownr_t, const void*)
...
+          i++;
+          rownr += incr;
+        } else {
+          rownr_t inx = (rownr - colCache_p.start()) * colCache_p.incr();
+          const T* cptr = static_cast<const T*>(colCache_p.dataPtr()) + inx;
+          rownr_t endrow = std::min (end, colCache_p.end());
+          while (rownr <= endrow) {
+	    vec[i++] = *cptr;
+            rownr += incr;
+	    cptr += incr * colCache_p.incr();
```
mnijhuis-tos, line 221: Since you're touching these lines anyway, please also fix the indentation.
```suggestion
	        vec[i++] = *cptr;
            rownr += incr;
	        cptr += incr * colCache_p.incr();
```

tables/DataMan/DataManagerColumn.cc
```diff
@@ -163,104 +163,109 @@ void DataManagerColumn::putOther (rownr_t, const void*)
...
+          }
+	}
+      }
```
mnijhuis-tos, line 224:
```suggestion
          }
	    }
      }
```

tables/DataMan/DataManagerColumn.cc
```diff
@@ -163,104 +163,109 @@ void DataManagerColumn::putOther (rownr_t, const void*)
...
+      const rownr_t* rows = rowvec.getStorage (delR);
+      const T* cptr = static_cast<const T*>(colCache_p.dataPtr());
+      rownr_t strow  = colCache_p.start();
+      rownr_t endrow = colCache_p.end();
+      for (rownr_t i=0; i<nr; ++i) {
+	rownr_t rownr = rows[i];
+        if (rownr >= strow  &&  rownr <= endrow) {
+	  vec[i] = cptr[(rownr-strow)*colCache_p.incr()];
+	} else {
+          getGeneric<T>(rownr, &(vec[i]));
```
mnijhuis-tos, line 241:
```suggestion
      for (rownr_t i=0; i<nr; ++i) {
	    rownr_t rownr = rows[i];
        if (rownr >= strow  &&  rownr <= endrow) {
	      vec[i] = cptr[(rownr-strow)*colCache_p.incr()];
	    } else {
          getGeneric<T>(rownr, &(vec[i]));
```

tables/DataMan/DataManagerColumn.h
```diff
@@ -449,6 +449,38 @@ class DataManagerColumn
     virtual void getOther    (rownr_t rownr, void* dataPtr);
     // </group>
 
+    template <typename T>
+    void getGeneric(rownr_t rownr, T* dataPtr)
+    {
+      if constexpr (std::is_same_v<T, Bool>) {
```
mnijhuis-tos, line 455: Instead of using std::is_same_v, template specialisation seems like a more natural approach. For example, use
```
    template <typename T>
    void getGeneric(rownr_t rownr, T* dataPtr) { getOther(rownr, dataPtr); }

    template<>
    void getGeneric(rownr_t, Bool* dataPtr) { getBool(rownr, dataPtr); }

    ... etc ...
```

The same argument holds for `putGeneric`.

aroffringa: I had asked Chatgpt what it prefers, and it prefers the current is_same_v structure.

> But that is substantially more verbose, and the mapping is spread over many separate function definitions. 
> Your current implementation keeps the entire dispatch table together>

I think I agree with chatgpt here; although I don't have a strong opinion, the use of `if` in this repetitive structured approach makes it quite easy to read, and the generated code is the same.

mnijhuis-tos: An advantage of template specialisation is that it works with old C++ versions, even pre-C++11. `std::is_same_v` is from C++-17. Still, that's not an argument for repositories like casacore which require C++17 (or later). Since both options work, it's just a matter of opinion, so I'll resolve this discussion.

mnijhuis-tos (approved):

PR #1513, opened 2026-09-19
Add code formatting scripts

aroffringa:
I've used the .clang-format file from wsclean, which I think is vetted quite thoroughly.

The formatting changes "//#" to "// #", and these lines should be ignored by doxygen. This PR also changes the doxygen parsing such that "// #" is ignored.


scripts/format-implementation.sh
```diff
@@ -0,0 +1,82 @@
...
+
+# Check arguments
+if [ -z "$SOURCE_DIR" ]; then
+  echo "Please define SOURCE_DIR using $BASH_SOURCE"
+  exit 1
+fi
+if [ -z "$CXX_SOURCES" ]; then CXX_SOURCES=(*.cc *.h *.tcc); fi
+
+# Detect run environment.
+if [ -n "$CI" ]; then
```
mnijhuis-tos, line 32: According to https://docs.github.com/en/actions/reference/workflows-and-actions/variables, GitHub indeed also defines a CI variable, like GitLab.

scripts/run-format.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# Usage: run-format.sh [<module>]
+# A module can optionally be specified as the first parameter, in
+# which case only that module will be formatted. If no module is specified,
+# all modules are formatted.
+
+# Disable globbing. This is needed when defining patterns that have wildcards.
+set -e -f
+
+CLANG_FORMAT_BINARY=clang-format-21
```
mnijhuis-tos, line 12:
```suggestion
# Use a specific clang-format version, since formatting may differ between versions.
CLANG_FORMAT_BINARY=clang-format-21
```

mnijhuis-tos (approved):

PR #1515, opened 2026-09-19
Format derivedmscal module

aroffringa:


gervandiepen:
I noticed that //# comments are turned into // #.
The script doxygen_p (in build_tools) is used to generate documentation
from the source files by doxygen. This script ignores comments starting
with //#. So this change will invalidate the script.

aroffringa:
> I noticed that //# comments are turned into // #. The script doxygen_p (in build_tools) is used to generate documentation from the source files by doxygen. This script ignores comments starting with //#. So this change will invalidate the script.

Thanks, I've updated the script to also ignore "// #" in #1513. The few occurrences in the code that are there before formatting seem already to be outside of doxygen comments. Running doxygen with the cfg file in the root seems to work fine with those changes -- thought we aren't running doxygen on the code regularly anymore though, so not sure what the state of it is.

tammojan (approved):

PR #1525, opened 2026-09-21
Fix operator!= in LCLattice due to C++20 switch

aroffringa:
In C++20, the operator== and != should be symmetric. This means that derived classes can't (trivially) have a comparison operator with their base class -- well, formally, it is allowed, but unfortunately g++11 implements this too strict. This causes LCRegion to fail to compile under g++11. It compiles okay under a later gcc, but it is still not the recommended approach in C++20.

The recommended method for having an asymmetric comparison operator is by implementing a protected virtual helper method. This is done in this PR, which fixes compilation under g++11.


mickveldhuis (approved):
LGTM! 😁

PR #1526, opened 2026-09-21
Fix some unit tests when running inside Docker

aroffringa:
When run inside Docker, some of the permission tests don't work.


tmillenaar:
Good to merge as far as I am concerned. It would have been nice to mark the affected tests as `skipped` if the user was root (instead of passing), but I don't think the current structure is set up well for this.

casa/OS/test/tDirectory.cc
```diff
@@ -182,11 +182,15 @@ void doIt (Bool doExcp)
     // Construct and create a new directory.
     Directory newDir ("tDirectory_tmp/newDir");
     tmp.setPermissions (0555);
+    const bool is_root_user = geteuid() == 0;
     if (doExcp) {
 	try {
 	    newDir.create();
+      // create() should have thrown an error, unless we are running this a root:
+      AlwaysAssertExit (is_root_user);
+      newDir.remove();
```
gmloose: Will this line ever be executed? The way I understand it is that `AlwaysAssertExit()` will raise an `AbortError`, so you will jump to the `catch` line (line 192).

gmloose: There also seems to be something wrong with indentation.

casa/OS/test/tDirectory.cc
```diff
@@ -215,8 +219,9 @@ void doIt (Bool doExcp)
     if (doExcp) {
 	try {
 	    Directory file1("tDirectory_tmp/something");
+      AlwaysAssertExit(is_root_user);
```
gmloose: Also here I see inconsistent indentation.

casa/OS/test/tRegularFile.cc
```diff
@@ -129,8 +131,12 @@ void doIt (Bool doExcp)
     if (doExcp) {
 	try {
 	    risFile1.copy (Path("tRegularFile_tmp/moveto/isFile1"));
+      // If user is root, the copy will have worked. Otherwise, an exception
+      // should have been thrown.
+      const bool is_root_user = (geteuid() == 0);
+      AssertAlways(is_root_user);
```
gmloose: Inconsistent indentation.

aroffringa:
> Good to merge as far as I am concerned. It would have been nice to mark the affected tests as `skipped` if the user was root (instead of passing), but I don't think the current structure is set up well for this.

Thanks! I agree, but indeed the current structure doesn't allow that. I'm moving slowly towards Boost tests, which does have that option, but the code base is probably too large to ever get all of it in Boost :-/.

casa/OS/test/tSymLink.out
```diff
@@ -1,7 +1,5 @@
 SymLink: tSymLink_tmp/isFile exists, but is no symbolic link
 SymLink: tSymLink_tmp/isDir exists, but is no symbolic link
-SymLink: tSymLink_tmp/isDir/newB does not exist and cannot be created
-SymLink::create error on a: Permission denied
```
gmloose: I guess you don't mind that the output of those two tests is now no longer checked against the reference output when the tests are run as an ordinary user, right?

gmloose (commented):
LGTM.  I found one real issue, I think (my first comment). The others are about indentation, and I stopped mentioning that at some stage, because all the modified files seem to suffer from this. It would be good to be consistent, though.

casa/OS/test/tDirectory.cc
aroffringa: If is_root_user is true, the newDir will be removed. This is necessary for the rest of the tests, which assume the dir is not created.

aroffringa: I'm about to run format checks by default (in a different PR ;) ), that'll take care of this :).

casa/OS/test/tSymLink.out
aroffringa: Indeed ; it's already a bit harsh to check the exact text of the error message. Also, the current test setup doesn't really allow skipping a test otherwise (I would have to manually `cout` that msg to skip the test).

aroffringa:
Thanks for the review, @gmloose . Timo had just reviewed it too so I had already merged it, but I responded to your comments.

